PR #3, opened 2015-08-25
main: add initial implementation

crawford:


main.go
```diff
@@ -0,0 +1,175 @@
...
+			LocalIPv4:  local,
+			Hostname:   hostname,
+		}, nil
+	default:
+		panic("bad provider")
+	}
+}
+
+func write_metadata(out *os.File, metadata metadata) error {
+	if _, err := out.Write([]byte(fmt.Sprintf("COREOS_IPV4_PUBLIC=%s\n", metadata.PublicIPv4))); err != nil {
```
marineam: Write is silly because: http://golang.org/pkg/os/#File.WriteString
But then Sprintf is silly because: http://golang.org/pkg/fmt/#Fprintf

marineam: Also, writing out a `net.IP` is fine for now but how is this going to look for providers that have different sets of addresses they define? Writing out the zero value `0.0.0.0` isn't going to be useful.

crawford: If only there was a way to determine if a value was a zero value.

crawford: Updated.

marineam: You mean like `ip.Equal(net.IPv4zero)` Mr. Snarky? :-P

marineam:
LGTM

main.go
crawford: No, a general `IsZero` function. That way I don't need those two helpers.

marineam: Oops, actually, I mislead you because IP is a slice so the zero value is nil, not IPv4zero

marineam: Meaning you would still wind up with `FOO_IPV4=<nil>` with the code as-is for non ec2 things in the future.

crawford: Shit. You're right.

PR #5, opened 2015-09-16
travis: enable go 1.5

marineam:


crawford:
LGTM. I'll add the `go vet` include for 1.5 once we actually vet the code.

PR #6, opened 2015-09-29
providers: add support for Azure

crawford:


build
```diff
@@ -4,7 +4,7 @@ NAME="coreos-metadata"
 ORG_PATH="github.com/coreos"
 REPO_PATH="${ORG_PATH}/${NAME}"
 VERSION=$(git describe --dirty)
-GLDFLAGS="-X main.version \"${VERSION}\""
+GLDFLAGS="-X main.version=\"${VERSION}\""
```
jonboulle: Huh. We have this nonsense: https://github.com/coreos/rkt/blob/master/configure.ac#L27

vcaputo: doesn't = work everywhere?

jonboulle: I assumed not based on
https://github.com/coreos/rkt/commit/2b1a60db047f5b2a408586e01678c461446a5c80

crawford: It built in Travis...

src/providers/azure/azure.go
```diff
@@ -0,0 +1,182 @@
...
+
+	if err := assertFabricCompatible(addr, FabricProtocolVersion); err != nil {
+		return config.Metadata{}, err
+	}
+
+	return fetchSharedConfig(addr)
+}
+
+func getFabricAddress() (net.IP, error) {
+	lease, err := os.Open("/run/systemd/netif/leases/2")
```
vcaputo: this looks like it may be assuming too much

build
jonboulle: I dunno, /cc @krnowak

crawford: Wow, you've got to be fucking kidding me. It silently fails. Go strikes again!

src/providers/azure/azure.go
```diff
@@ -0,0 +1,182 @@
...
+	client := retry.Client{
+		InitialBackoff: time.Second,
+		MaxBackoff:     time.Second * 5,
+		MaxAttempts:    10,
+		Header: map[string][]string{
+			"x-ms-agent-name": {AgentName},
+			"x-ms-version":    {FabricProtocolVersion},
+			"Content-Type":    {"text/xml; charset=utf-8"},
+		},
+	}
```
vcaputo, line 130: Why are you repeating yourself?  Stick this in a function which returns the populated retry.Client and reuse with above, they're doing identical things.

src/providers/azure/azure.go
```diff
@@ -0,0 +1,182 @@
...
+				Id             string `xml:"id,attr"`
+				Address        string `xml:"address,attr"`
+				InputEndpoints struct {
+					Endpoints []struct {
+						LoadBalancedPublicAddress string `xml:"loadBalancedPublicAddress,attr"`
+					} `xml:"Endpoint"`
+				}
+			} `xml:"Instance"`
+		}
+	}{}
```
vcaputo, line 173: It just feels icky to have these guys which amount to schema sprinkled inline with the code, but I don't know what's "idiomatic".

src/providers/azure/azure.go
crawford, line 130: Unrepeated.

vcaputo:
lgtm overall

src/providers/azure/azure.go
crawford: What about `net.InterfaceByName("eth0").Index()`?

crawford: I could also just test all leases.

src/providers/azure/azure.go
crawford, line 173: I don't know either, but I figured I do it since I only ever use it once.

vcaputo:
lgtm now and forever

PR #8, opened 2015-09-29
Client cleanups

vcaputo:


src/providers/azure/azure.go
```diff
@@ -118,18 +132,9 @@ func assertFabricCompatible(endpoint net.IP, desiredVersion string) error {
...
-		InitialBackoff: time.Second,
-		MaxBackoff:     time.Second * 5,
-		MaxAttempts:    10,
-		Header: map[string][]string{
-			"x-ms-agent-name": {AgentName},
-			"x-ms-version":    {FabricProtocolVersion},
-			"Content-Type":    {"text/xml; charset=utf-8"},
-		},
-	}
+	client := getClient(true)
```
crawford: I really hate this.

crawford: I'd rather you attach the agent to everthing than have a random boolean parameter.

crawford:
I would have rather you left these comments on the original PR. I've already tagged 0.2.0.

vcaputo:
wah wah

vcaputo:
for posterity: I mentioned consolidating the clients in the other PR.  The want of a variadic Get didn't emerge til I wrote this...  Cleanups are of the lowest priority, tagging a version is irrelevant WRT this.

crawford:
LGTM

crawford:
The reason I brought up the tagging was because I generally like to ship the latest code in the OS; regardless of what the changes are.

vcaputo:
You might want to re-evaluate that approach, because cleanups are basically 100% risk with 0% end-user benefit.

crawford:
Then why are you doing the clean-up in the first place?

vcaputo:
More maintainable/easier to read code, of course, why do any cleanup or refactor?  It's to make our lives easier.  End-user impact is realized in the long-term, and generally harmful in the short-term.  You know this.

crawford:
I agree. But why wouldn't you want to includes those changes in the original PR and in the latest release? You are arguing that we shouldn't ship the latest code because it is high risk, but it's only high risk because you made the changes in two passes.

PR #10, opened 2015-10-21
main: add support for kernel cmdline

crawford:
This add the `--cmdline` flag which will cause the utility to read the
provider from the kernel command line parameters (coreos.oem.id).


marineam:
LGTM

PR #11, opened 2015-11-20
providers: rename environment variables

crawford:
Going from:

```
COREOS_IPV4_LOCAL
COREOS_IPV4_PUBLIC
COREOS_HOSTNAME
```

to:

```
COREOS_EC2_IPV4_LOCAL
COREOS_EC2_IPV4_PUBLIC
COREOS_EC2_HOSTNAME
COREOS_AZURE_IPV4_DYNAMIC
COREOS_AZURE_IPV4_VIRTUAL
```


vcaputo:
It would be neat to leverage struct tags for associating the environment variable names with their respective metadata struct members then just have the environment file get written out using reflection, but this seems perfectly serviceable as-is.

lgtm

PR #13, opened 2016-04-14
*: add support for writing SSH keys

crawford:


crawford:
WIP: Still needs to learn how to fetch keys.

crawford:
This should be good to go now. It can fetch and write SSH keys on EC2.

```
core@ip-172-31-4-46 ~ $ ./coreos-metadata -cmdline -ssh-keys core
fetching "http://169.254.169.254/2009-04-04/meta-data/public-ipv4": attempt #1
fetching "http://169.254.169.254/2009-04-04/meta-data/local-ipv4": attempt #1
fetching "http://169.254.169.254/2009-04-04/meta-data/hostname": attempt #1
fetching "http://169.254.169.254/2009-04-04/meta-data/public-keys": attempt #1
fetching "http://169.254.169.254/2009-04-04/meta-data/public-keys/0/openssh-key": attempt #1

core@ip-172-31-4-46 ~ $ cat .ssh/authorized_keys.d/coreos-metadata 
ssh-rsa AAAA... crawford
```

marineam:
Looks ok, what will the interaction be between this and the cloudinit based OEM?

crawford:
This isn't enabled by default, so there is no interaction.

marineam:
What about the interaction when it _is_ enabled :-P

crawford:
Well, then you would get duplicate keys. They would end up in `~core/.ssh/authorized_keys.d/coreos-cloudinit` and `~core/.ssh/authorized_keys.d/coreos-metadata`.

marineam:
ok, seems fine as a start but what is the longer term plan? Should the OEM start always enabling the new service and drop it from the cloudinit goo?

marineam:
anyway, that can be sorted later. LGTM

crawford:
Dropping it from cloudinit is definitely an option (I hadn't much thought about). Ignition is going to enable it on certain platforms where it is supported.

PR #14, opened 2016-04-27
providers/gce: add initial implementation

crawford:


marineam:
After re-reading https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys it sounds like I got the old behavior wrong, the old sshKeys instance attribute was supposed to override the project one, not amend it. Oops, oh well, at least the new value does have that behavior. With adopting the new value we should probably be compatible with the expected behavior this time around and respect block-project-ssh-keys=TRUE

This being a new component happily gives us a transition away from my old misinterpreted behavior, coreos-metadata using the new keys while my old crappy script does the old. Some time after the new has been available for a while and google updates their docs accordingly to add us to the list we can eventually drop the old. https://cloud.google.com/compute/docs/instances/ssh-keys#supported_images

marineam:
So in short looks mostly good but just wrap the use of projectSshKeys in a conditional based on the block value and we'll be set :)

crawford:
WTF, Google? Why does the [metadata documentation](https://cloud.google.com/compute/docs/metadata) make no mention of this "block" key?

crawford:
@marineam OK, I added support for "block-project-ssh-keys" and the legacy "sshKeys". From my understanding (and knowledge of the scope) of the docs, this works properly.

marineam:
LGTM

PR #16, opened 2016-04-29
providers/gce: parse the boolean returned

crawford:
Despite the documentation's insistence that the API would return "TRUE",
it returned "true".


crawford:
Hoping this isn't too permissive...

PR #18, opened 2016-06-10
Switch to using dbus for dhcp lease info

ajeddeloh:
Use dbus for obtaining dhcp lease info instead of parsing files in /run/systemd/netif/leases/*. 
Depends on https://github.com/coreos/systemd/pull/50

### DIFF
--- a/internal/providers/azure/azure.go
+++ b/internal/providers/azure/azure.go
@@ -15,17 +15,17 @@
 package azure
 
 import (
-	"bufio"
 	"encoding/xml"
 	"fmt"
 	"net"
-	"os"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/coreos/coreos-metadata/internal/providers"
 	"github.com/coreos/coreos-metadata/internal/retry"
+
+	"github.com/godbus/dbus"
+	"github.com/godbus/dbus/introspect"
 )
 
 const (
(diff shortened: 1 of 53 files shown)


internal/providers/azure/azure.go
```diff
@@ -76,57 +76,68 @@ func getClient() retry.Client {
 	return client
 }
 
-func findLease() (*os.File, error) {
+/* godbus doesn't have a way to escape things. This will escape leading numbers, which is all we need */
+func escapeNumber(num int) string {
+	return "_3" + strconv.FormatUint(uint64(num), 10)
```
crawford, line 81: This only works where `num` is [0, 10), right?

ajeddeloh, line 81: Should work for all numbers, only leading numbers need to be escaped. E.g. 50 would become "_350"

crawford, line 81: Ah, gotcha.

internal/providers/azure/azure.go
```diff
@@ -76,57 +76,68 @@ func getClient() retry.Client {
...
+		if !hasLease {
+			continue
 		}
-	}
 
-	if len(rawEndpoint) == 0 || len(rawEndpoint) != 8 {
-		return nil, fmt.Errorf("fabric endpoint not found in leases")
-	}
+		//then get the data from the lease, if we found it
+		leaseObjPath := dbus.ObjectPath("/org/freedesktop/network1/link/" + escapeNumber(link.Index) + "/lease")
```
crawford: Can you reuse the linkObjPath above and append "/lease"?

ajeddeloh: Fixed

ajeddeloh:
Bump (thought I did this a day or two ago, guess I hadn't)

cgonyeo:
Is this still something that should get in? I can review if so.

ajeddeloh:
No, it relies on patches to systemd that were rejected upstream because they had other plans for a networkd dbus interface (see https://github.com/systemd/systemd/pull/4435 and https://github.com/systemd/systemd/pull/3528).

cgonyeo:
Should this be closed then?

PR #21, opened 2016-06-21
[WIP] Add initial support for the openstack OEM

dm0-:
This is a first-pass at supporting OpenStack via the EC2-compatible configdrive CD.  The network metadata service will be added eventually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,5 +22,11 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_GCE_HOSTNAME
       - COREOS_GCE_IP_EXTERNAL_0
       - COREOS_GCE_IP_LOCAL_0
+  - openstack (EC2 configdrive compatibility)
+    - SSH Keys
+    - Attributes
+      - COREOS_EC2_HOSTNAME
+      - COREOS_EC2_IPV4_LOCAL
+      - COREOS_EC2_IPV4_PUBLIC
 
 [ignition]: https://github.com/coreos/ignition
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -22,5 +22,11 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_GCE_HOSTNAME
       - COREOS_GCE_IP_EXTERNAL_0
       - COREOS_GCE_IP_LOCAL_0
+  - openstack (EC2 configdrive compatibility)
+    - SSH Keys
+    - Attributes
+      - COREOS_EC2_HOSTNAME
```
crawford, line 28: `COREOS_OPENSTACK_HOSTNAME`

internal/providers/configdrive/configdrive.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net"
+
+	"github.com/coreos/coreos-metadata/internal/providers"
+)
+
+var jsonmap map[string]interface{}
```
crawford: Global variables should be avoided at almost all cost. This one is pretty easy to keep local.

s-urbaniak:
@crawford Is there a possibility to priorize this? Currently we are working around this like so: https://github.com/coreos-inc/tectonic-platform-sdk/blob/master/platform-openstack-novanet/ignition-etcd.tf#L19

sdemos:
this pr is now stale because of #62, which rewrites coreos-metadata in rust. If this is still a metadata source we would like to support, it will have to be reimplemented in rust.

PR #22, opened 2016-07-18
providers: Add support for Packet

bison:
Adding support for Packet.

I chose not to vendor [packngo](https://github.com/packethost/packngo) and instead just have some structs with the fields I need, but I'm happy to change that.

Also, I don't think there's really a notion of a primary public or private address, so I just grab the first of each in the list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,4 +23,12 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_GCE_IP_EXTERNAL_0
       - COREOS_GCE_IP_LOCAL_0
 
+  - packet
+    - SSH Keys
+    - Attributes
+      - COREOS_PACKET_HOSTNAME
+      - COREOS_PACKET_IPV4_PUBLIC_0
+      - COREOS_PACKET_IPV4_PRIVATE_0
+      - COREOS_PACKET_IPV6_PUBLIC_0
+
 [ignition]: https://github.com/coreos/ignition
--- a/internal/providers/packet/packet.go
+++ b/internal/providers/packet/packet.go
@@ -1,0 +1,99 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package packet
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/coreos/coreos-metadata/internal/providers"
+	"github.com/coreos/coreos-metadata/internal/retry"
+	"github.com/packethost/packngo/metadata"
+)
+
+func FetchMetadata() (providers.Metadata, error) {
+	body, err := retry.Client{
+		InitialBackoff: time.Second,
+		MaxBackoff:     time.Second * 5,
+		MaxAttempts:    10,
+	}.Get(metadata.BaseURL + "/metadata")
+	if err != nil {
+		return providers.Metadata{}, err
+	}
+
+	var data struct {
+		Error string `json:"error"`
+		*metadata.CurrentDevice
+	}
+
+	if err := json.Unmarshal(body, &data); err != nil {
+		return providers.Metadata{}, err
+	}
+
+	if data.Error != "" {
+		return providers.Metadata{}, errors.New(data.Error)
+	}
+
+	attrs := ipAddresses(data.Network)
+	attrs["PACKET_HOSTNAME"] = data.Hostname
+
+	return providers.Metadata{
+		Attributes: attrs,
+		SshKeys:    data.SSHKeys,
+	}, nil
+}
+
+func ipAddresses(network metadata.NetworkInfo) map[string]string {
+	var publicIPv4, privateIPv4, publicIPv6, privateIPv6 []net.IP
+
+	for _, addr := range network.Addresses {
+		switch {
+		case addr.Family == 4 && addr.Public:
+			publicIPv4 = append(publicIPv4, addr.Address)
+
+		case addr.Family == 4 && !addr.Public:
+			privateIPv4 = append(privateIPv4, addr.Address)
+
+		case addr.Family == 6 && addr.Public:
+			publicIPv6 = append(publicIPv6, addr.Address)
+
+		case addr.Family == 6 && !addr.Public:
+			privateIPv6 = append(privateIPv6, addr.Address)
+		}
+	}
+
+	addresses := make(map[string]string)
+
+	for i, ip := range publicIPv4 {
+		addresses[fmt.Sprintf("PACKET_IPV4_PUBLIC_%d", i)] = providers.String(ip)
+	}
+
+	for i, ip := range privateIPv4 {
+		addresses[fmt.Sprintf("PACKET_IPV4_PRIVATE_%d", i)] = providers.String(ip)
+	}
+
+	for i, ip := range publicIPv6 {
+		addresses[fmt.Sprintf("PACKET_IPV6_PUBLIC_%d", i)] = providers.String(ip)
+	}
+
+	for i, ip := range privateIPv6 {
+		addresses[fmt.Sprintf("PACKET_IPV6_PRIVATE_%d", i)] = providers.String(ip)
+	}
+
+	return addresses
+}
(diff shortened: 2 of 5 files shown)


internal/providers/packet/packet.go
```diff
@@ -0,0 +1,138 @@
...
+			}
+
+			return ip, nil
+		}
+	}
+
+	return nil, errNotFound
+}
+
+func (p *packetMetadata) PublicIPv4() (net.IP, error) {
```
vcaputo: why isn't this a private method?

the others below as well..  isn't FetchMetadata() the only externally called method?

bison: Oops, yeah, there's no reason for that. Fixed.

internal/providers/packet/packet.go
```diff
@@ -0,0 +1,138 @@
+// Copyright 2015 CoreOS, Inc.
```
vcaputo: 2016

internal/providers/packet/packet.go
```diff
@@ -0,0 +1,138 @@
...
+			return ip, nil
+		}
+	}
+
+	return nil, errNotFound
+}
+
+func (p *packetMetadata) publicIPv4() (net.IP, error) {
+	ip, err := p.findIP(4, true)
+
```
vcaputo: Throughout this listing there are blank lines separating assignments from their error tests, which is inconsistent with the rest of the code in coreos-metadata.  Please remove these excess newlines.

internal/providers/packet/packet.go
```diff
@@ -0,0 +1,138 @@
...
+	ip, err := p.findIP(6, true)
+
+	if err == errNotFound {
+		return nil, nil
+	}
+
+	return ip, err
+}
+
+func FetchMetadata() (providers.Metadata, error) {
```
vcaputo: Another consistency thing: the other providers in coreos-metadata place the public FetchMetadata() at the top of the listing.  I actually prefer it this way, but being consistent is more important.

internal/providers/packet/packet.go
bison: Thanks. Fixed.

internal/providers/packet/packet.go
bison: Fixed.

internal/providers/packet/packet.go
bison: True. Fixed.

internal/providers/packet/packet.go
```diff
@@ -0,0 +1,130 @@
...
+			ip := net.ParseIP(addr.Address)
+			if ip == nil {
+				return nil, fmt.Errorf("couldn't parse %q as IP address", addr.Address)
+			}
+
+			return ip, nil
+		}
+	}
+
+	return nil, errNotFound
```
vcaputo: why not return nil, nil here, convert the below findIP callers to direct returns of findIP() and remove errNotFound altogether from this listing?

It's not clear to me what value that is adding other than additional verbosity

bison: Not needed. I changed that up and didn't finish removing the error.

Thanks. Fixed.

vcaputo:
lgtm, but this is @crawford's baby so I'll defer to him on the final stamp, particularly the choice to ad-hoc the packet structs vs. vendoring is debatable.

cgag:
Is the lack of a obvious primary address why the GCE IP attributes have the _0 suffix?

Maybe the word private in the attributes should be replaced with local to match the existing providers?  

This is the first time I've seen this project so these are as much questions as suggestions.

crawford:
When we wrote coreos-cloudinit, we attempted to map the network interfaces of different cloud providers to the same set of names (public and private IPv4). This works okay for simple deployments but completely falls apart as soon as you want to do something more than a demo.

With this project, I decided to toss all of that previous thinking and just make the names provider-specific. That's why the names are prefixed with the name of the provider and some providers have more entries than others.

If it's not obvious which interface is the primary, they should just be numbered instead (which is what GCE does) so that the user can make the choice.

bison:
@crawford the Packet metadata service only gives a list of addresses that each have a UUID and a boolean public attribute. Would you prefer something like this:

```
PACKET_IPV4_PUBLIC_UUID_0=3012704c-4dc1-11e6-bb29-cb949cdb635a
PACKET_IPV4_PUBLIC_ADDR_0=8.8.8.8
```

Or just `PACKET_ADDRESS_{index}_{attribute}` for all the attributes the metadata service provides?

There's a gist of the actual data [here](https://gist.github.com/bison/68dd53d11be46f39d6c346ea8b6c3bf8).

crawford:
Is that UUID predictable? If so, it might make sense to include that in the variable name.

I would just include the IP address. I don't think the other fields are particularly useful. We can add them later if they are needed though.

internal/providers/packet/packet.go
```diff
@@ -0,0 +1,112 @@
...
+	"encoding/json"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/coreos/coreos-metadata/internal/providers"
+	"github.com/coreos/coreos-metadata/internal/retry"
+)
+
+func FetchMetadata() (providers.Metadata, error) {
```
colhom, line 29: nit: put this func def below consts + structs in the file.

bison, line 29: I got conflicting feedback about that earlier because the style in existing providers is that this goes first. If we have a style guide, let me know...

bison:
@crawford the UUID is the same for the lifetime of the server as far as I know, if that's what you mean.

If the UUID is in the variable, how would the public / private status be encoded? Also, I think you can have dashes in environment variable names, but it seems pretty unusual.

internal/providers/packet/packet.go
colhom, line 29: Didn't know that. I personally find encountering const references before the definition to be annoying, but I've already wasted enough time talking about this. I would say precedent is the style guide ;)

bison:
Anyway, I don't have a strong opinion, I'm happy to format this however you want, but I think my vote would be for `PACKET_NETWORK_{index}_{attribute}` and maybe only include UUID, family, address, and public for now. So, something like:

```
PACKET_ADDRESS_0_UUID=5bc0cb8c-4ddd-11e6-853c-bfe22a816f1b
PACKET_ADDRESS_0_PUBLIC=true
PACKET_ADDRESS_0_FAMILY=4
PACKET_ADDRESS_0_ADDRESS=8.8.8.8
```

Maybe sub `PACKET_NETWORK_` for `PACKET_ADDRESS_`...

Let me know, and I'll get this updated.

internal/providers/packet/packet.go
jzelinskie, line 29: General Go style is to do what @colhom said. I'd be surprised if golint didn't complain about it.

bison, line 29: Yeah, that's definitely the norm. I'm happy to change it. I'll leave it up to @crawford I guess?

The only thing golint complains about here is the lack of comment for the exported function.

betawaffle:
The UUID isn't _super_ useful, it's just the ID of the address assignment, so it won't change for any of the non-elastic IPs.

bison:
Yeah, and on second thought, we probably want the family and public / private status encoded in the variable name. I think the most common use-case is "give me the first public IPv4 address".

So, maybe just `PACKET_IPV4_PUBLIC_<index>` and similar for private and the IPv6 addresses. Does that work?

betawaffle:
Yep, I think so.

crawford:
Sounds good.

betawaffle:
So, I noticed a potential bug in our API the other day. We might not be always returning the correct _first_ IP first. We'll want to confirm that before this goes out.

bison:
Updated. Let me know if there's anything else. Happy to vendor packngo if that's preferred.

README.md
```diff
@@ -23,4 +23,12 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_GCE_IP_EXTERNAL_0
       - COREOS_GCE_IP_LOCAL_0
 
+  - packet
+    - SSH Keys
+    - Attributes
+      - COREOS_PACKET_HOSTNAME
+      - COREOS_PACKET_IPV4_PUBLIC
```
crawford: Can you update these?

crawford:
Let's go ahead and vendor packngo. That way it's obvious who owns the structures.

bison:
Vendored the metadata subpackage from packngo and updated the provider to use that, though it still uses our client to fetch the metadata so that we get retries. Also updated the README.

crawford:
LGTM

internal/providers/packet/packet.go
vcaputo, line 29: Resurrecting this from the dead, when I suggested moving the public function to the top of the listing I didn't mean to move it above the consts and structs etc, just before the private functions, which I think is consistent with everything else in this project, no?

bison, line 29: Ah, I misunderstood. We ended up vendoring packngo and changing this around anyway, so it's all good I think.

PR #24, opened 2016-08-23
providers: add support for digitalocean

crawford:


internal/main.go
```diff
@@ -96,6 +99,11 @@ func main() {
 		fmt.Fprintf(os.Stderr, "failed to write metadata keys: %v\n", err)
 		os.Exit(1)
 	}
+
+	if err := writeNetworkUnits(flags.networkUnits, metadata); err != nil {
```
vcaputo: At a glance, this stuf in main.go outside of adding "digitalocean" should be at least in a distinct commit, since it's not necessarily digitalocean-specific though that seems to be the only application as of this PR...

crawford: Fair

mischief:
lgtm

PR #31, opened 2016-12-07
internal/vendor/update-ssh-keys: remove vim swap files

torstehu:


crawford:
Woops. _That's_ embarrassing :)

PR #32, opened 2017-03-07
providers: add initial sakuracloud support

higebu:
Related to: https://github.com/coreos/scripts/pull/654

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,5 +46,12 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_OPENSTACK_IPV4_LOCAL
       - COREOS_OPENSTACK_IPV4_PUBLIC
       - COREOS_OPENSTACK_INSTANCE_ID
+  - sakuracloud
+    - SSH Keys
+    - Network Configs
+    - Attributes
+      - COREOS_SAKURACLOUD_HOSTNAME
+      - COREOS_SAKURACLOUD_IPV4_0
+      - COREOS_SAKURACLOUD_IPV6_0
 
 [ignition]: https://github.com/coreos/ignition
--- a/internal/main.go
+++ b/internal/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/coreos/coreos-metadata/internal/providers/gce"
 	"github.com/coreos/coreos-metadata/internal/providers/openstackMetadata"
 	"github.com/coreos/coreos-metadata/internal/providers/packet"
+	"github.com/coreos/coreos-metadata/internal/providers/sakuracloud"
 
 	"github.com/coreos/update-ssh-keys/authorized_keys_d"
 )
@@ -81,7 +82,7 @@ func main() {
 	}
 
 	switch flags.provider {
-	case "azure", "digitalocean", "ec2", "gce", "packet", "openstack-metadata":
+	case "azure", "digitalocean", "ec2", "gce", "packet", "openstack-metadata", "sakuracloud":
 	default:
 		fmt.Fprintf(os.Stderr, "invalid provider %q\n", flags.provider)
 		os.Exit(2)
@@ -145,6 +146,8 @@ func fetchMetadata(provider string) (providers.Metadata, error) {
 		return packet.FetchMetadata()
 	case "openstack-metadata":
 		return openstackMetadata.FetchMetadata()
+	case "sakuracloud":
+		return sakuracloud.FetchMetadata()
 	default:
 		panic("bad provider")
 	}
--- a/internal/providers/sakuracloud/sakuracloud.go
+++ b/internal/providers/sakuracloud/sakuracloud.go
@@ -1,0 +1,226 @@
+// Copyright 2017 SAKURA Internet, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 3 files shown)


crawford:
The intent of this project is to provide a mechanism for reading dynamic metadata from the environments. I see that you are reading from a static file in `/usr/share/oem`, which would be created when we compile Container Linux. Does SAKURA have a dynamic metadata system for discovering things like IP addresses?

higebu:
@crawford SAKURA has a metadata system with modifying VM image before first boot.

crawford:
And this metadata system writes this data to the specified location? Is this information completely static (e.g. IP addresses won't change)?

higebu:
@crawford Yes, it writes the data to `/usr/share/oem/metadata.json` and this imformation is static for now.

cgonyeo:
This PR seems like it's convenience for using coreos-metadata to get the sakuracloud metadata into an easily `source`able form, instead of discovering externally-held information about the machine.

Just pointing it out, I'm not opposed. Means that users can use coreos-metadata on sakura the same way they could on the other supported cloud providers, even if the information is technically already on the machine.

crawford:
@higebu is there anywhere I can read more about this setup? I have some serious concerns around the usability of this metadata.

higebu:
@crawford We have no English document for disk modification feature, but Japanese version is here: https://help.sakura.ad.jp/hc/ja/articles/206227521-%E3%83%87%E3%82%A3%E3%82%B9%E3%82%AF%E4%BF%AE%E6%AD%A3
* This feature doesn't support CoreOS for now.
* It supports hostname, password, ssh key, network configuration, and "startup script".
* Network configuration supports first NIC only.
* "startup script" is kicked by our init script written into the disk before boot in the case of other OSes.
* In CoreOS, "startup script" will be used as ignition config by ignition. So users can use "startup script" as ignition config.

higebu:
@crawford How about this? Do you need more information?

crawford:
@higebu sorry, I've read through the docs but I haven't had time to review the code again. This week is going to be a little busy, but I'm hoping I'll have time at the end.

higebu:
@crawford No problem! Thanks!

cgonyeo:
I don't fully understand how sakuracloud injects the metadata file, how does it know which partition to put it on? Does it parse grub and then `/etc/fstab`, mount the filesystem, and write a normal file?

If it is just writing a file to the correct partition, would the file disappear when an os update happens, which reformats the `/usr` partition? You said this happens on first boot, would this need to happen on every boot? Alternatively when ignition runs on first boot, perhaps it needs to copy this file to somewhere outside of `/usr` to avoid this?

higebu:
@dgonyeo  `/usr/share/oem` is OEM partition and OEM partition is not rewritten by os update, right?

We write metadata file like this:
1. run gdisk command:

    ```
    $ gdisk -l coreos_production_sakuracloud_image.bin
    GPT fdisk (gdisk) version 1.0.1
    
    Partition table scan:
      MBR: hybrid
      BSD: not present
      APM: not present
      GPT: present
    
    Found valid GPT with hybrid MBR; using GPT.
    Disk coreos_production_sakuracloud_image.bin: 9289728 sectors, 4.4 GiB
    Logical sector size: 512 bytes
    Disk identifier (GUID): 00000000-0000-0000-0000-000000000001
    Partition table holds up to 128 entries
    First usable sector is 34, last usable sector is 9289694
    Partitions will be aligned on 2048-sector boundaries
    Total free space is 8125 sectors (4.0 MiB)
    
    Number  Start (sector)    End (sector)  Size       Code  Name
       1            4096          266239   128.0 MiB   EF00  EFI-SYSTEM
       2          266240          270335   2.0 MiB     EF02  BIOS-BOOT
       3          270336         2367487   1024.0 MiB  FFFF  USR-A
       4         2367488         4464639   1024.0 MiB  FFFF  USR-B
       6         4464640         4726783   128.0 MiB   8300  OEM
       7         4726784         4857855   64.0 MiB    FFFF  OEM-CONFIG
       9         4857856         9285631   2.1 GiB     FFFF  ROOT
    ```
2. number 6 partition is OEM partition.
3. mount OEM partition by mount command.
4. write metadata file.

cgonyeo:
Ah ok, this seems fine to me then. It'll need a rebase to be merge-able. @crawford any objections to me merging this?

crawford:
My primary concern is that the initial SSH keys added in this OEM metadata will continually be applied every boot. If an admin wants to remove access for a particular user, it will be re-added every boot. Does this cloud support modifying the metadata from a management console or does the admin need to manually edit the JSON file?

higebu:
@crawford The admin need to manually edit the JSON file. I'll document this issue on our support site.

higebu:
@crawford How about this p-r?

lucab:
I'm closing this stale PR as it was targeted at previous codebase and doesn't apply anymore. I'm in general a bit skeptical about this static first-boot "inject keys from OEM disk" being part of the metadata service.

@bgilbert perhaps this fits in the scope of your ignition changes in https://github.com/coreos/ignition/pull/475?

bgilbert:
coreos/ignition#475 could be used to support SSH keys, etc., but that doesn't seem like a great idea. No other platform dynamically generates its own Ignition configs and runs them alongside the user ones. coreos-metadata is the component that translates platform-specific metadata to config files inside a Container Linux instance, and I don't think there's sufficient reason to change that role just because the platform writes its metadata to the OEM partition.

@crawford's concern is not a problem unique to SAKURA.  https://github.com/coreos/bugs/issues/2070 seems like a reasonable fix.

So, I think the general approach of this PR is reasonable. It would still need to be rewritten for the new Rust codebase, though.

PR #33, opened 2017-03-08
Openstack metadata service support

cgonyeo:
This commit adds support for the metadata service in openstack.

Tested in vexxhost.

cc @s-urbaniak

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,5 +39,12 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_PACKET_IPV4_PUBLIC_0
       - COREOS_PACKET_IPV4_PRIVATE_0
       - COREOS_PACKET_IPV6_PUBLIC_0
+  - openstack-metadata
+    - SSH Keys
+    - Attributes
+      - COREOS_OPENSTACK_HOSTNAME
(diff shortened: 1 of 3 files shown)


s-urbaniak:
from my side LGTM :+1:

cgonyeo:
Fixes https://github.com/coreos/bugs/issues/1850

crawford (changes requested):
This needs to expose two different providers: openstack-configdrive and openstack-metadata. It is not possible to determine which one to use at runtime. The user will ultimately need to tell us which one to use.

cgonyeo:
Just pushed changes that splits the provider into the two requested labels.

cgonyeo:
I don't have a way to test the config drive currently, so I just changed this PR to only add support for the openstack metadata service (I dropped @dm0-'s commit from this). At the moment it's all that @s-urbaniak needs. I've also updated the PR's title and description to reflect this.

cgonyeo:
Collapsed if statements into less lines in `internal/providers/openstackMetadata/metadata.go`, as per oob request from @crawford

crawford (approved):

PR #36, opened 2017-03-14
providers/packet: expose phone home url

crawford:
This is needed in order to build the full URL for the phone-home
endpoint. This endpoint is POSTed to after boot to indicate that the
instance is ready to go.


crawford:
Verified that this works.

PR #38, opened 2017-04-18
main: allow overriding the platform with an environment variable

cgonyeo:
This commit allows overriding the provider as specified via flags or
kernel commandline with the environment variable
COREOS_METADATA_PROVIDER_OVERRIDE. If this variable is set to a
non-empty value, its contents will be used for the provider regardless
of what other flags are set.

This is due to the request here: https://github.com/coreos/container-linux-config-transpiler/pull/58#pullrequestreview-33290010

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -40,8 +40,9 @@ var (
 )
 
 const (
-	cmdlinePath    = "/proc/cmdline"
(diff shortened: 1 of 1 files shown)


cgonyeo:
Closing because we can just implement this logic in the service file

PR #40, opened 2017-05-05
providers: added CloudStack as metadata provider

sgoeminn:
Co-Authored-By: Frank Maximus <frank.maximus@nuagenetworks.net>

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/coreos/coreos-metadata/internal/providers"
 	"github.com/coreos/coreos-metadata/internal/providers/azure"
+	"github.com/coreos/coreos-metadata/internal/providers/cloudStackConfigDrive"
 	"github.com/coreos/coreos-metadata/internal/providers/digitalocean"
 	"github.com/coreos/coreos-metadata/internal/providers/ec2"
 	"github.com/coreos/coreos-metadata/internal/providers/gce"
@@ -150,6 +151,8 @@ func getMetadataProvider(providerName string) (func() (providers.Metadata, error
 		return openstackMetadata.FetchMetadata, nil
 	case "vagrant-virtualbox":
 		return vagrant_virtualbox.FetchMetadata, nil
+	case "cloudstack-configdrive":
+		return cloudStackConfigDrive.FetchMetadata, nil
 	default:
 		return nil, ErrUnknownProvider
 	}
--- a/README.md
+++ b/README.md
@@ -50,5 +50,14 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_OPENSTACK_IPV4_LOCAL
       - COREOS_OPENSTACK_IPV4_PUBLIC
       - COREOS_OPENSTACK_INSTANCE_ID
+  - cloudstack-configdrive
+    - SSH Keys
+    - Attributes
+      - CLOUDSTACK_AVAILABILITY_ZONE
+      - CLOUDSTACK_INSTANCE_ID
+      - CLOUDSTACK_SERVICE_OFFERING
+      - CLOUDSTACK_CLOUD_IDENTIFIER
+      - CLOUDSTACK_LOCAL_HOSTNAME
+      - CLOUDSTACK_VM_ID
 
 [ignition]: https://github.com/coreos/ignition
--- a/internal/providers/cloudStackConfigDrive/cloudstack.go
+++ b/internal/providers/cloudStackConfigDrive/cloudstack.go
@@ -1,0 +1,169 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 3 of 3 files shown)


crawford:
I cannot find the config-drive mechanism documented for CloudStack. Is this specific to a particular CloudStack cloud provider?

sgoeminn:
Config Drive is a new feature that is currently available as a PR for CloudStack (aimed for CloudStack release 4.11). It is not tied to a specific provider. The PR can be found at the following url:  https://github.com/apache/cloudstack/pull/2097

The design document for this PR can be found here:
https://cwiki.apache.org/confluence/display/CLOUDSTACK/Using+ConfigDrive+for+Metadata%2C+Userdata+and+Password

internal/main.go
```diff
@@ -145,6 +146,8 @@ func fetchMetadata(provider string) (providers.Metadata, error) {
 		return packet.FetchMetadata()
 	case "openstack-metadata":
 		return openstackMetadata.FetchMetadata()
+	case "cloudstack":
```
crawford: Let's use `--cloudstack-configdrive` and name the package cloudStackConfigDrive. We'll be adding support for the network metadata service eventually. We can adopt a similar naming convention for OpenStack.

internal/providers/cloudstackMetadata/cloudstack.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func getPath(label string) (string, error) {
+	path := diskByLabelPath + label
+
+	if fileExists(path) {
+		return path, nil
+	}
+
+	out, err := exec.Command("blkid", "-l", "-t", "LABEL="+label, "-o", "device").Output()
```
crawford: This step doesn't seem necessary since udev will make sure `/dev/disk/by/label/` is populated.

internal/providers/cloudstackMetadata/cloudstack.go
```diff
@@ -0,0 +1,181 @@
...
+	return string(output), true, nil
+}
+
+func unmountConfigDrive(mnt string) {
+	defer os.Remove(mnt)
+	defer syscall.Unmount(mnt, 0)
+}
+
+func fetchAndSet(key, mnt string, attrKey string, attributes map[string]string) error {
+	val, ok, err := readFromConfigDrive(mnt, key)
```
crawford: Can we just use ReadFile directly here and check the result of [IsNotExist](https://golang.org/pkg/os/#IsNotExist)?

internal/providers/cloudstackMetadata/cloudstack.go
```diff
@@ -0,0 +1,181 @@
...
+		return nil, err
+	}
+	if !ok || keysListBlob == "" {
+		return nil, nil
+	}
+	keysList := strings.Split(keysListBlob, "\n")
+
+	var keys []string
+
+	for _, key := range keysList {
```
crawford: Why is this copy necessary?

crawford (changes requested):

internal/main.go
sgoeminn: We changed it.

internal/providers/cloudstackMetadata/cloudstack.go
sgoeminn: We added it as a backup. But normally it should not be needed indeed, so we removed it again.

internal/providers/cloudstackMetadata/cloudstack.go
sgoeminn: We could. That's what is currently happening in the readFromConfigDrive function. However, I personally prefer to leave the code that is reading and checking if the file exist to the  readFromConfigDrive function instead of putting this checks in both the fetchAndSet and fetchKeys functions.

internal/providers/cloudstackMetadata/cloudstack.go
sgoeminn: It was not (it's removed).

crawford (approved):
Overall, this looks fine to me. Can you update the commit title to be lowercase?

sgoeminn:
Done, I also made [PR](https://github.com/coreos/container-linux-config-transpiler/pull/104) for the container-linux-config-transpiler. However, the downside of calling the metadata provider "CloudStackConfigDrive" is that the user who specifies the userdata needs to know which platform is used (OpenStack or CloudStack) and which services are present in the network (VR or ConfigDrive). Which might not always be the case.

crawford:
> However, the downside of calling the metadata provider "CloudStackConfigDrive" is that the user who specifies the userdata needs to know which platform is used (OpenStack or CloudStack) and which services are present in the network (VR or ConfigDrive). Which might not always be the case.

We are okay with that for now. I think it's reasonable to teach coreos-metadata how to take a non-pseudo provider (e.g. `cloudstack`) which will cause the tool to look at both metadata sources. Users who know exactly where the data is coming from can save time and avoid errors by specifying the pseudo provider (e.g. `cloudstack-configdrive`).

PR #41, opened 2017-05-10
providers/ec2: Fetch EC2 availability zone.

okushchenko:
Hi

This change is required in order to mount the AWS EFS endpoint correctly, which uses availability zone as a part of connection string.

I've tested this change on CoreOS 1353.7.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_EC2_HOSTNAME
       - COREOS_EC2_IPV4_LOCAL
       - COREOS_EC2_IPV4_PUBLIC
+      - COREOS_EC2_AVAILABILITY_ZONE
   - gce
     - SSH Keys
     - Attributes
--- a/internal/providers/ec2/ec2.go
+++ b/internal/providers/ec2/ec2.go
@@ -43,6 +43,10 @@ func FetchMetadata() (providers.Metadata, error) {
 	if err != nil {
 		return providers.Metadata{}, err
 	}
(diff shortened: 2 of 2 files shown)


cgonyeo:
This change results in coreos-metadata now fetching more information than just hostname and network addresses, so I'm curious what your motivation is to include fetching this information here as opposed to in some other application.

If you could include a quick example on how you're using it, that would be great.

okushchenko:
In order to mount the AWS EFS volume I need to create a systemd unit like this (variable interpolations are from terraform)
```
[Mount]
What=us-east-1a.fs-${var.efs_id}.efs.${var.region}.amazonaws.com:/
Where=/mnt/efs-galaxy
Type=nfs4
Options=nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2

[Install]
WantedBy=remote-fs.target
```

The AWS EFS connection string `us-east-1a.fs-${var.efs_id}.efs.${var.region}.amazonaws.com` contains the `us-east-1a` part which is an availability zone. It is required in order to mount the AWS EFS volume correctly. I don't know the AZ before boot because the instance is in the autoscaling group. Therefore, the AZ should be fetched from AWS metadata on boot.

cgonyeo:
@crawford thoughts on this? Are we fine with extra metadata sliding in for EC2 specifically or do we want to keep the metadata available more consistent across platforms?

crawford (approved):

crawford:
This is exactly what coreos-metadata is designed to do. All of the metadata is platform-specific so there's no sense in trying to keep the scope uniform between them.

PR #42, opened 2017-05-10
news: add news for v0.9.0

cgonyeo:


crawford:
This is not a bugfix so it should bump the minor field.

cgonyeo:
Changed to be 0.9.0

lucab (approved):

lucab:
LGTM. I don't see any other explicit blocker in here, but perhaps I'm missing some implicit state. Should we tag a release here to target next CL-alpha?

cgonyeo:
Yup, we should. Doing that now.

cgonyeo:
https://github.com/coreos/coreos-overlay/pull/2569

PR #44, opened 2017-05-19
internal/providers/azure: add support for ssh keys

cgonyeo:
This PR adds support for fetching the user's ssh key in azure

Not ready for merge yet.

### DIFF
--- a/internal/providers/azure/azure.go
+++ b/internal/providers/azure/azure.go
@@ -18,12 +18,15 @@ import (
 	"bufio"
 	"encoding/xml"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
+	"path"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/coreos/coreos-metadata/internal/exec"
 	"github.com/coreos/coreos-metadata/internal/providers"
 	"github.com/coreos/coreos-metadata/internal/retry"
 )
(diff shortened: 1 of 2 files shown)


lucab:
I have to say I'm not super happy about having exec-wrappers around openssl and ssh-keygen and feeding them (plaintext) network input. Is there any reason for not using go crypto&codec primitives from stdlib?

internal/providers/azure/azure.go
```diff
@@ -230,3 +277,158 @@ func fetchSharedConfig(endpoint net.IP) (metadata, error) {
...
+	if err != nil {
+		return "", err
+	}
+	return pubkey, nil
+}
+
+// genCerts will generate a new 2048 bit rsa key and certificate for use in
+// certificate retrieval.
+func genCerts(keyFilePath, certFilePath string) error {
+	err := exec.ExecExpectZero("openssl", "req", "-x509", "-nodes", "-subj", "/CN=LinuxTransport", "-days", "32768", "-newkey", "rsa:2048", "-keyout", keyFilePath, "-out", certFilePath)
```
lucab, line 363: Are the common-name and expiration mandated somewhere? If so, please add a reference. If not, please use some meaningful cn and a shorter expiration (or leave a note why you are using such a large one, eg. early-boot time skews).

lucab (changes requested):

internal/providers/azure/azure.go
cgonyeo: I'm not sure if they're mandated, this command was directly copied from the waagent code: https://github.com/Azure/WALinuxAgent/blob/master/azurelinuxagent/common/utils/cryptutil.py#L30

I can play around with changing the CN and shortening the expiration, if it works with more reasonable values I'll change it to use them.

cgonyeo:
There isn't a strong reason for using exec-wrappers other than this is how the waagent works, and it took me less work to implement it here the same way it works in waagent than to work through how to do this in pure go. I wouldn't be opposed to spending the time on switching over to the golang stdlib, but I can't say how much time it would take me to figure that out.

crawford:
> it took me less work to implement it here the same way it works in waagent than to work through how to do this in pure go

I'm not interested in work being done unless it's being done correctly. Please investigate doing it properly.

PR #47, opened 2017-05-31
internal/providers/ec2: fix ssh keys

### LINKED ISSUE #1981: 1423.0.0 AWS no instance ssh key
# Issue Report #

### Container Linux Version ###

```1423.0.0```

### Environment ###

AWS `ami-a8c2d6ce`

### Expected Behavior ###

I can ssh to instance using the key specified in AWS

### Actual Behavior ###

Instance asks for a password.

### Reproduction Steps ###

Create a machine with with ignition config (From my experiments- it does not matter whats inside)
```
{
    "ignition": {
      "version": "2.0.0",
      "config": {}
    },
    "storage": {
      "files": [
      ]
    },
    "systemd": {
      "units": [
      ]
    },
    "networkd": {},
    "passwd": {}
  }
```
Try to ssh.

### Other Information ###

* Instance without any ignition or cloud config receives ssh keys correctly and I am able to connect to it.
* Works fine with `1409.1.0`
* Not working log - https://gist.github.com/monder/624c5de1369435f7b2065a446a6e3050
* Working log - https://gist.github.com/monder/cb2611556571bc4d331f0966db656add
* Ignition itself seems to work because when I put units/files in I can see that they're starting

cgonyeo:
A recent commit resulted in the ec2 fetching logic to need to fetch a
resource from the dynamic endpoint in the ec2 metadata service in
addition to the meta-data endpoint. This resulted in all calls to fetch
a resource now needing to specify an additional part of the url prefix.
The logic for fetching ssh keys was overlooked when this change was
made.

This commit adds this meta-data prefix to the ssh key fetching logic.

Tested on an ec2 vm.

Fixes https://github.com/coreos/bugs/issues/1981.

### DIFF
--- a/internal/providers/ec2/ec2.go
+++ b/internal/providers/ec2/ec2.go
@@ -122,7 +122,7 @@ func fetchIP(key string) (net.IP, error) {
 }
 
 func fetchSshKeys() ([]string, error) {
-	keydata, present, err := fetchString("public-keys")
+	keydata, present, err := fetchString("meta-data/public-keys")
 	if err != nil {
 		return nil, fmt.Errorf("error reading keys: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


lucab (approved):
LGTM

lucab:
On a related note, is there any reason why we pinned to `2009-04-04` instead of using `latest` (or at least some recent revision)?

cgonyeo:
I wouldn't know, that change predates me.

crawford:
@lucab we pinned to a version to make sure that we could predict the structure. If we used latest, it might break whenever AWS updated it. We should bump it to a newer revision after validating.

PR #48, opened 2017-06-01
internal: Simplify checking for supported providers and metadata function to call

joonas:
This duplication bothered me a bit, so I figured I'd submit a PR to get rid of it.

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -37,6 +38,8 @@ import (
 var (
 	version       = "was not built properly"
 	versionString = fmt.Sprintf("coreos-metadata %s", version)
+
+	ErrUnknownProvider = errors.New("unknown provider")
 )
 
 const (
@@ -80,14 +83,13 @@ func main() {
 		flags.provider = parseCmdline(args)
 	}
 
-	switch flags.provider {
-	case "azure", "digitalocean", "ec2", "gce", "packet", "openstack-metadata":
-	default:
+	metadataFn, err := getMetadataProvider(flags.provider)
+	if err != nil {
 		fmt.Fprintf(os.Stderr, "invalid provider %q\n", flags.provider)
 		os.Exit(2)
 	}
 
-	metadata, err := fetchMetadata(flags.provider)
+	metadata, err := metadataFn()
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "failed to fetch metadata: %v\n", err)
 		os.Exit(1)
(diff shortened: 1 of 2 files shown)


lucab:
Perhaps a better suggestion here: introduce a `metadataFetcher(provider string) (fetchFn, error)` which returns the provider-specific fetcher or an error for unknown ones, then either react to the error or invoke the returned function. While at it, please also introduce some unit-test covering positive and negative cases.

internal/main.go
```diff
@@ -80,14 +80,26 @@ func main() {
...
+	case "digitalocean":
+		metadataFn = digitalocean.FetchMetadata
+	case "ec2":
+		metadataFn = ec2.FetchMetadata
+	case "gce":
+		metadataFn = gce.FetchMetadata
+	case "packet":
+		metadataFn = packet.FetchMetadata
+	case "openstack-metadata":
+		metadataFn = openstackMetadata.FetchMetadata
```
lucab: I'd rather see this logic splits in its own testable function, bubbling-up errors. See the general comment for details.

lucab (changes requested):

joonas:
> Perhaps a better suggestion here: introduce a metadataFetcher(provider string) (fetchFn, error) which returns the provider-specific fetcher or an error for unknown ones, then either react to the error or invoke the returned function. While at it, please also introduce some unit-test covering positive and negative cases.

I went down this path at first, but felt that it was more than I should introduce in one go. However, I'm happy to take a pass at implementing something like this if that'd be the preferred approach.

Thanks for the review!

joonas:
@lucab - is this in line with what you were thinking?

internal/main.go
```diff
@@ -114,6 +114,25 @@ func main() {
...
+	case "ec2":
+		return ec2.FetchMetadata, nil
+	case "gce":
+		return gce.FetchMetadata, nil
+	case "packet":
+		return packet.FetchMetadata, nil
+	case "openstack-metadata":
+		return openstackMetadata.FetchMetadata, nil
+	default:
+		return nil, errors.New("unknown provider")
```
cgonyeo: This is a super tiny nit, but I prefer the pattern of doing a `var ErrUnknownProvider = ...` at the top of the file and returning it here. Makes it easier to expose known error values for tests and such.

cgonyeo (approved):
LGTM (Luca is current on PTO)

internal/main.go
joonas: Totally makes sense, just did that 👍

cgonyeo (approved):
LGTM, rebase this on master and I'll merge

joonas:
Cool, should be good now 👍

internal/main_test.go
```diff
@@ -0,0 +1,38 @@
...
+	}{
+		{
+			desc: "supported provider",
+			name: "digitalocean",
+			err:  nil,
+		},
+		{
+			desc: "unknown provider",
+			name: "not-supported",
+			err:  errors.New("unknown provider"),
```
crawford, line 23: @dgonyeo the point of that refactor was to make use of it here.

PR #51, opened 2017-06-26
providers/packet: Add support for generating network units

bgilbert:
The Packet provisioner is currently responsible for this.

There is a small change on DigitalOcean: network units are now created with priority 10 instead of 0 so other units can be sequenced ahead of them.  This doesn't affect the stock configuration.

### DIFF
--- a/internal/providers/packet/packet.go
+++ b/internal/providers/packet/packet.go
@@ -50,7 +50,12 @@ func FetchMetadata() (providers.Metadata, error) {
 		return providers.Metadata{}, errors.New(data.Error)
 	}
 
-	attrs := ipAddresses(data.Network)
+	network, err := parseNetwork(data.Network)
+	if err != nil {
+		return providers.Metadata{}, fmt.Errorf("failed to parse network config from metadata: %v", err)
+	}
+
+	attrs := getNetworkAttrs(data.Network)
 	attrs["PACKET_HOSTNAME"] = data.Hostname
 	attrs["PACKET_PHONE_HOME_URL"] = data.PhoneHomeURL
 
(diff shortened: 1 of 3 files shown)


internal/providers/packet/packet.go
```diff
@@ -118,7 +119,33 @@ func parseNetwork(network metadata.NetworkInfo) ([]providers.NetworkInterface, e
...
+		{"Mode", network.BondingMode().String()},
+	}
+	if network.BondingMode() == metadata.BondingLACP {
+		attrs = append(attrs, [2]string{"LACPTransmitRate", "fast"})
+	}
+	netdevs := []providers.NetworkDevice{
+		{
+			Name:            bondDev,
+			Kind:            "bond",
+			HardwareAddress: ifaces[0].HardwareAddress,
```
cgonyeo, line 146: Won't the bond device be the last item in `ifaces`, not the first one?

internal/providers/packet/packet.go
```diff
@@ -69,6 +69,7 @@ func FetchMetadata() (providers.Metadata, error) {
 
 func parseNetwork(network metadata.NetworkInfo) ([]providers.NetworkInterface, error) {
 	ifaces := []providers.NetworkInterface{}
+	bondDev := "bond0"
```
cgonyeo: Minuscule nit: why not just do `Name: "bond"` further down? Making an extra variable for this seems unnecessary.

dm0- (approved):
This looks okay, and I agree with the config changes, but they are not backwards compatible (specifically, the `.network` matching patterns, names, and priorities.)  There were cases where users had to mask `00-.network` to override network settings.  Maybe @euank has an opinion if this is significant enough to warrant notification.

internal/providers/packet/packet.go
bgilbert: It acquires additional references in subsequent commits.

internal/providers/packet/packet.go
bgilbert: We want to clone the MAC address of the first real interface.

bgilbert:
@dm0- AFAICT the only other platform that uses coreos-metadata to write network units is DigitalOcean, and the priority should be the only change there.  I'll do a quick test to confirm the latter.

dm0-:
I thought I remember at least VMware also having issues with these `.network` files (and @crawford had fixes for it, so maybe he knows).

bgilbert:
VMware network units are still handled by coreos-cloudinit.

I've confirmed that the DigitalOcean unit is unchanged, except for the filename.

cgonyeo (approved):

bgilbert:
Rebased.

PR #52, opened 2017-06-30
providers: add vagrant_virtualbox provider

AlexNPavel:
This adds support for the vagrant_virtualbox provider (oem-id
vagrant-virtualbox) with variables VAGRANT_VIRTUALBOX_IPV4_PRIVATE
and VAGRANT_VIRTUALBOX_HOSTNAME

Depends on https://github.com/coreos/coreos-overlay/pull/2637

### DIFF
--- a/internal/providers/vagrant_virtualbox/vagrant_virtualbox.go
+++ b/internal/providers/vagrant_virtualbox/vagrant_virtualbox.go
@@ -1,0 +1,89 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vagrant_virtualbox
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"os"
+	"time"
+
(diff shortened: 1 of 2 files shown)


internal/providers/vagrant_virtualbox/vagrant_virtualbox.go
```diff
@@ -0,0 +1,82 @@
...
+	return config, nil
+}
+
+func getIP() (net.IP, error) {
+	var iface net.Interface
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return nil, err
+	}
+	for _, iface = range interfaces {
```
cgonyeo: Why define `iface` earlier in the function? It doesn't appear to be used outside of this for loop, so just use a `:=` here.

cgonyeo (changes requested):
I'm not familiar with vagrant at all, is there a way for users to pass ssh keys that we'll want to support?

AlexNPavel:
The only way for a user to pass in an ssh key is via the ignition config. There is also a vagrant specific way to replace the insecure key after ssh comes up, but we don't support that at the moment. Also, the way this metadata works is different since we don't read from a network host, we just read data from the system (which is why we need to configure everything via ignition first)

cgonyeo (approved):
LGTM, cc @crawford for design sanity-check of having all the metadata come from host introspection

AlexNPavel:
I added a bugfix that waits for eth1 to be ready and squash the commits. Look good @dgonyeo ?

cgonyeo:
Do you know if it's possible to configure a machine to not have a network interface? Is it possible that `eth1` would just never appear?

AlexNPavel:
With our Vagrantfile, there should always be an eth1. It would only not exist if someone removes that part of the Vagrantfile (in which case it would never appear)

cgonyeo:
That seems fine, even if they did that coreos-metadata with this commit will eventually still return successfully.

LGTM.

crawford:
@dgonyeo this didn't update the list of providers on the README.

PR #55, opened 2017-08-14
internal/main.go: support for overriding metadata

AlexNPavel:
This adds the flag `set` which will set the specified attributes or
override the one's given by a provider.

### DIFF
--- a/internal/main.go
+++ b/internal/main.go
@@ -81,7 +81,7 @@ func main() {
 	flag.StringVar(&flags.hostname, "hostname", "", "The file into which the hostname should be written")
 	flag.StringVar(&flags.networkUnits, "network-units", "", "The directory into which network units are written")
 	flag.StringVar(&flags.provider, "provider", "", "The name of the cloud provider")
-	flag.Var(&flags.overrides, "set", "Set an environment variable, overriding the default/detected variable")
+	flag.Var(&flags.overrides, "set", "Set a metadata attribute, potentially overriding the attribute given by a provider. This flag can be specified multiple times and expects a key=value pair. Eg: VIRTUALBOX_IPV4_PRIVATE=172.17.4.101")
 	flag.StringVar(&flags.sshKeys, "ssh-keys", "", "Update SSH keys for the given user")
 	flag.BoolVar(&flags.version, "version", false, "Print the version and exit")
 
--- a/internal/main.go
+++ b/internal/main.go
@@ -55,10 +55,10 @@ func (sm *StringMapFlag) String() string {
 }
 
 func (sm *StringMapFlag) Set(v string) error {
-	vals := strings.Split(v, "=")
+	vals := strings.SplitN(v, "=", 1)
 	if len(vals) != 2 {
-		fmt.Println("Could not split override set correctly")
-		return nil
+		fmt.Printf("Incorrectly formatted -set argument: %s\n", v)
+		return fmt.Errorf("Incorrectly formatted -set argument: %s", v)
 	}
 	(*sm)[vals[0]] = vals[1]
 	return nil
--- a/internal/main.go
+++ b/internal/main.go
@@ -81,9 +81,9 @@ func main() {
 	flag.StringVar(&flags.hostname, "hostname", "", "The file into which the hostname should be written")
 	flag.StringVar(&flags.networkUnits, "network-units", "", "The directory into which network units are written")
 	flag.StringVar(&flags.provider, "provider", "", "The name of the cloud provider")
+	flag.Var(&flags.overrides, "set", "Set an environment variable, overriding the default/detected variable")
 	flag.StringVar(&flags.sshKeys, "ssh-keys", "", "Update SSH keys for the given user")
 	flag.BoolVar(&flags.version, "version", false, "Print the version and exit")
-	flag.Var(&flags.overrides, "set", "Set an environment variable, overriding the default/detected variable")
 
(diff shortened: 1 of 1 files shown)


internal/main.go
```diff
@@ -66,6 +83,7 @@ func main() {
 	flag.StringVar(&flags.provider, "provider", "", "The name of the cloud provider")
 	flag.StringVar(&flags.sshKeys, "ssh-keys", "", "Update SSH keys for the given user")
 	flag.BoolVar(&flags.version, "version", false, "Print the version and exit")
+	flag.Var(&flags.overrides, "set", "Set an environment variable, overriding the default/detected variable")
```
bgilbert: - Should be in alphabetical order.
- The `attributes` help text calls them `metadata attributes`, so we should be consistent here.
- The help text should make clear that a key=value pair is wanted and that the option can be specified multiple times.

internal/main.go
```diff
@@ -48,6 +48,22 @@ const (
...
+type StringMapFlag map[string]string
+
+func (sm *StringMapFlag) String() string {
+	return fmt.Sprintf("%+v", *sm)
+}
+
+func (sm *StringMapFlag) Set(v string) error {
+	vals := strings.Split(v, "=")
+	if len(vals) != 2 {
+		fmt.Println("Could not split override set correctly")
```
bgilbert: This is potentially a user-visible error message and should clearly describe the problem.

internal/main.go
```diff
@@ -48,6 +48,22 @@ const (
...
+		return nil
```
bgilbert: Don't we want to fail in this case?

internal/main.go
```diff
@@ -48,6 +48,22 @@ const (
...
 )
 
+type StringMapFlag map[string]string
+
+func (sm *StringMapFlag) String() string {
+	return fmt.Sprintf("%+v", *sm)
+}
+
+func (sm *StringMapFlag) Set(v string) error {
+	vals := strings.Split(v, "=")
```
bgilbert: Values should be allowed to contain `=`, so this should be a `SplitN`.

bgilbert (changes requested):
@crawford @dgonyeo, thoughts on whether this is a good idea?  The motivator is to allow Vagrant to generate its own attributes (e.g. private IP address) that will be merged into whatever coreos-metadata produces on its own.

cgonyeo:
A couple questions:
- How would this be used? Would the user have to add a systemd drop-in to add the arguments to coreos-metadata via an environment variable or something?
- If this is vagrant specific, why not put the logic for getting information about the environment into a vagrant provider and have coreos-metadata run normally, instead of entering in the information on the command line only to be spit back out?

bgilbert:
1. Yes.
2. We're trying to get rid of the Vagrant-specific OEMs (and just use e.g. the `virtualbox` and `vmware` OEMs), since the Vagrant-specific ones conflate the virtualization environment with the tool driving it.  The generic OEMs can't know when they should e.g. wait for a second NIC to come up, and it seems too hacky to e.g. add a `-vagrant` option which causes special behavior in particular OEMs.

This would also allow kola to inject public and private IPs for ESX, assuming coreos-metadata gained a vmware OEM (which may or may not make sense).

bgilbert:
Still missing the last two comments from https://github.com/coreos/coreos-metadata/pull/55#discussion_r133076632.

internal/main.go
```diff
@@ -57,13 +73,15 @@ func main() {
...
-	}{}
+		overrides    StringMapFlag
+	}{overrides: make(map[string]string)}
 
 	flag.StringVar(&flags.attributes, "attributes", "", "The file into which the metadata attributes are written")
 	flag.BoolVar(&flags.cmdline, "cmdline", false, "Read the cloud provider from the kernel cmdline")
 	flag.StringVar(&flags.hostname, "hostname", "", "The file into which the hostname should be written")
 	flag.StringVar(&flags.networkUnits, "network-units", "", "The directory into which network units are written")
 	flag.StringVar(&flags.provider, "provider", "", "The name of the cloud provider")
+	flag.Var(&flags.overrides, "set", "Set a metadata attribute, potentially overriding the attribute given by a provider. This flag can be specified multiple times and expects a key=value pair. Eg: VIRTUALBOX_IPV4_PRIVATE=172.17.4.101")
```
bgilbert: tiny nit: should be `e.g.`

internal/main.go
```diff
@@ -48,6 +48,22 @@ const (
...
+
+func (sm *StringMapFlag) String() string {
+	return fmt.Sprintf("%+v", *sm)
+}
+
+func (sm *StringMapFlag) Set(v string) error {
+	vals := strings.SplitN(v, "=", 1)
+	if len(vals) != 2 {
+		fmt.Printf("Incorrectly formatted -set argument: %s\n", v)
+		return fmt.Errorf("Incorrectly formatted -set argument: %s", v)
```
bgilbert: I get:

```
$ bin/coreos-metadata -set foo
Incorrectly formatted -set argument: foo
invalid value "foo" for flag -set: Incorrectly formatted -set argument: foo
Usage of bin/coreos-metadata:
[...]
```

bgilbert:
Discussed with @dgonyeo offline.  To avoid Vagrant etc. having to inject systemd dropins, the proposed approach is now to have coreos-metadata walk one or more directories containing files with key=value pairs, where the directory paths are possibly provided via command-line options.

PR #56, opened 2017-08-14
providers: rename vagrant-virtualbox to virtualbox

AlexNPavel:
This allows us to avoid using specific oem-ids for vagrant provisioned
machines and use more ids

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,6 @@ The supported cloud providers and their respective metadata are as follows:
   - virtualbox
     - Attributes
       - COREOS_VIRTUALBOX_HOSTNAME
-      - COREOS_VIRTUALBOX_IPV4 (only via vagrant)
+      - COREOS_VIRTUALBOX_IPV4_PRIVATE (only via vagrant)
 
 [ignition]: https://github.com/coreos/ignition
--- a/README.md
+++ b/README.md
@@ -50,5 +50,9 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_OPENSTACK_IPV4_LOCAL
       - COREOS_OPENSTACK_IPV4_PUBLIC
       - COREOS_OPENSTACK_INSTANCE_ID
+  - virtualbox
+    - Attributes
+      - COREOS_VIRTUALBOX_HOSTNAME
+      - COREOS_VIRTUALBOX_IPV4 (only via vagrant)
 
 [ignition]: https://github.com/coreos/ignition
--- a/internal/main.go
+++ b/internal/main.go
@@ -148,7 +148,7 @@ func getMetadataProvider(providerName string) (func() (providers.Metadata, error
 		return packet.FetchMetadata, nil
 	case "openstack-metadata":
 		return openstackMetadata.FetchMetadata, nil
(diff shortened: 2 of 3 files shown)


internal/main.go
```diff
@@ -148,8 +148,8 @@ func getMetadataProvider(providerName string) (func() (providers.Metadata, error
 		return packet.FetchMetadata, nil
 	case "openstack-metadata":
 		return openstackMetadata.FetchMetadata, nil
-	case "vagrant-virtualbox":
-		return vagrant_virtualbox.FetchMetadata, nil
+	case "virtualbox":
```
cgonyeo: Can you also add `"vagrant-virtualbox"` to this line? Otherwise won't coreos-metadata will stop working for users when they upgrade to this version?

cgonyeo (changes requested):

README.md
```diff
@@ -50,5 +50,9 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_OPENSTACK_IPV4_LOCAL
       - COREOS_OPENSTACK_IPV4_PUBLIC
       - COREOS_OPENSTACK_INSTANCE_ID
+  - virtualbox
+    - Attributes
+      - COREOS_VIRTUALBOX_HOSTNAME
+      - COREOS_VIRTUALBOX_IPV4 (only via vagrant)
```
bgilbert: Will we want to distinguish between public and private IPv4?

internal/main.go
```diff
@@ -148,8 +148,8 @@ func getMetadataProvider(providerName string) (func() (providers.Metadata, error
 		return packet.FetchMetadata, nil
 	case "openstack-metadata":
 		return openstackMetadata.FetchMetadata, nil
-	case "vagrant-virtualbox":
-		return vagrant_virtualbox.FetchMetadata, nil
+	case "virtualbox", "vagrant-virtualbox":
```
bgilbert: Needs a comment explaining the second arm.

README.md
crawford: What happens if there are multiple network interfaces?

AlexNPavel: @crawford I meant that to be the private ipv4. Fixed

README.md
```diff
@@ -50,5 +50,9 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_OPENSTACK_IPV4_LOCAL
       - COREOS_OPENSTACK_IPV4_PUBLIC
       - COREOS_OPENSTACK_INSTANCE_ID
+  - virtualbox
+    - Attributes
+      - COREOS_VIRTUALBOX_HOSTNAME
+      - COREOS_VIRTUALBOX_IPV4_PRIVATE (only via vagrant)
```
crawford, line 56: > only via vagrant

This alone seems to indicate that the platform should incorporate both virtualbox and vagrant.

sdemos:
there doesn't seem to be consensus on this rename, and it is now stale. if we still consider this a problem with the rust rewrite, we should open a new pr.

bgilbert:
There is a consensus, I just haven't written it up or implemented it.  I'll reopen on the other project.

PR #57, opened 2017-08-29
internal/providers/oracleoci: initial implementation

cgonyeo:
Adds an initial implementation for the oracle-oci provider.

Tested attributes, ssh keys, and hostname manually on an oracle VM.

Fixes https://coreosdev.atlassian.net/projects/OST/issues/OST-40?filter=allopenissues

cc @bgilbert

### DIFF
--- a/internal/providers/oracleoci/oracleoci.go
+++ b/internal/providers/oracleoci/oracleoci.go
@@ -1,0 +1,65 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package oracleoci
+
+import (
+	"encoding/json"
+	"strings"
+	"time"
+
+	"github.com/coreos/coreos-metadata/internal/providers"
+	"github.com/coreos/coreos-metadata/internal/retry"
+)
+
+type instanceData struct {
+	AvailabilityDomain string   `json:"availabilityDomain"`
+	CompartmentId      string   `json:"compartmentId"`
+	DisplayName        string   `json:"displayName"`
+	Id                 string   `json:"id"`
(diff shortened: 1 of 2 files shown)


bgilbert:
This will eventually need an `-sshkeys` unit added to Ignition's `oem.go`.

Do we plan to parse IP addresses out of the lease data?

cgonyeo:
I've got the code written to do that if we want, the information isn't available on the metadata server.

@crawford do you have thoughts on which is better: parsing lease data for IP addresses, or not including the information?

crawford:
I would not include it unless we know for certain. It sounds like we won't be able to know which interfaces are which and what the public IPs are going to be, which makes me think we should hold off for now and add the functionality once the platform supports it.

internal/providers/oracleoci/oracleoci.go
```diff
@@ -0,0 +1,64 @@
...
+	var data instanceData
+	err = json.Unmarshal(instanceDataBlob, &data)
+	if err != nil {
+		return providers.Metadata{}, err
+	}
+	return providers.Metadata{
+		Attributes: map[string]string{
+			"ORACLE_OCI_HOSTNAME": data.DisplayName,
+		},
+		Hostname: data.DisplayName,
```
bgilbert: OCI provides separate fields for the display name and the hostname.  (`scripts/oem/oracle/launch-instance` currently sets both to the same value.)  It appears the only way to get the hostname is via reverse lookup of the private IP.  There can be a separate hostname for each VNIC, but we're probably only interested in the first one.

bgilbert: In particular, the display name can be up to 255 characters and can contain e.g. spaces.

cgonyeo:
After talking with @crawford we're going to not set the hostname here and just leave it as `localhost` as a result. If the information ever becomes available via DHCP or the metadata server, then we can set it.

cgonyeo:
I've also added a couple more things that seemed like they could be useful to the attributes.

internal/providers/oracleoci/oracleoci.go
```diff
@@ -0,0 +1,65 @@
...
+
+	var data instanceData
+	err = json.Unmarshal(instanceDataBlob, &data)
+	if err != nil {
+		return providers.Metadata{}, err
+	}
+	return providers.Metadata{
+		Attributes: map[string]string{
+			"ORACLE_OCI_DISPLAY_NAME": data.DisplayName,
+			"ORACLE_OCI_ID":           data.Id,
```
bgilbert: `ORACLE_OCI_INSTANCE_ID`?

crawford (approved):

PR #60, opened 2017-09-20
providers/cloudstack: support for network metadata

larstobi:
This adds network metadata in addition to config drive metadata that
is not supported on all CloudStack versions.

I have tested this on our CloudStack 4.6 / XenServer 6.5.0 rig, and it works.

@crawford Would you mind reviewing this?


crawford:
> This adds network metadata in addition to config drive metadata that is not supported on all CloudStack versions.

Does this mean that config-drive is not supported on all CloudStack versions or network metadata is not supported on all CloudStack version? If it's the latter, I think we will want to rename the provider to "cloudstack-metadata". Otherwise, Container Linux will attempt to read from the network metadata by default (which may not work).

larstobi:
@crawford 
I do believe that network metadata is supported on all cloudstack versions. However, configdrive is not.

Ref. https://issues.apache.org/jira/browse/CLOUDSTACK-9813 the issue is not closed yet, and I am not aware of any released versions with configdrive available. It is also not mentioned in the latest docs, as far as I can see: http://cloudstack.apache.org/api/apidocs-4.10/apis/deployVirtualMachine.html

I don't surely know what the coreos.oem.id in /proc/cmdine will say when there is configured a configdrive, but looking at the code, it seems like it will say "cloudstack-configdrive", and not "cloudstack", https://github.com/coreos/coreos-metadata/blob/master/internal/main.go#L155 

Which, I think, means that there is no conflict here. I might be wrong. Maybe @sgoeminn can shed some light on that?

crawford:
> I do believe that network metadata is supported on all cloudstack versions. However, configdrive is not.

Ah, great. My only concern is that users who are attempting to use config-drive may have to jump through extra hoops just to get it working. This might happen if the metadata and userdata only show up in either the network service or config-drive (this is how OpenStack sometimes works).

> I don't surely know what the coreos.oem.id in /proc/cmdine will say when there is configured a configdrive

The ID will always be "cloudstack". Because of that, the network metadata always needs to be available. If that's not the case, I think we should use two different pseudo-providers (cloudstack-configdrive and cloudstack-metadata) and let `ct` do the provider override.

larstobi:
@crawford
Hmm, if the ID will always be "cloudstack", then I'm unsure how the configdrive provide could ever work, as it looks for "cloudstack-configdrive" string.

If you don't mind terribly, would you be able to produce an image on your build servers with both Ignition with network metadata support and this patch, so that I can test it at boot time? I couldn't figure out how to replace the coreos-metadata binary with torcx or any other way, and building an image from scratch seems non-trivial.

crawford:
> Hmm, if the ID will always be "cloudstack", then I'm unsure how the configdrive provide could ever work, as it looks for "cloudstack-configdrive" string.

This is done in two stages. Ignition looks for `cloudstack` and will attempt to read from both locations (potentially slow). The Ignition config will then write a drop-in for `coreos-metadata.service` which overrides the provider to either `cloudstack-configdrive` or `cloudstack-metadata`. This override is generated by CT (which the user runs beforehand to generate a config) which uses the `--provider` flag (again, either `cloudstack-configdrive` or `cloudstack-metadata`). The reason coreos-metadata doesn't look at both locations is because that operation may be slow (as it's potentially waiting for an endpoint which will never appear). In the case of OpenStack, the information can also be wrong, which is why we must specify where that data should come from.

coreos-metadata can be replaced. You just need to write your new binary to some location (e.g. `/opt/bin/coreos-metadata`) and then override the `coreos-metadata.service` to exec that location instead of the one in `/usr`.

larstobi:
> The ID will always be "cloudstack". Because of that, the network metadata always needs to be available. If that's not the case, I think we should use two different pseudo-providers (cloudstack-configdrive and cloudstack-metadata) and let ct do the provider override.

The configdrive feature in cloudstack was written for the need to not run the virtual router that hosts the userdata and metadata for CloudStack. Then, network metadata will not be available. So, it looks like either configdrive or network metadata.

I made this alternative: https://github.com/larstobi/coreos-metadata/tree/cloudstack_network_metadata_second
Should I replace this PR with that code?

It also works:
```
disco-0 larstobi # ./coreos-metadata --cmdline --provider=cloudstack-metadata  --attributes=/run/metadata/coreos
Fetching "http://10.30.5.1/latest/meta-data/cloud-identifier": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/instance-id": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/local-ipv4": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/availability-zone": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/public-ipv4": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/service-offering": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/vm-id": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/local-hostname": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/public-hostname": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/local-hostname": Attempt #1
Fetching "http://10.30.5.1/latest/meta-data/public-keys": Attempt #1
```

We want to use the [Ignition provider](https://github.com/terraform-providers/terraform-provider-ignition) for Terraform to configure our coreos instances. I don't know if they run `ct` or not, but I don't think so. Is that important?

crawford:
> I made this alternative: https://github.com/larstobi/coreos-metadata/tree/cloudstack_network_metadata_second
Should I replace this PR with that code?

I think so. Just rename "cloudstack" to "cloudstack-metadata" in the README.

> We want to use the Ignition provider for Terraform to configure our coreos instances. I don't know if they run ct or not, but I don't think so. Is that important?

Nope. It just means you will need to create the `coreos-metadata.service` drop-in yourself:

```ini
[Service]
Environment=COREOS_METADATA_OPT_PROVIDER=--provider=cloudstack-metadata
```

larstobi:
@crawford Okay, done! :-)

larstobi:
@crawford I have updated the PR with requested changes. Does it look good? :-)

crawford (approved):

PR #62, opened 2017-10-02
*: rewrite coreos-metadata in rust

sdemos:
we've been reviewing it as we go and did a review of the project as a whole as well.


CONTRIBUTING.md
```diff
@@ -37,17 +44,23 @@ This is a rough outline of what a contributor's workflow looks like:
 
 Thanks for your contributions!
 
+### Coding Style
+
+CoreOS projects written in Go follow a set of style guidelines that we've documented
```
arithx: Doesn't seem necessary to link to the Go style guide when the entire projects in Rust.

sdemos: I just copied this from the template project repo without reading it. I'll update it.

sdemos: actually do we have a convenient place to link to for our rust coding style or should I just delete this section? 

/cc @crawford

crawford: We can just recommend that rustfmt be used.

CONTRIBUTING.md
```diff
@@ -44,11 +44,9 @@ This is a rough outline of what a contributor's workflow looks like:
 
 Thanks for your contributions!
 
-### Coding Style
+### Code Style
 
-CoreOS projects written in Go follow a set of style guidelines that we've documented
-[here](https://github.com/coreos/docs/tree/master/golang). Please follow them when
-working on your contributions.
+Code should be formatted according to the output of [rustfmt](https://github.com/rust-lang-nursery/rustfmt)
```
arithx: Can we add a `rustfmt` call to the Travis job similar to how we run `gofmt` on other projects?

sdemos: I'm going to add this in the future to both coreos-metadata and update-ssh-keys.

src/util/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    for l in contents.lines() {
+        let l = l?;
+        if let Some(v) = key_lookup_line(delim, key, &l) {
+            return Ok(Some(v));
+        }
+    }
+    Ok(None)
+}
+
+pub fn key_lookup(delim: char, key: &str, contents: &str) -> Option<String> {
```
dm0-, line 51: Building this results in warnings:

```
warning: function is never used: `key_lookup`
  --> src/util/mod.rs:51:1
   |
51 | / pub fn key_lookup(delim: char, key: &str, contents: &str) -> Option<String> {
52 | |     for l in contents.lines() {
53 | |         if let Some(v) = key_lookup_line(delim, key, l) {
54 | |             return Some(v)
...  |
57 | |     None
58 | | }
   | |_^
   |
   = note: #[warn(dead_code)] on by default
```

lucab: There are a few other clippy warnings as well. I think we can tackle that after merging and setting up travis-ci for it as well. I already did for the other crates we have, and I'm planning to do it here as well.

Cargo.toml
```diff
@@ -0,0 +1,45 @@
+[package]
+name = "coreos-metadata"
+version = "0.1.0"
```
dm0-: So it's mentioned here and not just in the overlay PR, this should also be updated to `1.0.0`.

dm0- (changes requested):

crawford (approved):

burdiyan:
Hi there! I wonder, what was the purpose of the rewrite?

crawford:
@burdiyan It was a technical one. In order to support fetching SSH keys from Azure, we needed crypto primitives that weren't available in Go (namely CMS). The added benefit of going to Rust is a smaller binary, smaller memory footprint, and less bugs, just to name a few.

burdiyan:
Thanks! Good to know.

PR #64, opened 2017-10-26
cargo: log less in release builds

sdemos:
the debug build is a very verbose logger, which is a good thing when you
are developing, but we don't want the release build to talk as loudly
about what it is doing.


lucab (approved):
LGTM

cgonyeo (approved):

PR #68, opened 2017-11-08
cargo: bump dependency versions

sdemos:
in particular, this bumps to the newest version of update-ssh-keys and
openssh-keys which fixes an issue with parsing authorized_keys files
(sdemos/openssh-keys#7)


bgilbert (approved):
That sure is a lot of version bumps.  LGTM.

PR #70, opened 2017-11-09
main: accept long args with one hyphen

### LINKED ISSUE #2240: coreos-metadata no longer accepts single-dash command line flags
# Issue Report #

### Container Linux Version ###

```
$ cat /etc/os-release 
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=1590.0.0
VERSION_ID=1590.0.0
BUILD_ID=2017-11-08-0831
PRETTY_NAME="Container Linux by CoreOS 1590.0.0 (Ladybug)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
COREOS_BOARD="amd64-usr"
```

### Environment ###

Bare metal on Packet.

### Expected Behavior ###

The coreos-metadata service starts with original systemd unit.

### Actual Behavior ###

During a recent alpha update, the coreos-metadata binary stopped accepting the golang style single dash command line arguments. When my hosts where deployed, they required a custom coreos-metadata systemd unit configured with Ignition, and that unit used the single dash arguments. The metadata service now fails to start, and will until either I update the Ignition and roll each host in the cluster or the binary accepts the old style flags.

sdemos:
the "go style" flags where flags can be accepted with either one or two
hyphens at the beginning needs to be supported, since we are trying to
match the original go interface as closely as possible. the solution is
to do a little preprocessing on the arguments that are passed, where if
a flag with a single hyphen is passed, we add another hyphen. this only
works because the flags we have defined don't have short arguments. we
would have to be more careful if short arguments are ever added.

fixes coreos/bugs#2240

as a separate note, looks like I didn't run `cargo build` after bumping the
 version in `Cargo.toml`, so the version number didn't get propogated to
 `Cargo.lock`. this also happens to fix that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,7 +193,7 @@ dependencies = [
 
 [[package]]
 name = "coreos-metadata"
-version = "1.0.1"
+version = "1.0.2"
 dependencies = [
  "base64 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/src/bin/coreos-metadata.rs
+++ b/src/bin/coreos-metadata.rs
@@ -28,6 +28,7 @@ extern crate slog_scope;
 
 extern crate coreos_metadata;
 
+use std::env;
 use std::fs::File;
 use std::io::prelude::*;
 use clap::{Arg, App};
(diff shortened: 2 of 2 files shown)


cgonyeo (approved):
Note: I think this'll have an issue if a user on the system whose ssh keys are being updated starts with a `-`.

e.g.
```
coreos-metadata -cmdline -ssh-keys -myuser
```

I don't know if that would have ever worked though.

sdemos:
we only accept usernames and files as arguments to any of our flags, and enough other things would break if those started with dashes that I don't think it will be too big of an issue. 

fwiw it would work with the go version, but clap already doesn't support having dashes there anyway unless you pass the flag as `--attributes=-file` (for instance, `coreos-metadata --attributes --file` breaks it)

PR #72, opened 2017-11-20
providers/digitalocean: private ip index bug fix

tfussell:
I just noticed that my private IP was empty in dependent systemctl services because I expected it to be set as `COREOS_DIGITALOCEAN_IPV4_PRIVATE_0` but `/run/metadata/coreos` named it `COREOS_DIGITALOCEAN_IPV4_PRIVATE_1`. I checked the code and it looks like a minor copy-and-paste error.

### DIFF
--- a/src/providers/digitalocean/mod.rs
+++ b/src/providers/digitalocean/mod.rs
@@ -121,7 +121,7 @@ fn parse_attrs(data: &Metadata) -> Result<Vec<(String,String)>> {
         for a in ifaces {
             if a.ipv4.is_some() {
                 attrs.push((
-                    format!("DIGITALOCEAN_IPV4_PRIVATE_{}", public_counter),
+                    format!("DIGITALOCEAN_IPV4_PRIVATE_{}", private_counter),
                     format!("{}", a.ipv4.unwrap().ip_address)
                 ));
             }


lucab:
Thanks for spotting and fixing this. Just one request and a minor suggestion:
 * please briefly fill in the body of the commit message, so that `tailor` is happy
 * if you are up for some Rust coding, that function may benefit from a little cleanup. In particular, those two `*_counter` variables should go away in favor of just [enumerate()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate).

src/providers/digitalocean/mod.rs
```diff
@@ -121,7 +121,7 @@ fn parse_attrs(data: &Metadata) -> Result<Vec<(String,String)>> {
         for a in ifaces {
```
lucab, line 121: Note: this should `enumerate()` to also track the loop counter. I didn't try but a straight fix should look like:
```
for (i, a) in ifaces.iter().cloned().enumerate() { ... }
```

(there is also probably a bit too much cloning going on)

tfussell:
@lucab Thanks for reviewing this. I'm afraid I've never used Rust before (though it doesn't look too different from C++) and I don't have much time at the moment to set up the build system on my machine and learn the syntax for the full fix. I just wanted to submit this small change before I forgot about it.

The commit message should be fixed now at least.

I'll be happy to contribute more in the future. Sorry about that.

sdemos:
That's totally fine. Commit message looks good to me. Please feel free to contribute in the future if you get the chance, we always welcome contributors!

sdemos (approved):

lucab:
@tfussell no worries, and thanks a lot for your PR!

PR #73, opened 2017-11-21
providers/do: get rid of mutable iface counters

lucab:
This updates DigitalOcean interface metadata parser to get rid
of top-level mutable counters in favor of scoped enumeration indexes.
It also avoids redundant cloning when traversing metadata structures.


lucab:
This is a followup to my previous comment at https://github.com/coreos/coreos-metadata/pull/72#discussion_r151967187. It also switches to references instead of owned data clones.

/cc @sdemos @dgonyeo @tfussell

sdemos (approved):

cgonyeo (approved):

PR #74, opened 2017-11-29
version: bump to 1.0.4

sdemos:
release notes draft:

* fixed a bug in digitalocean provider where private ip attributes would be improperly indexed if there were also public ips


sdemos:
\cc @bgilbert

PR #75, opened 2017-12-13
providers/oracle: use default when key is missing

sdemos:
the oracle metadata service doesn't require that a machine has an ssh
key assigned to it. when an ssh key is missing, it returns metadata
without the `ssh_authorized_keys` field in the json, which caused serde
to fail when attempting to deserialize it. this change makes serde use
the default instance for string (which is "") if it can't find the field
when deserializing.

ref coreos/bugs#2288

cc @bgilbert

### DIFF
--- a/src/providers/oracle/mod.rs
+++ b/src/providers/oracle/mod.rs
@@ -25,6 +25,7 @@ struct InstanceData {
 
 #[derive(Debug, Deserialize, Clone)]
 struct Metadata {
+    #[serde(default)]
     ssh_authorized_keys: String,
 }
 


cgonyeo:
While we're fixing this, are there any `InstanceData` fields which are optional?

bgilbert:
@dgonyeo I'm not aware of any documentation on the subject, and it's not obvious that any of those fields are optional.

bgilbert (approved):
This change fixed the problem for me.

cgonyeo (approved):

PR #76, opened 2017-12-14
providers/oracle: always say oracle-oci

sdemos:
oracle has multiple clouds, but we only support oracle-oci. this pr makes clear in log messages and code organization which one of the multiple metadata services are supported. 

ref coreos/bugs#2289

cc @bgilbert


cgonyeo (approved):

bgilbert (approved):
LGTM

PR #77, opened 2017-12-15
 providers/packet: respect bond interface requests 

sdemos:
the metadata for network interfaces has a bond field. up until now, we ignored it and just always created a bond named bond0, attached all addresses to it, and bonded both physical interfaces to it. now, we create any bond interfaces mentioned in the network interfaces list, so there may be multiple bonds. it also doesn't configure bonds for interfaces that don't have a bond key listed in the metadata.

/cc @bgilbert

### DIFF
--- a/src/providers/packet/mod.rs
+++ b/src/providers/packet/mod.rs
@@ -132,6 +132,9 @@ fn parse_network(netinfo: &PacketNetworkInfo) -> Result<(Vec<Interface>,Vec<Devi
             nameservers: Vec::new(),
             ip_addresses: Vec::new(),
             routes: Vec::new(),
+            // the interface should be unmanaged if it doesn't have a bond
+            // section
+            unmanaged: i.bond.is_none(),
         });
 
         // if there is a bond key, make sure we have a bond device for it
@@ -144,6 +147,7 @@ fn parse_network(netinfo: &PacketNetworkInfo) -> Result<(Vec<Interface>,Vec<Devi
                 bond: None,
                 ip_addresses: Vec::new(),
                 routes: Vec::new(),
+                unmanaged: false,
             };
             if !bonds.iter().any(|&(_, ref b): &(MacAddr, Interface)| &bond == b) {
                 bonds.push((mac, bond));
--- a/src/providers/packet/mod.rs
+++ b/src/providers/packet/mod.rs
@@ -102,8 +102,10 @@ pub fn fetch_metadata() -> Result<Metadata> {
 }
 
 fn get_dns_servers() -> Result<Vec<IpAddr>> {
-    let f = File::open("/run/systemd/netif/state")?;
-    let ip_strings = util::key_lookup_reader('=', "DNS", f)?
+    let f = File::open("/run/systemd/netif/state")
+        .chain_err(|| "failed to open /run/systemd/netif/state")?;
+    let ip_strings = util::key_lookup_reader('=', "DNS", f)
+        .chain_err(|| "failed to parse /run/systemd/netif/state")?
         .ok_or("DNS not found in netif state file")?;
     let mut addrs = Vec::new();
     for ip_string in ip_strings.split(' ') {
--- a/src/providers/packet/mod.rs
+++ b/src/providers/packet/mod.rs
@@ -62,6 +62,7 @@ struct PacketBondingMode {
 struct PacketInterfaceInfo {
     name: String,
     mac: String,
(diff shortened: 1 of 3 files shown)


sdemos:
nightly is going to fail because of some changes in the way macro scoping works, since we are not using the latest slog.

src/providers/packet/mod.rs
```diff
@@ -117,83 +118,103 @@ fn get_dns_servers() -> Result<Vec<IpAddr>> {
...
             ip_addresses: Vec::new(),
             routes: Vec::new(),
         });
+
+        // if there is a bond key, make sure we have a bond device for it
+        if let Some(ref bond_name) = i.bond {
+            let bond = Interface {
+                name: Some(bond_name.clone()),
+                priority: Some(5),
+                nameservers: get_dns_servers()?,
```
bgilbert: Strictly speaking, it's hypothetically possible for the return value to change between calls.

src/providers/packet/mod.rs
```diff
@@ -117,83 +118,103 @@ fn get_dns_servers() -> Result<Vec<IpAddr>> {
...
+                mac_address: Some(mac),
+                bond: None,
+                ip_addresses: Vec::new(),
+                routes: Vec::new(),
+            };
+            if !bonds.contains(&bond) {
```
bgilbert: Won't this always return false because of `mac_address`?

src/providers/packet/mod.rs
```diff
@@ -117,83 +118,103 @@ fn get_dns_servers() -> Result<Vec<IpAddr>> {
...
+                bonds.push(bond);
+            }
+        }
+    }
+
+    // according to the folks from packet, all the addresses given to us in the
+    // network section should be attached to the first bond we find in the list
+    // of interfaces. there probably won't be no bonds in the list, but if we
```
bgilbert: clarify `won't be no bonds`

src/providers/packet/mod.rs
```diff
@@ -117,83 +118,103 @@ fn get_dns_servers() -> Result<Vec<IpAddr>> {
...
-        nameservers: get_dns_servers()?,
-        mac_address: None,
-        bond: None,
-        ip_addresses: Vec::new(),
-        routes: Vec::new(),
-    };
+
+    // remove panics if the index is out of bounds, but we know that there is at
+    // least one bond in the vector because we return if it's empty
+    let mut bond0 = bonds.remove(0);
```
bgilbert: There's nothing requiring it to actually be named `bond0`, so maybe e.g. `first_bond` would be more descriptive.

src/providers/packet/mod.rs
```diff
@@ -117,83 +118,103 @@ fn get_dns_servers() -> Result<Vec<IpAddr>> {
...
-    interfaces.push(iface);
+    bonds.push(bond0);
 
     let mut attrs = vec![
-		("TransmitHashPolicy".to_owned(), "layer3+4".to_owned()),
-		("MIIMonitorSec".to_owned(), ".1".to_owned()),
-		("UpDelaySec".to_owned(), ".2".to_owned()),
-		("DownDelaySec".to_owned(), ".2".to_owned()),
-		("Mode".to_owned(), network::bonding_mode_to_string(&netinfo.bonding.mode)?),
+        ("TransmitHashPolicy".to_owned(), "layer3+4".to_owned()),
```
bgilbert, line 195: Inconsistent indentation?

bgilbert (changes requested):
Please clean up the commit messages to provide as much detail as possible.  We should explicitly say that the no-bond case is for type 1E.

src/providers/packet/mod.rs
sdemos: it would do that. I didn't notice because it created another file of the same name with almost the same contents that overwrote the other one, so it still emitted the right number of bond devices

src/providers/packet/mod.rs
sdemos: that sounded better in my head

src/providers/packet/mod.rs
sdemos: fixed.

sdemos:
@bgilbert modified according to suggestions.

bgilbert (approved):

sdemos:
even though @bgilbert is not technically an approved reviewer on this repository, he went over it very closely and we've tested this change, so I'm going to merge it so it gets into the next alpha.

PR #79, opened 2017-12-25
providers/libvirt: add libvirt provider using eth0 as network iface

jforman:
Based on the virtualbox provider, but using eth0 instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,10 @@ The supported cloud providers and their respective metadata are as follows:
       - COREOS_GCE_HOSTNAME
       - COREOS_GCE_IP_EXTERNAL_0
       - COREOS_GCE_IP_LOCAL_0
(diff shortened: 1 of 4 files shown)


jforman:
I'm trying to get the tailor tests to pass, but not quite sure which part of my request is making it fail (even with the reasons provided). Any help greatly appreciated.

crawford:
The scope can't have spaces. I would use "providers/libvirt" in this case. Your commit also needs a description after the title.

The rules can be found here: https://github.com/coreos/coreos-metadata/blob/master/.github/tailor.yaml

jforman:
@crawford Yep, just saw it now. Thanks!

jforman:
Following along to close the loop, this PR has been superseded by PR #81 .

PR #81, opened 2017-12-25
providers/libvirt: support for using eth0 as iface

jforman:
running coreos on libvirt (ubuntu) on bare-metal, the default
network interface is eth0, not eth1 as in use by other
bare-metal providers (vagrant-virtualbox). This provider mirrors
that provider almost exactly, just s/eth1/eth0/g.

### DIFF
--- a/src/providers/libvirt/mod.rs
+++ b/src/providers/libvirt/mod.rs
@@ -1,0 +1,64 @@
+// Copyright 2017 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! libvirt metadata fetcher
+
+use metadata::Metadata;
+
+use errors::*;
+
+use std::net::IpAddr;
+use std::time::Duration;
+use std::thread;
+
+use hostname;
+use pnet;
+
+pub fn fetch_metadata() -> Result<Metadata> {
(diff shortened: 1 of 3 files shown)


src/providers/libvirt/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+use hostname;
+use pnet;
+
+pub fn fetch_metadata() -> Result<Metadata> {
+    let h = hostname::get_hostname().ok_or("unable to get hostname")?;
+    let ip = get_ip()?;
+
+    Ok(Metadata::builder()
+       .add_attribute("LIBVIRT_PRIVATE_IPV4".to_owned(), ip)
+       .add_attribute("LIBVIRT_HOSTNAME".to_owned(), h.clone())
```
lucab, line 34: `COREOS_LIBVIRT_HOSTNAME`

src/providers/libvirt/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use hostname;
+use pnet;
+
+pub fn fetch_metadata() -> Result<Metadata> {
+    let h = hostname::get_hostname().ok_or("unable to get hostname")?;
+    let ip = get_ip()?;
+
+    Ok(Metadata::builder()
+       .add_attribute("LIBVIRT_PRIVATE_IPV4".to_owned(), ip)
```
lucab, line 33: `COREOS_LIBVIRT_IPV4_LOCAL`

lucab (changes requested):

lucab:
Naming aside, I'm not really sure we should introduce this new `libvirt` platform. It doesn't really use any externally-sourced metadata, just transient machine state.
libvirt has a dedicated [metadata section](https://libvirt.org/formatdomain.html#elementsMetadata) but the two are not guaranteed to match (and I think there is no way to introspect that from the guest).

I think `vagrant-virtualbox` is the odd one here for historical reasons and we are still addressing that in several places. I won't introduce more of that under a new platform.

jforman:
Okay, thanks Luca, that all sounds reasonable. Could you point me to where those vagrant-virtualbox 'historical reasons' are being addressed? I'd like to follow along.

lucab:
I don't have a single pointer for that, but there a few reference to chase starting from https://github.com/coreos/bugs/issues/2264#issuecomment-346582204.

PR #82, opened 2018-01-04
automated PR: update CoC

ericchiang:


cgonyeo (approved):

ericchiang:
Rust beta failures look real.

sdemos:
the failures are because of changes in the rust macro system that the current version of slog we are using doesn't support. we will have to upgrade to a new version of that library at some point in the future, but realistically we only care about the version of rust that is being used to build things in the sdk currently, which is 1.21.0. I changed the travis config to only use that for builds instead of the current alpha, beta, and stable in #83.

PR #88, opened 2018-06-18
Refactor metadata fetching into `MetadataProvider` trait

csssuf:
Instead of `fetch_metadata()` fetching all potentially needed metadata, return a `MetadataProvider` trait object which has drop-in replacements for `Metadata`'s functions. This allows only the requested metadata types to be fetched and avoids potentially expensive fetching work.

There's still a little bit of cleanup work that could take place in some of the providers (mostly rolling now-duplicate functions into the trait implementation), but those seemed like they should wait for another PR. This PR contains roughly the minimal amount of refactoring necessary to switch over to the `MetadataProvider` trait.

Fixes coreos/bugs#2362.


src/lib.rs
```diff
@@ -85,20 +85,24 @@ use metadata::Metadata;
...
 
+macro_rules! box_result {
+    ($exp:expr) => (Ok(Box::new($exp?)))
+}
+
 /// `fetch_metadata` is the generic, top-level function that is used by the main
 /// function to fetch metadata. The configured provider is passed in and this
 /// function dispatches the call to the correct provider-specific fetch function
-pub fn fetch_metadata(provider: &str) -> Result<Metadata> {
+pub fn fetch_metadata(provider: &str) -> Result<Box<providers::MetadataProvider>> {
```
crawford, line 93: Why is the Box needed?

csssuf: Since we're returning multiple different concrete types that all implement `MetadataProvider`, `impl Trait` can't be used and we need a `Box` to create an actual trait object.

crawford: I thought that was the whole point of `impl Trait`...

csssuf: Unfortunately not. From [the RFC](https://github.com/rust-lang/rfcs/blob/master/text/1522-conservative-impl-trait.md#semantics):
>If a function returns impl Trait, its body can return values of any type that implements Trait, but all return values need to be of the same type.

lucab: @csssuf for my own reference: pushing the `-> impl Trait` one more level down too (i.e. returned by `new()`), would that also not work?

csssuf: [Nope](https://play.rust-lang.org/?gist=6e939251c0048645127f65272cd7f875&version=stable&mode=debug). The compiler needs to be able to figure out static dispatch to use `impl Trait`, so there's no way to trick it into doing the thing we want here.

csssuf: [Nope](https://play.rust-lang.org/?gist=5adf057156d6a29d6ab2791d314a546b&version=stable&mode=debug). The compiler needs to be able to figure out static dispatch to use `impl Trait`, so there's no way to trick it into doing the thing we want here.

cgwalters:
This is the first time I've ever looked at this code, and my Rust is beginner-intermediate-ish...so take my comments with that in mind.

Overall, makes a lot of sense.  A personal style thing, I would tend to use a single commit that adds and deletes code at once, but eh.

I was about to ask about tests, but I found they exist just for EC2.  I assume Kola will cover other cases?

arithx:
@cgwalters `kola` does indeed cover a number of cases. The tests are [located here](https://github.com/coreos/mantle/blob/master/kola/tests/metadata/contents.go) (also some fields are used in other individual tests).

lucab:
Fully testing for this basically means reproducing all API and quirks of all supported cloud providers, so the mocking is mostly assuring that it doesn't break at a very coarse level. It only covers EC2 mostly due to time limitations, and because real/subtle regressions will hopefully break the integration CI on the relevant cloud provider.

src/lib.rs
cgwalters: https://www.reddit.com/r/rust/comments/8sqiin/newbie_question_why_is_the_impl_and_dyn_in_the/ is also useful on this topic (there's various others too)

With rustc beta today for example, one can write:

```
diff --git a/src/lib.rs b/src/lib.rs
index 0f80d1c..2ec3af3 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -90,7 +90,7 @@ macro_rules! box_result {
 /// `fetch_metadata` is the generic, top-level function that is used by the main
 /// function to fetch metadata. The configured provider is passed in and this
 /// function dispatches the call to the correct provider-specific fetch function
-pub fn fetch_metadata(provider: &str) -> Result<Box<providers::MetadataProvider>> {
+pub fn fetch_metadata(provider: &str) -> Result<Box<dyn providers::MetadataProvider>> {
     match provider {
         "azure" => box_result!(azure::Azure::new()),
         "cloudstack-metadata" => box_result!(cloudstack::network::CloudstackNetwork::new()),
```

Then it's explicitly clear the goal is dynamic dispatch.

BTW I love how much discussion there's been over a single heap allocation :smile:

src/providers/azure/mod.rs
```diff
@@ -275,19 +279,36 @@ impl Azure {
     }
 }
 
-pub fn fetch_metadata() -> Result<Metadata> {
-    let provider = Azure::new()
-        .chain_err(|| "azure: failed create metadata client")?;
+impl MetadataProvider for Azure {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        let attributes = self.get_attributes()?;
+        let mut out = HashMap::new();
```
lucab: Here and in all the other providers too, what about `with_capacity(2)` instead?

src/providers/cloudstack/configdrive.rs
```diff
@@ -69,36 +82,67 @@ impl ConfigDrive {
...
-       .add_attribute_if_exists("CLOUDSTACK_INSTANCE_ID".into(), drive.fetch_value("instance_id")?)
-       .add_attribute_if_exists("CLOUDSTACK_SERVICE_OFFERING".into(), drive.fetch_value("service_offering")?)
-       .add_attribute_if_exists("CLOUDSTACK_CLOUD_IDENTIFIER".into(), drive.fetch_value("cloud_identifier")?)
-       .add_attribute_if_exists("CLOUDSTACK_LOCAL_HOSTNAME".into(), drive.fetch_value("local_hostname")?)
-       .add_attribute_if_exists("CLOUDSTACK_VM_ID".into(), drive.fetch_value("vm_id")?)
-       .build())
+impl ::std::ops::Drop for ConfigDrive {
+    fn drop(&mut self) {
+        if self.temp_dir.is_some() {
+            ConfigDrive::unmount(&self.path).unwrap();
```
lucab, line 145: I'm not sure we should panic here. (This is carried-over code, feel free to defer this)

lucab:
@csssuf I did a cursory review of this, and it seems overall fine. Most of this is just moving functions around, I didn't dig into those.

lucab (approved):

src/lib.rs
sdemos: without having looked into it at all, I wonder if it would be possible to simultaneously use `dyn` and `impl Trait` to do what we want without a heap allocation? mostly curious, the current version is totally fine by me too.

csssuf: No, they're mutually exclusive. We _could_ avoid the `Box` by using an enum as the function's return type, and essentially provide our own dynamic dispatch by implementing the trait for the enum, but IMO that's even less clean than simply returning a `Box`ed trait object.

src/providers/gce/mod.rs
```diff
@@ -14,82 +14,122 @@
...
-    if let Some(key_data) = key_data {
-        let mut keys = Vec::new();
-        for l in key_data.lines() {
-            if l.is_empty() {
-                continue
+impl MetadataProvider for GceProvider {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        let mut out = HashMap::with_capacity(3);
+
+        let add_value = |map: &mut HashMap<_, _>, key: &str, name| -> Result<()> {
```
sdemos, line 94: Unfortunate that there is so much duplication with the `add_value` closure, but I'm not sure if there is a better way to do it after thinking through some of the alternatives.

sdemos (approved):
lgtm after we get positive tests on all the major cloud providers (it might need a rebase on master, shouldn't be any conflicts though)

csssuf:
Verified that the refactor's output matches the current release's (but without the unnecessary requests) on:
* EC2
* Azure
* DigitalOcean
* Packet
* GCE

PR #89, opened 2018-06-18
travis: bump rust version to match sdk

csssuf:
The SDK is now on Rust 1.26.0, so we should be building with that
version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,3 @@
 language: rust
 rust:
-  - 1.21.0
+  - 1.26.0


crawford (approved):

lucab:
@csssuf I think the customary travis setup for Rust projects is:
 * a version-hardcoded entry for the minimum toolchain version that the project can be built with
 * `stable` + `beta` + `nightly` entries, with `allow_failures` on the last one
 * a travis scheduled cronjob on master to test new toolchains

sdemos:
@lucab this project originally had that, but there was an issue at some point with changes that got into the new stable compiler relating to logging, where in order for the travis build to succeed we would've had to update slog, but the updated slog wouldn't have worked on the version in the sdk at the time. The solution was to only do CI with the version in the sdk (although obviously the actual version never got updated, and so now we were lagging behind quite a bit). 

We can move back to that model if we want to, but we will be at risk of having the same problem with another dependency in the future.

sdemos (approved):

csssuf:
It might make sense to take a hybrid approach, where the only version required for a Travis build to pass is the current SDK version, but stable/beta/nightly are all run and allowed to fail.

lucab:
@sdemos I remember that, but we weren't testing for an hardcoded minimum version at that point. Thus we didn't catch the SDK-version breakage on the PR updating slog.

Anyway, please merge this and then we can followup if we want.

PR #90, opened 2018-06-21
Add systemd units and a basic Makefile to install

cgwalters:
We're looking at adapting coreos-metdata for [FCOS](https://coreos.fedoraproject.org/).
We'll need to package it again as an RPM (probably).

Regardless, I think it's best to have systemd unit files upstream.
Taken from:
https://github.com/coreos/coreos-overlay/tree/master/coreos-base/coreos-metadata/files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,15 @@
+DESTDIR ?=
+PREFIX ?= /usr
+
+DEFAULT_INSTANCE ?= core
+
+units = $(addprefix systemd/,coreos-metadata.service coreos-metadata-sshkeys@.service)
+
+%.service: %.service.in
+	sed -e 's,@DEFAULT_INSTANCE@,'$(DEFAULT_INSTANCE)',' < $< > $@.tmp && mv $@.tmp $@
+
+all:
+	cargo build --release
+
+install-units: $(units)
+	for unit in $(units); do install -D --target-directory=$(DESTDIR)$(PREFIX)/lib/systemd/system/ $$unit; done
--- a/systemd/.gitignore
+++ b/systemd/.gitignore
@@ -1,0 +1,1 @@
+coreos-metadata-sshkeys@.service
--- a/systemd/coreos-metadata-sshkeys@.service.in
+++ b/systemd/coreos-metadata-sshkeys@.service.in
@@ -1,0 +1,11 @@
+[Unit]
+Description=CoreOS Metadata Agent (SSH Keys)
+
+[Service]
+Type=oneshot
+Environment=COREOS_METADATA_OPT_PROVIDER=--cmdline
+ExecStart=/usr/bin/coreos-metadata ${COREOS_METADATA_OPT_PROVIDER} --ssh-keys=%i
+
+[Install]
+DefaultInstance=@DEFAULT_INSTANCE@
+RequiredBy=multi-user.target
--- a/systemd/coreos-metadata.service
+++ b/systemd/coreos-metadata.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=CoreOS Metadata Agent
+
+[Service]
+Type=oneshot
+Environment=COREOS_METADATA_OPT_PROVIDER=--cmdline
+ExecStart=/usr/bin/coreos-metadata ${COREOS_METADATA_OPT_PROVIDER} --attributes=/run/metadata/coreos
+
+[Install]
+RequiredBy=metadata.target


Makefile
```diff
@@ -0,0 +1,10 @@
+DESTDIR ?=
+PREFIX ?= $(DESTDIR)/usr
```
bgilbert: Overriding both DESTDIR and PREFIX will produce a surprising result.

systemd/coreos-metadata-sshkeys@.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=CoreOS Metadata Agent (SSH Keys)
+
+[Service]
+Type=oneshot
+Environment=COREOS_METADATA_OPT_PROVIDER=--cmdline
+ExecStart=/usr/bin/coreos-metadata ${COREOS_METADATA_OPT_PROVIDER} --ssh-keys=%i
+
+[Install]
+DefaultInstance=core
```
bgilbert: This embeds the assumption that there is a `core` user.  Perhaps we don't want to do that upstream?

bgilbert (commented):
Moving the units upstream is probably okay, except for the assumption about the `core` user.

Please make sure the commit message includes the commit hash from which the unit files were taken.

Makefile
cgwalters: Fixed, thanks.

systemd/coreos-metadata-sshkeys@.service
cgwalters: Good point.  Hmm, that definitely makes things harder.  We could add `make install-units USERNAME=core` or so?

arithx:
tailor disable commit description

systemd/coreos-metadata-sshkeys@.service
cgwalters: I think my vote for now is to move forward with the assumption FCOS-derived systems will also have a `core` user by default.

sdemos (dismissed):
this all seems reasonable to me.

systemd/coreos-metadata-sshkeys@.service
bgilbert: FCOS-derived systems will have a `core` user, but this is going into the upstream package.  Maybe default the Makefile to `core`, with an override?

lucab: I just checked, and `DefaultInstance=` can be overridden via a drop-in (which is properly honored at `enable` time). I'm not much of a fan of templating this here.

bgilbert: Sure, but an arbitrary distro shouldn't have to (and, in practice, won't) ship a dropin if it wants to ship coreos-metadata.

sdemos: It seems to me like if we want to keep coreos-metadata as distro-agnostic as possible, the solution is templating in the user with the Makefile. I'm not too much of a fan of it as well, but unless someone has a different approach, I think we should just go with that for now.

sdemos:
Also this needs a rebase on master. Shouldn't be any conflicts though.

systemd/coreos-metadata-sshkeys@.service
crawford: CT is not meant to be distro-agnostic though. It's _very_ specific to Container Linux (e.g. etcd, flannel, locksmith).

EDIT: One of these days, I'll learn to read. Ignore me.

cgwalters:
:surfing_man: Rebased

systemd/coreos-metadata-sshkeys@.service
sdemos: ping on this discussion - I think it would be nice to get this in before we cut v2.0.0 (#93).

cgwalters:
I don't think the templating was overlooked, it seemed like there was more disagreement about whether it was necessary among the repo owners here?

If no one *objects* to doing it though I don't mind updating this PR to do that.

systemd/coreos-metadata-sshkeys@.service
csssuf: After a bit of offline discussion with @sdemos, I'm not actually sure providing a `DefaultInstance` in the upstream unit makes sense at all. Non-{CL, RHCOS, FCOS} distros which may want to ship coreos-metadata are unlikely to be able to choose a single `DefaultInstance` that makes sense, and setting `DefaultInstance=core` via a dropin is simple enough for CL/RHCOS/FCOS.

bgilbert: > Non-{CL, RHCOS, FCOS} distros which may want to ship coreos-metadata are unlikely to be able to choose a single DefaultInstance that makes sense

AFAIK the cloud image of pretty much every distro has a default user, whether it's `ec2-user`, `ubuntu`, `root`, etc.

csssuf: Ah, I suppose I was thinking mostly of non-cloud images, which doesn't really make sense in the context of coreos-metadata :slightly_smiling_face:

sdemos: @cgwalters @bgilbert @csssuf @lucab I think that given the constraints of the problem, templating is the best solution we have. Unless anyone is strongly opposed to it, I think @cgwalters should just implement it for this pr. If we come up with a better solution later, we can revisit this.

cgwalters: > AFAIK the cloud image of pretty much every distro has a default user, whether it's ec2-user, ubuntu, root, etc.

Kind of...one really important detail wrt cloud-init is that that default user is only created if you don't specify another user.  Concretely for Fedora Atomic (and Fedora Cloud) if you boot a cloud-init config that looks like this (a reduced version of my default one):

```
users:
  - name: walters
    gecos: Colin Walters
    sudo: ALL=(ALL) NOPASSWD:ALL
    ssh-authorized-keys:
      - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOQkQHeKan3X+g1jILw4a3KtcfEIED0kByKGWookU7ev walters+2015-general@verbum.org
```

You *don't* get a `fedora` user.  Or to say it conversely, the user is only created dynamically on boot - and this is different from CL today where the `core` user exists by default.

cgwalters: So any distribution that wants to use `coreos-metadata` would have to choose to either do what CL does and precreate that user, or alternatively have coreos-metadata do some sort of search dynamically for the "first" human user created by Ignition (possibly one that's in the `wheel/sudo` group?)

The thing that makes me uncomfortable about pre-creating e.g. a `core` user is - what if you join the node to LDAP/ActiveDirectory or whatever and there's already a "core" user?  You wouldn't want coreos-metadata granting ssh keys to that user...

In the end of course, an admin could delete the `core` user of course, but it does seem nicer to me to have the cloud-init semantics where it's a default user, but if you specify another one it doesn't exist.

bgilbert: There's not a good way to express the "conditional default user" thing in Ignition right now; that'd be a good issue for coreos/ignition.

Note that we're talking about `coreos-metadata-sshkeys@.service`, not the entirety of coreos-metadata.  Distros which conditionally create the default user can't unconditionally enable that unit anyway, since they don't know what account to associate it with.  In that case, I think it makes sense for the distro to either

1. ignore the `DefaultInstance` and always specify the desired instance, or
2. set the `DefaultInstance` to the conditional default user, and only enable `-sshkeys@.service` when the default user is actually created.

Either way, setting `DefaultInstance` is either useful or harmless.

cgwalters:
Rebased 🏄 and updated to template per discussion.

systemd/.gitignore
```diff
@@ -0,0 +1 @@
+coreos-metadata-sshkeys@.service
```
bgilbert: Missing trailing newline.

systemd/coreos-metadata-sshkeys@.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
bgilbert: File should be dropped.

systemd/coreos-metadata-sshkeys@.service
cgwalters: Split that to https://github.com/coreos/ignition/issues/600

systemd/coreos-metadata-sshkeys@.service
cgwalters: Oops, thanks!  Sorry about being slow here, I saw this comment go by and thought somehow you meant that section in the `.in` file should be dropped, and I started questioning whether there was some fundamental misunderstanding, then took a context switch.   Anyways updated :arrow_down:

sdemos (approved):
lgtm, as long as @bgilbert doesn't have any more comments.

bgilbert (approved):

sdemos:
@cgwalters sorry, github is complaining, can you rebase this one more time?

sdemos:
ehh it'll probably merge cleanly

PR #91, opened 2018-06-22
travis: build on minimum and current toolchains

lucab:
This checks that master and further PRs won't break on current          
toolchains, nor on older ones used by downstream distributions          
(pinning to 1.26 for historical ContainerLinux compatibility).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,10 @@
 language: rust
 rust:
-  - 1.26.0
+  - 1.26.0  # minimum supported toolchain
+  - stable
+  - beta
+  - nightly
+
+matrix:
+  allow_failures:
+    - rust: nightly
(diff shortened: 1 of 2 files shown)


lucab:
I enabled a travis weekly job for this repo so that it'll catch if master breaks with newer toolchains.

sdemos (approved):
lgtm

sdemos:
Digging into this more, it looks like the nightly build failed. I'm okay merging it anyway (since we allow failures on that one) but I thought the error was strange and I couldn't figure out where it was coming from.

lucab:
@sdemos it failed because of https://github.com/rust-lang/rust/issues/51699. Which is (one of) the reason why nightly is allowed to fail :)

sdemos:
ha I love getting hitting things and discovering that it was literally introduced yesterday. that happened to me with the new `panic_implementation` macro replacing `lang(panic)` too. I was surprised by it and discovered it was merged earlier that day when I hit it. 

anyway, thanks!

PR #93, opened 2018-07-02
version: release v2.0.0

csssuf:
Since #88 changed the public API in a backwards-incompatible way, we need to do a major version bump.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,7 +157,7 @@ dependencies = [
 
 [[package]]
 name = "coreos-metadata"
-version = "1.0.6"
+version = "2.0.0"
 dependencies = [
  "base64 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


lucab:
If we are doing a major bump, it would be good to update error-chain too first (as I think there was a public change there). That depends on https://github.com/sdemos/openssh-keys/pull/23 though.

csssuf:
@sdemos and I had talked about the possibility of publishing update-ssh-keys (and maybe eventually coreos-metadata as well) to crates.io and if we're interested in doing that, it might be worthwhile to switch over to using `failure` in the interest of ecosystem compatibility before doing a major version bump here as well. @lucab What do you think?

lucab:
I don't have any strong opinions on having those two binaries on crates.io. Regarding `failure`, I'd expect some major movements in that area in the near future (see https://github.com/rust-lang-nursery/failure/issues/224 and the bottom of https://github.com/rust-lang-nursery/failure/issues/209), so this is likely not the best time for a switch.

csssuf:
Hrmph. Agreed.

csssuf:
@sdemos @lucab Finally ready? :smile:

sdemos (approved):

lucab (approved):

PR #97, opened 2018-07-03
providers: don't write ssh keys if we have none

csssuf:
If we don't get any SSH keys from the metadata provider, don't even
bother trying to write them as update-ssh-keys will yell at us.

See coreos/bugs#2312.

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -72,20 +72,26 @@ pub trait MetadataProvider {
     }
 
     fn write_ssh_keys(&self, ssh_keys_user: String) -> Result<()> {
-        // find the ssh keys user and open their ssh authorized keys directory
-        let user = users::get_user_by_name(&ssh_keys_user)
-            .ok_or_else(|| format!("could not find user with username {:?}", ssh_keys_user))?;
-        let mut authorized_keys_dir = AuthorizedKeys::open(user, true, None)
-            .chain_err(|| format!("failed to open authorzied keys directory for user '{}'", ssh_keys_user))?;
+        let ssh_keys = self.ssh_keys()?;
 
-        // add the ssh keys to the directory
-        authorized_keys_dir.add_keys("coreos-metadata", self.ssh_keys()?, true, true)?;
+        if !ssh_keys.is_empty() {
+            // find the ssh keys user and open their ssh authorized keys directory
+            let user = users::get_user_by_name(&ssh_keys_user)
+                .ok_or_else(|| format!("could not find user with username {:?}", ssh_keys_user))?;
+            let mut authorized_keys_dir = AuthorizedKeys::open(user, true, None)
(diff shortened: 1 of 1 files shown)


cgwalters:
LGTM

sdemos (approved):

src/providers/mod.rs
```diff
@@ -72,20 +72,26 @@ pub trait MetadataProvider {
     }
 
     fn write_ssh_keys(&self, ssh_keys_user: String) -> Result<()> {
-        // find the ssh keys user and open their ssh authorized keys directory
-        let user = users::get_user_by_name(&ssh_keys_user)
-            .ok_or_else(|| format!("could not find user with username {:?}", ssh_keys_user))?;
-        let mut authorized_keys_dir = AuthorizedKeys::open(user, true, None)
-            .chain_err(|| format!("failed to open authorzied keys directory for user '{}'", ssh_keys_user))?;
+        let ssh_keys = self.ssh_keys()?;
```
lucab, line 75: I'd like to double-check the semantics here.
The new behavior with this PR would be to GET on the metadata endpoint, receive a 404, stop at the first attempt due to the 404, return an `Err` from `ssh_keys()` and thus bail out here. An empty set (which I'm not exactly sure can happen on AWS) would instead proceed straight without errors nor side-effects.

The behavior is technically correct, but at the same time I think it doesn't address the service unit failure as described in https://github.com/coreos/bugs/issues/2312.

Do we want to log an error here and proceed synthesizing an empty-set? @sdemos did you have additional thoughts for this not captured in https://github.com/coreos/bugs/issues/2312#issuecomment-357021493?

lucab (commented):
Code is fine but behavior likely does not fix the CL ticket.

src/providers/mod.rs
csssuf: Based on my reading of the error message in the original bug report, the call to `ssh_keys()` isn't failing but actually is returning an empty key set. The `authorized_keys_dir.sync()` call is failing since it has no keys to update, and [`update-ssh-keys` won't write an empty set of keys](https://github.com/coreos/update-ssh-keys/blob/1ac025c8f203889e129c7cb8c453c9ab57725352/src/lib.rs#L258-L262), so I believe the new behavior would be correct.

lucab: My bad, you are right. The endpoint returns a 404 and we ignore it in the client.
I just tried on an AWS t2.micro instance with no keypair and the result is:
```
$ sudo /opt/bin/coreos-metadata --cmdline --ssh-keys=core && echo OK
Jul 05 17:08:42.414 INFO Fetching http://169.254.169.254/2009-04-04/meta-data/public-keys: Attempt #1
Jul 05 17:08:42.415 INFO Fetch failed with 404: resource not found

OK
```

lucab (approved):

PR #99, opened 2018-07-11
metadata: fix minor typo in error messages

jlebon:
Just noticed this while perusing the codebase.


arithx (approved):

lucab:
@jlebon it looks like your commit is not on top of current master (but I'd argue for disabling this setting in GH, and consider using a proper queuing bot instead at a later point). Also, do you have merging right here?

jlebon:
Ahh yeah, I was working off an older tree. :deciduous_tree: 
It looks like the typo was fixed in #97 so nevermind! :)

PR #102, opened 2018-07-12
clippy: remove from cargo, add to travis

lucab:
This moves clippy setup from an optional cargo feature to an automatic CI check enforced via travis.

The side-effects of this is that we are now automatically checking every PR against clippy, and that we are dropping a few dependencies from the lockfile:
```
    Removing clippy v0.0.195                                                    
    Removing clippy_lints v0.0.195
    Removing either v1.5.0
    Removing getopts v0.2.17
    Removing if_chain v0.1.2
    Removing itertools v0.7.8
    Removing pulldown-cmark v0.1.2
    Removing quine-mc_cluskey v0.2.4
    Removing semver v0.9.0
    Removing semver-parser v0.7.0
    Removing toml v0.4.6
```

Travis pass is pinned to a specific clippy+nightly pair, which will need to be bumped from time to time in the future to opt-in new lints.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: rust
 rust:
-  - 1.26.0  # minimum supported toolchain
+  - nightly-2018-06-11  # pinned toolchain for clippy
+  - 1.26.0              # minimum supported toolchain
   - stable
   - beta
   - nightly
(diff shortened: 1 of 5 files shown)


lucab:
I usually enable hard-failures on warnings too (`cargo clippy -- -D warnings`) but there are a few warnings reported in the current codebase, and those can be addressed later.

csssuf (approved):

sdemos (changes requested):
we should remove the `cfg_attr` checks for the clippy feature from `lib.rs` and `bin/coreos-metadata.rs` as well, since we are now intending clippy to be run with `cargo clippy` instead of as part of the main compilation.

lucab:
@sdemos Indeed, I missed those bits, thanks. I added a commit on top to remove them. Is the direction of this otherwise ok? (I think you were the main one using the compiler-plugin approach, and I'm willing to trade that off for automatic CI)

sdemos:
honestly I set it up that way because I hadn't used clippy much back then and wasn't sure of the best way to invoke it. I haven't actually activated the compiler plugin approach in quite a while, I've since just integrated clippy suggestions into my editor. I think the automatic CI approach is a better one, since it automatically validates all incoming prs instead of relying on someone to run it.

sdemos (approved):

sdemos:
looks like the compiler plugin and optional dependency strategies have since been depricated in fact - https://github.com/rust-lang-nursery/rust-clippy/commit/654ff185881b6f5ab7048ed2b0c02227323fe0df and https://github.com/rust-lang-nursery/rust-clippy/commit/fd8a1d20ccad8a2dd7b95a9a9a6fe94048a74d5f

PR #103, opened 2018-07-13
src: apply clippy recommendations

sdemos:
either we haven't run clippy in a while, or clippy has some new lints,
or both, but there were a bunch of small changes. None of these changes
should modify the overall code behavior.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -42,9 +42,9 @@ const BONDING_MODES: [(u32,&str); 7] = [
     (BONDING_MODE_BALANCE_ALB,"balance-alb"),
 ];
 
-pub fn bonding_mode_to_string(mode: &u32) -> Result<String> {
+pub fn bonding_mode_to_string(mode: u32) -> Result<String> {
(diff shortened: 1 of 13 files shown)


sdemos:
I also went ahead and enabled failure on warnings as @lucab mentioned in #102

src/network.rs
```diff
@@ -42,9 +42,9 @@ const BONDING_MODES: [(u32,&str); 7] = [
     (BONDING_MODE_BALANCE_ALB,"balance-alb"),
 ];
 
-pub fn bonding_mode_to_string(mode: &u32) -> Result<String> {
```
csssuf, line 45: :eyes:

csssuf (approved):

src/network.rs
lucab: I think this was https://rust-lang-nursery.github.io/rust-clippy/master/index.html#trivially_copy_pass_by_ref

lucab (approved):

PR #104, opened 2018-07-19
providers/ec2: added EC2_PUBLIC_HOSTNAME attribute

fspijkerman:
Added attribute EC2_PUBLIC_HOSTNAME for the EC2 provider

Attribute EC2_PUBLIC_HOSTNAME for EC2, was introduced at `2007-01-19`

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ The supported cloud providers and their respective metadata are as follows:
     - SSH Keys
     - Attributes
       - COREOS_EC2_HOSTNAME
+      - COREOS_EC2_PUBLIC_HOSTNAME
       - COREOS_EC2_IPV4_LOCAL
       - COREOS_EC2_IPV4_PUBLIC
       - COREOS_EC2_AVAILABILITY_ZONE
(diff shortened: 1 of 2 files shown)


lucab:
@fspijkerman thanks for this PR! `tailor` is complaining that your commit does not have a commit body, would you mind filling that in (copying the PR body would be enough)?

This looks like an optional attribute. ~~It should be marked as such in the README, to highlight that it may be unavailable.~~ This in turn brings a couple more questions from my side:
 * What's the endpoint behavior when such attribute is unavailable? I'd expect a 404 which is just ignored by the client, but I'd better ask explicitly if you tested this scenario (we should consider adding a test too at some point)
 * Where do you plan to consume this? In particular, is it ok to put the burden on the consumer to deal with the fact that the attribute may be missing? (I think that's a fair expectation, and the implicit status quo)

lucab:
Forgot to say: I checked API compatibility too, this attribute was introduced in `2007-01-19` and we are using `2009-04-04`, so :+1:.

fspijkerman:
I agree, its optional indeed. 

* Endpoint behaviour:
  * if a public ip address (floating or shared) is assigned to a vm and enableDnsHostnames=true (default) the public hostname is available. 
  * if the EC2 instance has no public ip address an 404 is returned
  * if enableDnsHostnames=false an 404 is returned.

 * Consumer:
I think thats a fair expectation too, it's depending on how the EC2 instance has been setup.

fspijkerman:
I amend my last commit, it seem tailor is still not happy.

lucab (approved):

PR #110, opened 2018-08-03
azure/crypto: remove all unwraps

lucab:
This converts all `unwrap` usages in `azure/crypto` into proper
error chaining.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,6 +66,7 @@ pub mod errors {
             Io(::std::io::Error);
             Reqwest(::reqwest::Error);
             Hyper(::hyper::error::Error);
+            OpensslStack(::openssl::error::ErrorStack);
         }
         errors {
             UnknownProvider(p: String) {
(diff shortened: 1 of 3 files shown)


lucab:
This is related to the discussion at https://github.com/coreos/coreos-metadata/issues/109, however it is not meant to fix it. https://github.com/coreos/coreos-metadata/pull/88 should have already addressed that.

cgwalters:
LGTM

csssuf (approved):
LGTM as-is, but we may also want to remove the index accesses [here](https://github.com/coreos/coreos-metadata/blob/58a102b67e2c497f79b5393e30a3388996efbfe4/src/providers/azure/mod.rs#L204) and [here](https://github.com/coreos/coreos-metadata/blob/58a102b67e2c497f79b5393e30a3388996efbfe4/src/providers/azure/mod.rs#L257), as they can also panic.

sdemos (approved):

lucab:
@csssuf Indeed. Would you mind noting that into its own ticket? I think it would be a good "low fruit/help wanted" issue.

PR #112, opened 2018-08-08
providers/gcp: scrape new endpoint for ssh keys

andor44:
The old `sshKeys` endpoint for project and instance keys has been
deprecated for more than a year. The new name of the attribute is
`ssh-keys`.

For reference: https://github.com/gc3-uzh-ch/elasticluster/issues/391

https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys does not mention the new key (`ssh-keys`) but it mentions that the old one is deprecated.

### DIFF
--- a/src/providers/gce/mod.rs
+++ b/src/providers/gce/mod.rs
@@ -46,10 +46,20 @@ impl GceProvider {
     }
 
     fn fetch_all_ssh_keys(&self) -> Result<Vec<String>> {
+        // The Google metadata API has a total of 4 endpoints to retrieve SSH keys from:
+        // First, there are instance-level and project-level SSH keys.
+        // Additionally, there are two attributes on both levels where these are stored, one called
+        // `sshKeys`, and one called `ssh-keys`. The former is considered deprecated on both levels
+        // but it can still be found in some setups, therefore we have to handle that.
+        // https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys
+
+        // Instance-level, old endpoint
+        // If there are any of these, don't do anything else.
         let keys = self.fetch_ssh_keys("instance/attributes/sshKeys")?;
         if !keys.is_empty() {
             return Ok(keys);
         }
+        // Instance-level, new endpoint
         let mut keys = self.fetch_ssh_keys("instance/attributes/ssh-keys")?;
 
         let block_project_keys: Option<String> = self.client
@@ -61,7 +71,10 @@ impl GceProvider {
             return Ok(keys);
         }
 
+        // Project-level, old endpoint
         keys.append(&mut self.fetch_ssh_keys("project/attributes/sshKeys")?);
+        // Project-level, new endpoint
+        keys.append(&mut self.fetch_ssh_keys("project/attributes/ssh-keys")?);
 
         Ok(keys)
     }


lucab:
@andor44 looking at IRC backlog, I think this is coming after an hard `coreos-metadata-sshkeys` failure. Can you please share error logs and instance cloud settings for your setup?

andor44:
It's a combination of issues: the hard failure is due to https://github.com/coreos/coreos-metadata/pull/97 and that's already fixed in the latest, 2.0.0, release which is not yet in CL, which has 1.0.6. Here's the logs:

```
-- Logs begin at Wed 2018-08-08 15:09:48 UTC, end at Thu 2018-08-09 09:17:23 UTC. --
Aug 08 15:10:09 systemd[1]: Starting CoreOS Metadata Agent (SSH Keys)...
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.937 INFO Fetching http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/access-configs/0/external-ip: Attempt #1
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.980 INFO Fetch successful
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.983 INFO Fetching http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces/0/ip: Attempt #1
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.985 INFO Fetch successful
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.985 INFO Fetching http://metadata.google.internal/computeMetadata/v1/instance/hostname: Attempt #1
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.988 INFO Fetch successful
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.988 INFO Fetching http://metadata.google.internal/computeMetadata/v1/instance/attributes/sshKeys: Attempt #1
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.990 INFO Fetch failed with 404: resource not found
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.990 INFO Fetching http://metadata.google.internal/computeMetadata/v1/instance/attributes/ssh-keys: Attempt #1
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.992 INFO Fetch failed with 404: resource not found
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.992 INFO Fetching http://metadata.google.internal/computeMetadata/v1/instance/attributes/block-project-ssh-keys: Attempt #1
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.996 INFO Fetch failed with 404: resource not found
Aug 08 15:10:09 coreos-metadata[777]: Aug 08 15:10:09.996 INFO Fetching http://metadata.google.internal/computeMetadata/v1/project/attributes/sshKeys: Attempt #1
Aug 08 15:10:10 coreos-metadata[777]: Aug 08 15:10:10.000 INFO Fetch failed with 404: resource not found
Aug 08 15:10:10 coreos-metadata[777]: Error: writing ssh keys
Aug 08 15:10:10 coreos-metadata[777]: Caused by: failed to update authorized keys
Aug 08 15:10:10 coreos-metadata[777]: Caused by: update-ssh-keys: no keys found in "/home/core/.ssh/authorized_keys.d"
Aug 08 15:10:10 systemd[1]: coreos-metadata-sshkeys@core.service: Main process exited, code=exited, status=1/FAILURE
Aug 08 15:10:10 systemd[1]: coreos-metadata-sshkeys@core.service: Failed with result 'exit-code'.
Aug 08 15:10:10 systemd[1]: Failed to start CoreOS Metadata Agent (SSH Keys).
```

As you can see it attempts both `sshKeys` and `ssh-keys` on the instance level, but only the former on the project level. However, on the project level the former is deprecated as can be seen in the issue linked in the original post. 

```
$ curl -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/project/attributes/ssh-keys -q
andor:ssh-rsa AAAAB3NzaC1yc2EAAAADAQABA....

$ curl -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/project/attributes/sshKeys -q
<!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 404 (Not Found)!!1</title>
...
```
This is the latest, most minimal setup possible in GCP. I created a completely new and fresh project, configured nothing just created an instance with the latest CL stable image and ssh'd to it with `gcloud compute ssh instance-1`

My PR just adds compatibility for the new endpoint where SSH pubkeys are stored.

src/providers/gce/mod.rs
```diff
@@ -62,6 +62,10 @@ impl GceProvider {
         }
 
         keys.append(&mut self.fetch_ssh_keys("project/attributes/sshKeys")?);
+        if !keys.is_empty() {
```
lucab: This re-uses `keys` from the instance-level metadata, so it may result in skipping new-endpoint project keys because of instance ones (which I think is unexpected behavior).

I suggest to split and rename into two variables, `instance_keys` and `project_keys`, and also add a comment at the top detailing the deprecated endpoint and referencing GCloud docs.

andor44: Good catch. Actually, what do you prefer to do if both new-endpoint and old-endpoint exist for project level? Return both? Or check new and fall back to old?

src/providers/gce/mod.rs
```diff
@@ -46,24 +46,41 @@ impl GceProvider {
...
 
-        keys.append(&mut self.fetch_ssh_keys("project/attributes/sshKeys")?);
+        // Project-level, first check new endpoint, if it doesn't exist, check old
+        let mut project_keys = Vec::new();
+        project_keys.append(&mut self.fetch_ssh_keys("project/attributes/ssh-keys")?);
+        if !project_keys.is_empty() {
+            project_keys.append(&mut self.fetch_ssh_keys("project/attributes/sshKeys")?);
+        }
+
+        let all_keys = instance_keys.into_iter().chain(project_keys).collect();
```
sdemos: no need to turn this into an iterator, you can just use `append`. e.g. `let all_keys = instance_keys.append(&mut project_keys);`

src/providers/gce/mod.rs
```diff
@@ -46,24 +46,41 @@ impl GceProvider {
...
-            return Ok(keys);
+        // The Google metadata API has a total of 4 endpoints to retrieve SSH keys from:
+        // First, there are instance-level and project-level SSH keys.
+        // Additionally, there are two attributes on both levels where these are stored, one called
+        // `sshKeys`, and one called `ssh-keys`. The former is considered deprecated on both levels
+        // but it can still be found in some setups, therefore we have to handle that.
+        // https://cloud.google.com/compute/docs/instances/adding-removing-ssh-keys
+
+        // Instance-level, old endpoint
+        // If there are any of these, don't do anything else.
```
sdemos, line 57: I know that this is the current behavior of the function, but it seems...weird. Is it true that if the old instance ssh keys endpoint exists, we shouldn't retrieve the project keys at all?

src/providers/gce/mod.rs
```diff
@@ -46,24 +46,41 @@ impl GceProvider {
...
 
         if block_project_keys == Some("true".to_owned()) {
-            return Ok(keys);
+            return Ok(instance_keys);
         }
 
-        keys.append(&mut self.fetch_ssh_keys("project/attributes/sshKeys")?);
+        // Project-level, first check new endpoint, if it doesn't exist, check old
+        let mut project_keys = Vec::new();
+        project_keys.append(&mut self.fetch_ssh_keys("project/attributes/ssh-keys")?);
```
sdemos: nit: `let project_keys = self.fetch_ssh_keys("project/attributes/ssh-keys")?;` instead of creating a new vector and appending to it.

lucab:
@sdemos @andor44 answering both of your questions.

Docs says:
> If instance-level metadata is set to block project-wide SSH keys or has a deprecated instance-only sshKeys value, the instance will ignore all project-wide SSH keys. To apply project-wide keys to an instance, make sure the instance allows project-wide public SSH keys and, if present, remove the deprecated instance-only sshKeys value from instance metadata.

> If you want your instance to use both project-wide and instance-level public SSH keys, set the instance metadata to allow project-wide SSH keys.

So I think the logic should be:
 1. fetch instance `sshKeys`
 1. if not empty, return it
 1. fetch and merge instance `ssh-keys`
 1. fetch project `sshKeys`, if not empty merge with all of the above and return
 1. fetch project `ssh-keys`, merge with all of the above and return

src/providers/gce/mod.rs
andor44: `append` modifies the Vec in place, doesn't return a new one, therefore that won't work. At first I had 
```rust
let mut all_keys = Vec::new();
all_keys.append(&mut instance_keys);
all_keys.append(&mut project_keys);
```
But then the wrote the iterator one, which after optimizations becomes essentially the same code. And the iterator one feels 1.) more natural to read to me 2.) lends itself better to any possible future extension, as it'd be
```rust
instance_keys.into_iter()
    .chain(project_keys)
    .chain(vpc_keys)
    .chain(whatever_keys)
    .collect();
``` 
for example.

andor44:
Updated to reflect that behavior. The latest version is pretty similar to the original and reads reasonably well IMO.

src/providers/gce/mod.rs
sdemos: fair points. I don't know why I always get the append behavior confused, it was even staring me in the face and I still missed it.

sdemos (approved):
lgtm. I'll merge it if @lucab also approves.

lucab (approved):

lucab:
LGTM :+1: 
@andor44 thanks a lot for reporting this, tracking down the doc reference and iterating through the PR!

PR #113, opened 2018-08-17
src: make a single-binary only project

lucab:
This ensures that the coreos-metadata crate only provides a single
binary. The corresponding library is not supposed to be externally
consumed and is not guaranteed to have a stable API (at the moment).

It also updates dependencies and adjusts cargo manifest for publishing
the next version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,11 +1,11 @@
 [[package]]
 name = "adler32"
-version = "1.0.2"
+version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "aho-corasick"
-version = "0.6.4"
+version = "0.6.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


lucab:
@sdemos @csssuf please double-check me, but I think we were not aiming at providing a public library (at least for now). Given that I'd like to upload this to crates.io, I think it would be better not to expose that at all.

sdemos:
yeah, the primary purpose is definitely just to provide the binary. There isn't really a good reason to pin ourselves in by exposing the library part at this point, at least as far as I can see.

sdemos (approved):

csssuf (approved):
I suppose technically this requires a `3.0` :slightly_smiling_face:

lucab:
@csssuf why? We never published the crate and we are dropping the library, so the semver does not apply to that. For the binary, IMHO the versioning is pretty much arbitrary. I think we will need to bump to `3.0` if/when we want to re-introduce the public library.

lucab:
I'm merging this for now. I am happy to continue the versioning discussion here and account for that when preparing the next release PR.

csssuf:
@lucab We bumped to 2.0.0 when we changed the API of the library portion for the `MetadataProvider` trait, but if we don't think there are any actual consumers of the library then I'd be fine sticking with `2.0`.

lucab:
@csssuf my reasoning is that:
 1. for 2.0.0, there can't be any stable consumer of this library as we never published it to crates.io
 1. for (an hypothetical) 3.0.0, there can't be any stable consumer of this library as the library (as in `extern crate coreos-metadata`) does not exist anymore

If the first is not true, then your concern is valid. In that case, I think it may be safer to just bump to 3.0.0 and be done with the library. People (if any) could still lock into library 2.0.0 via git dependency.

lucab:
As the outcome of this discussion, to be on the safe side I'm bumping the current development series version in https://github.com/coreos/coreos-metadata/pull/114.

PR #116, opened 2018-09-24
util: minor fixes to cmdline parser

lucab:
This fixes few minor issues in /proc/cmdline parser. In particular,
it now trims newlines and tabs from ID values, ignore empty values
and fronts up key-value filtering when collecting elements to reduce
the number of iterations.

OEM cmdline parser has now moved to its own file and gained a few
basic testcases.

### DIFF
--- a/src/util/cmdline.rs
+++ b/src/util/cmdline.rs
@@ -1,0 +1,98 @@
+// Copyright 2018 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Kernel cmdline parsing - utility functions
+//!
+//! NOTE(lucab): this is not a complete/correct cmdline parser, as it implements
+//!  just enough logic to extract the OEM ID value. In particular, it doesn't
+//!  handle separator quoting/escaping, list of values, and merging of repeated
+//!  flags.
+
+use errors::*;
+use std::{fs, io};
+use std::io::Read;
+
+// Get OEM ID value from cmdline file.
+pub fn get_oem(fpath: &str, flagname: &str) -> Result<String> {
+    // open the cmdline file
+    let file =
+        fs::File::open(fpath).chain_err(|| format!("Failed to open cmdline file ({})", fpath))?;
+
+    // read the contents
+    let mut bufrd = io::BufReader::new(file);
+    let mut contents = String::new();
+    bufrd
+        .read_to_string(&mut contents)
+        .chain_err(|| format!("Failed to read cmdline file ({})", fpath))?;
+
+    match find_flag_value(flagname, &contents) {
+        Some(s) => Ok(s),
+        None => bail!(
+            "Couldn't find flag '{}' in cmdline file ({})",
+            flagname,
+            fpath
+        ),
+    }
+}
+
+// Find OEM ID flag value in cmdline string.
+fn find_flag_value(flagname: &str, cmdline: &str) -> Option<String> {
+    // split the contents into elements and keep key-value tuples only.
+    let params: Vec<(&str, &str)> = cmdline
+        .split(' ')
+        .filter_map(|s| {
+            let kv: Vec<&str> = s.splitn(2, '=').collect();
+            match kv.len() {
+                2 => Some((kv[0], kv[1])),
+                _ => None,
+            }
+        }).collect();
+
+    // find the oem flag
+    for (key, val) in params {
+        if key != flagname {
+            continue;
+        }
+        let bare_val = val.trim();
+        if !bare_val.is_empty() {
+            return Some(bare_val.to_string());
+        }
+    }
+    None
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn test_find_flag() {
+        let flagname = "coreos.oem.id";
+        let tests = vec![
+            ("", None),
+            ("foo=bar", None),
+            ("coreos.oem.id", None),
+            ("coreos.oem.id=", None),
+            ("coreos.oem.id=\t", None),
+            ("coreos.oem.id=ec2", Some("ec2".to_string())),
+            ("coreos.oem.id=\tec2", Some("ec2".to_string())),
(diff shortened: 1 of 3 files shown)


lucab:
/cc @jlebon @sdemos for a review

This is because I got a trailing newline in `/proc/cmdline` which resulted in:
```
coreos-metadata[4114]: Error: fetching metadata from provider
coreos-metadata[4114]: Caused by: unknown provider 'ec2
coreos-metadata[4114]: '
```

src/util/cmdline.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn test_find_flag() {
+        let flagname = "coreos.oem.id";
+        let tests = vec![
+            ("", None),
+            ("foo=bar", None),
+            ("coreos.oem.id", None),
+            ("coreos.oem.id=", None),
+            ("coreos.oem.id=\t", None),
+            ("coreos.oem.id=ec2", Some("ec2".to_string())),
+            ("coreos.oem.id=\tec2", Some("ec2".to_string())),
```
cgwalters, line 88: Is this one really valid?

src/util/cmdline.rs
```diff
@@ -0,0 +1,91 @@
...
+            ("coreos.oem.id=ec2\n", Some("ec2".to_string())),
+            ("foo=bar coreos.oem.id=ec2", Some("ec2".to_string())),
```
cgwalters, line 90: Might as well add a test case with trailing arguments too?

src/util/cmdline.rs
```diff
@@ -0,0 +1,91 @@
...
+            fpath
+        ),
+    }
+}
+
+// Find OEM ID flag value in cmdline string.
+fn find_flag_value(flagname: &str, cmdline: String) -> Option<String> {
+    // split the contents into elements and keep key-value tuples only.
+    let params: Vec<(&str, &str)> = cmdline
+        .split(' ')
```
cgwalters: `char::is_whitespace` ?

cgwalters (commented):
Presumably this occurred in real-world testing; can you elaborate on that?  Did it happen with CL or FCOS?

cgwalters:
One other thing to note; the Linux kernel command line is a [multi-map](https://en.wikipedia.org/wiki/Multimap) - while for us there can only be one OEM, in [for example libostree's kernel argument class](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-kernel-args.c) we go to some effort to support arguments with multiple values.  Probably worth adding a comment that this isn't a "real" parser and would need to be generalized to do so.

src/util/cmdline.rs
lucab: I don't have good arguments on either sides. I just thought that a `Some("\tec2")` would be less useful than just trimming on both ends of the value.

src/util/cmdline.rs
lucab: `bootparam(7)` says `a list of strings (boot arguments) separated by spaces`. I don't think we should go for the UTF-8 correct way here.

lucab:
Indeed, this is not a complete+correct parser. I've seen a couple on crates.io but I'm not familiar with any, so I can leave a note/ticket for now and then evaluate using an existing one.

As per my second comment, I hit that on a real-world trailing newline. It was latest RHCOS on AWS, and the cmdline looked like:
```
$ hexdump -C /proc/cmdline | tail
00000110  2d 66 38 31 36 2d 34 30  66 39 2d 39 63 37 36 2d  |-f816-40f9-9c76-|
00000120  66 33 65 34 66 62 62 35  38 30 33 63 20 6f 73 74  |f3e4fbb5803c ost|
00000130  72 65 65 3d 2f 6f 73 74  72 65 65 2f 62 6f 6f 74  |ree=/ostree/boot|
00000140  2e 30 2f 72 68 63 6f 73  2f 33 62 65 63 37 63 32  |.0/rhcos/3bec7c2|
00000150  33 62 38 33 31 62 32 61  38 39 34 35 65 37 35 64  |3b831b2a8945e75d|
00000160  38 39 62 34 64 30 61 36  36 64 36 66 36 32 36 62  |89b4d0a66d6f626b|
00000170  64 33 38 39 62 32 32 66  33 35 33 65 31 39 32 32  |d389b22f353e1922|
00000180  36 39 35 62 37 38 35 65  38 2f 30 20 63 6f 72 65  |695b785e8/0 core|
00000190  6f 73 2e 6f 65 6d 2e 69  64 3d 65 63 32 0a        |os.oem.id=ec2.|
```

src/util/cmdline.rs
jlebon: This will get mixed up with quoted spaces, right? E.g. `arg="my quoted string"`. Though feel free to file this under the "this is not a 100% compliant parser" ticket. :)

src/util/cmdline.rs
```diff
@@ -0,0 +1,91 @@
...
+        .filter_map(|s| {
+            let kv: Vec<&str> = s.split('=').collect();
```
jlebon: `splitn('=', 1)` ? So that values can have `=`.

src/util/cmdline.rs
lucab: Indeed, this doesn't have any kind of quoting/escaping awareness either (but I haven't checked if that kind of quoting is actually allowed, or backslash-escaping or no-spaces).

lucab:
Thanks everybody for the feedback! I've added an additional testcase, bound label splitting to the first equal sign, and documented all the known shortcomings of this naïf parser. Once merged, I will leave a ticket open to investigate (re-using) a proper parser. PTAL.

src/util/cmdline.rs
```diff
@@ -0,0 +1,98 @@
...
+    // open the cmdline file
+    let file =
+        fs::File::open(fpath).chain_err(|| format!("Failed to open cmdline file ({})", fpath))?;
+
+    // read the contents
+    let mut bufrd = io::BufReader::new(file);
+    let mut contents = String::new();
+    bufrd
+        .read_to_string(&mut contents)
+        .chain_err(|| format!("Failed to read cmdline file ({})", fpath))?;
```
jlebon, line 37: Minor: hmm interesting formatting -- is this `rustfmt`?

jlebon (approved):
LGTM!

src/util/cmdline.rs
lucab: Yes, that's rustfmt-preview from 1.29.0 toolchain.

lucab:
@sdemos without hurry, GH needs a green review to let me cleanly merge this. I'm going for a 3.0.1 PR afterwards.

sdemos (approved):

lucab:
@jlebon @cgwalters forgot to say, we have a @coreos/team-os-internal-crates-owners [team](https://crates.io/teams/github:coreos:team-os-internal-crates-owners) for shared maintenance of this. Let me know if you'd like to join, I can send invitations.

PR #119, opened 2018-10-24
cargo: update deps to latest compatible

lucab:
This performs a lockfile-only bump to update dependencies to their
latest compatible version. This does not cover dependencies with a new
major version available, which require individual massaging for
breaking changes in their API.


jlebon:
Looks sane to me!

sdemos (approved):

PR #123, opened 2018-11-05
cargo: update some dependencies

lucab:
This is a general dependencies update, for those crates that didn't require major reworks.


lucab:
Chore update which has been around for a while. Assuming no objections / blocking-reviews, I'd plan to self-merge this tomorrow.

src/providers/cloudstack/configdrive.rs
```diff
@@ -84,7 +84,7 @@ impl ConfigDrive {
     }
 
     fn mount_ro(source: &Path, target: &Path, fstype: &str) -> Result<()> {
-        mount::mount(Some(source), target, Some(fstype), mount::MS_RDONLY, None::<&str>)
+        mount::mount(Some(source), target, Some(fstype), mount::MsFlags::MS_RDONLY, None::<&str>)
```
ashcrow, line 87: Did this mean to be put into this commit?

lucab: Yes, it's due to an API change in `nix`.

Cargo.toml
```diff
@@ -15,29 +15,29 @@ path = "src/main.rs"
...
+serde-xml-rs = "0.2"
+serde_derive = "1.0"
+serde_json = "1.0"
+slog-async = "2.1"
+slog-scope = "4.0"
+slog-term = "2.2"
 tempdir = "0.3"
-nix = "0.9"
+update-ssh-keys = "0.3.0"
+users = "0.7"
```
jlebon, line 40: Minor: looks like this commit also alphabetizes the dependency list too? I usually split those out as a separate commit and explicitly mention there are no functional changes. Otherwise, mixing lots of movements with other changes makes it harder to tell what really changed.

src/providers/cloudstack/configdrive.rs
jlebon: Yeah, looks like the constants were moved from 0.9 to 0.11.

jlebon: Ahh, too slow. :)

jlebon:
One nit comment, otherwise LGTM.

Cargo.toml
lucab: Indeed. I've split that into a dedicated earlier commit.

jlebon (approved):

PR #127, opened 2018-12-17
cargo: update most dependencies to latest versions

lucab:


lucab:
This should be the last round of bulk dependencies updates. After this, we can enable dependabot to take care of bumping (compatible) updates for us.

@dm0- unfortunately due to transitive dependencies (in particular, `mockito` and `encoding_rs`), minimum toolchain version jumps to 1.30.0 (which for CL is only in alpha, I think).

dm0- (approved):
Beta has Rust 1.30.1, and that beta will be promoted stable after the holiday.  Even if there is some issue requiring upgrading coreos-metadata in stable before then, I'd be okay with updating its Rust since we only have two Rust programs in prod images, and they've had plenty of testing time with 1.30 in the beta channel.

lucab:
@dm0- thanks for double-checking, that's a good news! I somehow inferred beta was at 1.29.0.

lucab:
On the other hand, @jlebon suggested that 1.30 may be too new for building with RHEL7 devtoolset, which currently seems to be at 1.29. Pinging @cuviper in case we missed something. In that case, I can check if I can hold back a reduced set of dependencies to stay on 1.29.

cuviper:
Right, the Q4 release for RHEL7 was rust-toolset-1.29.

lucab:
https://github.com/coreos/coreos-metadata/pull/127/commits/17bb8e60cf4d017e78c3b19536f72d090ebdba44 pins `mockito` and `slog-scope` to earlier versions. Everything else is fine with 1.29 as a minimum toolchain.

jlebon:
Looks sane to me!

PR #130, opened 2018-12-21
[WIP] provider: add boot check-in on azure and packet

lucab:
DO NOT MERGE: neither providers have been tested on the field.

---

This adds a phone-home feature via the `--check-in` flag, in order to
support reporting back readiness state to the hosting infrastructure.
Initial implementation supports azure and packet.

Closes: https://github.com/coreos/coreos-metadata/issues/120

PR #145, opened 2019-01-14
azure: hardcode fallback for wireserver endpoint

jlebon:
If we can't fetch the wireserver address from the DHCP lease option, do
what WALinuxAgent does and use the fallback wireserver endpoint at
168.63.129.16.

### DIFF
--- a/src/providers/azure/mod.rs
+++ b/src/providers/azure/mod.rs
@@ -46,6 +46,10 @@ Content-Transfer-Encoding: base64
 
 ";
 
+/// This is a known working wireserver endpoint within Azure.
+/// See: https://blogs.msdn.microsoft.com/mast/2015/05/18/what-is-the-ip-address-168-63-129-16/
+const FALLBACK_WIRESERVER_ADDR: [u8; 4] = [168, 63, 129, 16]; // for grep: 168.63.129.16
+
 #[derive(Debug, Deserialize, Clone, Default)]
 struct GoalState {
     #[serde(rename = "Container")]
@@ -150,8 +154,7 @@ pub struct Azure {
 
 impl Azure {
     pub fn try_new() -> Result<Azure> {
-        let addr = Azure::get_fabric_address()
-            .chain_err(|| "failed to get fabric address")?;
+        let addr = Azure::get_fabric_address();
         let client = retry::Client::try_new()?
             .header(HeaderName::from_static(HDR_AGENT_NAME),
                     HeaderValue::from_static(MS_AGENT_NAME))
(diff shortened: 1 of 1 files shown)


jlebon:
Prep for #130.

jlebon:
I've also verified that this fallback is compatible with our metadata version.

src/providers/azure/mod.rs
```diff
@@ -46,6 +46,9 @@ Content-Transfer-Encoding: base64
 
 ";
 
+// this is a known working wireserver endpoint within Azure
```
lucab: Nit: if you make this a docstring (`///`) it will appear in rustdocs.

Also, can you please add a reference to https://blogs.msdn.microsoft.com/mast/2015/05/18/what-is-the-ip-address-168-63-129-16/?

src/providers/azure/mod.rs
```diff
@@ -179,7 +181,19 @@ impl Azure {
         .ok_or_else(|| "failed to get goal state: not found response".into())
     }
 
-    fn get_fabric_address() -> Result<IpAddr> {
+    fn get_fabric_address() -> IpAddr {
+        // try to fetch from dhcp, else use fallback; this is similar to what WALinuxAgent does
+        match Azure::get_fabric_address_from_dhcp() {
```
lucab: Tip: this is an `unwrap_or_else(|| ...)`.

I don't have particular style preferences in case, so pick the form you prefer.

lucab (commented):
Just two minor comments, looks ok otherwise. Thanks!

src/providers/azure/mod.rs
jlebon: Ahh of course, thanks. Used below! :arrow_down:

lucab (approved):

lucab:
@jlebon can you self-merge this here?

jlebon:
Doesn't look like it.

src/providers/azure/mod.rs
```diff
@@ -46,6 +46,10 @@ Content-Transfer-Encoding: base64
 
 ";
 
+/// This is a known working wireserver endpoint within Azure.
+/// See: https://blogs.msdn.microsoft.com/mast/2015/05/18/what-is-the-ip-address-168-63-129-16/
+const FALLBACK_WIRESERVER_ADDR: [u8; 4] = [168, 63, 129, 16]; // for grep: 168.63.129.16
```
bgilbert, line 51: > for grep: 168.63.129.16

:100:

bgilbert: Four years later, I did indeed end up finding this line with grep.  :tada:

PR #147, opened 2019-01-14
provider: add boot check-in on azure and packet

jlebon:
This is a respin of #130. So far it's mostly #130 rebased on top of #145 + minor fixes squashed in. A basic test in Azure works, though I'd like to test it some more. No testing done on Packet yet.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,6 +27,7 @@ extern crate reqwest;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde;
+#[cfg_attr(test, macro_use)]
 extern crate serde_json;
 extern crate serde_xml_rs;
 #[macro_use]
@@ -63,6 +64,7 @@ const CMDLINE_OEM_FLAG: &str = "coreos.oem.id";
 struct Config {
     provider: String,
     attributes_file: Option<String>,
+    check_in: bool,
     ssh_keys_user: Option<String>,
     hostname_file: Option<String>,
     network_units_dir: Option<String>,
@@ -110,6 +112,12 @@ fn run() -> Result<()> {
         .map_or(Ok(()), |x| metadata.write_network_units(x))
         .chain_err(|| "writing network units")?;
 
+    // perform boot check-in.
+    if config.check_in {
+        metadata.boot_checkin()
+            .chain_err(|| "checking-in instance boot to cloud provider")?;
+    }
+
     debug!("Done!");
 
     Ok(())
@@ -142,6 +150,9 @@ fn init() -> Result<Config> {
              .long("attributes")
              .help("The file into which the metadata attributes are written")
              .takes_value(true))
+        .arg(Arg::with_name("check-in")
+             .long("check-in")
+             .help("Check-in this instance boot with the cloud provider"))
         .arg(Arg::with_name("cmdline")
              .long("cmdline")
              .help("Read the cloud provider from the kernel cmdline"))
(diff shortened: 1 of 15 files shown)


jlebon:
OK, this is now tested working on both Packet and Azure. :heavy_check_mark:
Dropping WIP!

src/retry/client.rs
```diff
@@ -66,7 +67,7 @@ impl Deserializer for Json {
             .chain_err(|| "failed json deserialization")
     }
     fn content_type(&self) -> header::HeaderValue {
-        header::HeaderValue::from_static("text/json; charset=utf-8")
+        header::HeaderValue::from_static("application/json")
```
lucab, line 70: This is clearly the correct form, but I can just hope that the previous form was not there to workaround some buggy cloud provider.

src/retry/client.rs
```diff
@@ -175,14 +192,41 @@ impl<D> RequestBuilder<D>
             .chain_err(|| "failed to parse uri")?;
         let mut req = Request::new(Method::GET, url);
         req.headers_mut().extend(self.headers.clone().into_iter());
-        req.headers_mut().append(header::CONTENT_TYPE, self.d.content_type());
```
lucab, line 178: I guess this was supposed to be a `header::ACCEPT` originally, but I'm also fine if we prefer to just drop it for now.

src/main.rs
```diff
@@ -142,6 +149,9 @@ fn init() -> Result<Config> {
              .long("attributes")
              .help("The file into which the metadata attributes are written")
              .takes_value(true))
+        .arg(Arg::with_name("check-in")
```
lucab, line 152: I picked this initially as a random name, but I still had some minor doubts in my mind:
 1) is `--check-in` a good flag choice for this?
 2) in case we have some cloud platform in the future with a user-parameter, should this option take an argument too?

src/retry/client.rs
jlebon: Yeah, I had done some git history sleuthing around this. The original commit introducing this is 4b217bd71bc01a79dd01e1d89f4c998cfa5ed4a3, which... looks like this string might've just been adjusted from the `Raw` version of this block?

src/main.rs
jlebon: > is `--check-in` a good flag choice for this?

I personally like it. It's short and descriptive. Another possibility is `--phone-home` maybe? It matches the Packet terminology and might make it more obvious it's about *sending* data (whereas the other switches are about retrieving data).

> in case we have some cloud platform in the future with a user-parameter, should this option take an argument too?

The parameter would have to be specified through ct/Ignition, right? In which case, it'd probably be cleaner to pick it up from from an env var or a file instead? If it really comes down to adding CLI support for a parameter, I think I'd rather make it a new switch to be explicit, e.g. `--check-in-arg=..` or something.

lucab: Yes, it would be an additional parameter, I was thinking about the case where we lift it from `EnvinronmentFile` to the CLI, and whether to make this already future-proof for that.

Thanks for the feedback on both, it looks like the status quo is fine then.

lucab:
@jlebon LGTM, thanks for catching my mistakes! Any last minute thing before I merge this?

jlebon:
Nope, this is good to go!

lucab (approved):

mrguitar:
@lucab This is significant enough to warrant a new release. Can you guys create one?

lucab:
@mrguitar 3.1.0 in progress at https://github.com/coreos/coreos-metadata/pull/175.

PR #175, opened 2019-03-13
Release 3.1.0

lucab:
coreos-metadata 3.1.0:
 * provider: add boot check-in on azure and packet
 * azure: hardcode fallback for wireserver endpoint
 * providers/packet: minor code cleanup
 * cargo: update most dependencies to latest versions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,7 +176,7 @@ dependencies = [
 
 [[package]]
 name = "coreos-metadata"
-version = "3.1.0"
+version = "3.1.1-alpha.0"
 dependencies = [
  "base64 0.10.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,7 +176,7 @@ dependencies = [
 
 [[package]]
 name = "coreos-metadata"
-version = "3.0.3-alpha.0"
+version = "3.1.0"
 dependencies = [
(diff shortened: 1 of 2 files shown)


jlebon (approved):
LGTM!

imcleod:
LGTM!

ashcrow (approved):

ashcrow:
LGTM

arithx (approved):

jlebon:
So, this needs a final piece here for RHCOS/FCOS, which is the actual systemd unit that will replace the phone home service on CL. Should we include it in this repo?

imcleod:
> So, this needs a final piece here for RHCOS/FCOS, which is the actual systemd unit that will replace the phone home service on CL. Should we include it in this repo?

My ideal case, which may be mis-informed, is that once this update goes in, provided azure is identified as the oem id on the kernel command line, all the needfuls will happen on a *COS system to keep in online in Azure.  Likewise for packet.

If that means including the unit file, then I think my answer is yes.

ashcrow:
> If that means including the unit file, then I think my answer is yes.

I agree.

jlebon:
Yeah, definitely no doubt we need a systemd unit *somewhere*. My question was more about *where*. :) It'd be nice to keep it here, but I'd like to confirm the interactions with CL since we probably don't want to change how check-in works there at this point. (Or do we?)

lucab:
@jlebon I think that the service file could live here (next to the existing one). However I think that right now it is quicker to tag this, add the unit via RPM packaging, test everything and then move the unit to here once everything is confirmed ok. If no objections, I'll proceed with the release.

jlebon:
Sure, WFM!

PR #177, opened 2019-03-15
providers/gce: fix panic fetching metadata

bgilbert:
`http::header::HeaderName::from_static()` does not accept uppercase letters.  Introduced in b1e2a5ddbef0.

Fixes panic:

    thread 'main' panicked at 'invalid header name', [...]/http-0.1.14/src/header/name.rs:1660:23

### DIFF
--- a/src/providers/gce/mod.rs
+++ b/src/providers/gce/mod.rs
@@ -25,7 +25,7 @@ use network;
 use providers::MetadataProvider;
 use retry;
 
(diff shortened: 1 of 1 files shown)


lucab:
LGTM. I checked the rest of the code and this seems to be the only place where we do that. I'll need to add some basic tests for GCE later on.

lucab (approved):

PR #178, opened 2019-03-15
Drop merging of authorized_keys.d into authorized_keys

### LINKED ISSUE #157: Drop merging of authorized_keys.d into authorized_keys
# Feature Request #

## Environment ##

Any

## Desired Feature ##

At present, `coreos-metadata --ssh-keys` implicitly assumes that [update-ssh-keys](https://github.com/coreos/update-ssh-keys) is distributed as part of the OS.  That is, it assumes that SSH keys should be written into `~/.ssh/authorized_keys.d/<file>`, and then all files in `authorized_keys.d` should be merged into `authorized_keys`.  On distros without update-ssh-keys, this model doesn't immediately cause problems: coreos-metadata writes `~/.ssh/authorized_keys.d/coreos-metadata` and then performs the sync via an embedded copy of the update-ssh-keys library.  However, if the user updates `authorized_keys` directly and reboots, and `coreos-metadata --ssh-keys` runs again, it will clobber changes to `authorized_keys`.  Instead, the user is expected to put their keys in e.g. `~/.ssh/authorized_keys.d/my-keys`, but then doesn't have a tool to perform the synchronization.

Ultimately, update-ssh-keys is a hack for supporting `authorized_keys.d` directories without any cooperation from sshd, and should be dropped.  It changes the workflow of updating SSH keys and isn't used on other distros; Fedora CoreOS and Red Hat CoreOS currently do not intend to ship it.  Instead, in the short term, Fedora CoreOS will have special-case support for `authorized_keys.d` files written by individual tools, and in the medium term it will add generic support for the key directory (https://github.com/coreos/fedora-coreos-tracker/issues/139#issuecomment-459073242).

Modify coreos-metadata as follows:
- [x] After writing the `authorized_keys.d` fragment file, don't merge the fragment file into `authorized_keys` except on CL.  Probably gate this on a compile-time flag.
- [x] Document that `coreos-metadata --ssh-keys` will only work on distros that include `.ssh/authorized_keys.d/coreos-metadata` in sshd's `AuthorizedKeysFile`, or that otherwise arrange to read all keyfiles in `.ssh/authorized_keys.d`.

bgilbert:
Fixes #157.  Container Linux will need to build with `--features cl-legacy`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,8 +35,11 @@ slog-async = "2.1"
 slog-scope = "~4.0"
 slog-term = "2.2"
 tempdir = "0.3"
-update-ssh-keys = "^0.4.0"
+update-ssh-keys = { version = "^0.4.0", optional = true }
 users = "^0.8.1"
+
+[features]
+cl-legacy = ["update-ssh-keys"]
 
 [dependencies.slog]
 version = "2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ slog-async = "2.1"
 slog-scope = "~4.0"
 slog-term = "2.2"
 tempdir = "0.3"
+tempfile = "^3.0.7"
 update-ssh-keys = { version = "^0.4.0", optional = true }
 users = "^0.8.1"
 
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -37,6 +37,7 @@ use std::fs::{self, File};
 use std::io::prelude::*;
 use std::path::Path;
 
+use openssh_keys::PublicKey;
 use update_ssh_keys::{AuthorizedKeys, AuthorizedKeyEntry};
 use users;
 
@@ -58,7 +59,7 @@ fn create_file(filename: &str) -> Result<File> {
 pub trait MetadataProvider {
     fn attributes(&self) -> Result<HashMap<String, String>>;
     fn hostname(&self) -> Result<Option<String>>;
-    fn ssh_keys(&self) -> Result<Vec<AuthorizedKeyEntry>>;
+    fn ssh_keys(&self) -> Result<Vec<PublicKey>>;
     fn networks(&self) -> Result<Vec<network::Interface>>;
     fn network_devices(&self) -> Result<Vec<network::Device>>;
     fn boot_checkin(&self) -> Result<()>;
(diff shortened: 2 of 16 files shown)


src/providers/mod.rs
```diff
@@ -34,11 +34,19 @@ pub mod vagrant_virtualbox;
...
 
+use openssh_keys::PublicKey;
+use users::{self, User};
+#[cfg(feature = "legacy")]
 use update_ssh_keys::{AuthorizedKeys, AuthorizedKeyEntry};
-use users;
+#[cfg(not(feature = "legacy"))]
+use users::os::unix::UserExt;
+#[cfg(not(feature = "legacy"))]
+use tempfile;
```
lucab: I think you can plainly drop this.

src/providers/mod.rs
```diff
@@ -34,11 +34,19 @@ pub mod vagrant_virtualbox;
 
 use std::collections::HashMap;
 use std::fs::{self, File};
+#[cfg(not(feature = "legacy"))]
```
lucab: If you want to avoid all these conditionals below, you can move most of the `use` statements directly inside function body.

Cargo.toml
```diff
@@ -35,9 +35,13 @@ slog-async = "2.1"
 slog-scope = "~4.0"
 slog-term = "2.2"
 tempdir = "0.3"
-update-ssh-keys = "^0.4.0"
+tempfile = "^3.0.7"
+update-ssh-keys = { version = "^0.4.0", optional = true }
 users = "^0.8.1"
 
+[features]
+legacy = ["update-ssh-keys"]
```
lucab: Features are usually additive. Do we want to call this something more specific like `merge-authorized-keys`? If we are going to add more CL-legacy features, we can always add an umbrella one called `cl-legacy`. Thoughts?

lucab (commented):
Code looks fine already. I only have a couple nits and a doubt on feature naming.

src/providers/mod.rs
lucab: Discussed out-of-band: `tempfile::foo` is supported in newer compiler, but our baseline toolchain doesn't have that yet. Let's keep this.

Cargo.toml
lucab: Discussed out-of-band: this will likely ends up entangled in other things, such as project renaming. Let's stick to a monolithic `cl-legacy` to keep complexity low.

bgilbert:
Updated.

lucab (approved):
LGTM!

PR #181, opened 2019-03-19
providers/gce: add basic hostname mock-test

lucab:
This adds a simple mock-test covering hostname fetching.
It would prevent regressions like https://github.com/coreos/coreos-metadata/pull/177 in the future.


bgilbert (approved):
LGTM

PR #185, opened 2019-03-25
util/cmdline: rename "OEM ID" to "Platform ID"

lucab:
This switches from "coreos.oem.id" to "ignition.platform.id" in
non-legacy mode. 

Ref: https://github.com/coreos/coreos-assembler/issues/428

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -58,8 +58,8 @@ use std::env;
 use errors::*;
 use metadata::fetch_metadata;
 
+/// Path to kernel command-line (requires procfs mount).
 const CMDLINE_PATH: &str = "/proc/cmdline";
-const CMDLINE_OEM_FLAG: &str = "coreos.oem.id";
 
 #[derive(Debug)]
 struct Config {
@@ -199,7 +199,7 @@ fn init() -> Result<Config> {
             Some(provider) => String::from(provider),
             None => {
                 if matches.is_present("cmdline") {
-                    util::get_oem(CMDLINE_PATH, CMDLINE_OEM_FLAG)?
+                    util::get_platform(CMDLINE_PATH)?
                 } else {
                     return Err("Must set either --provider or --cmdline".into());
                 }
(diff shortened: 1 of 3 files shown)


dustymabe:
> This switches from "coreos.oem.id" to "coreos.platform.id"

shouldn't it be `ignition.platform.id` ? ref: https://github.com/coreos/ignition-dracut/pull/45

lucab:
@dustymabe indeed, I didn't see that PR. Renamed.

bgilbert:
Per https://github.com/coreos/coreos-assembler/pull/433#issuecomment-476177645, maybe we should go ahead and accept only the new flag in non-legacy mode.

dustymabe:
> Per [coreos/coreos-assembler#433 (comment)](https://github.com/coreos/coreos-assembler/pull/433#issuecomment-476177645), maybe we should go ahead and accept only the new flag in non-legacy mode.

That makes sense. IOW the media creator will generate both `coreos.oem.id` and `ignition.platform.id` so the consumer software doesn't have to handle both cases.

jlebon (approved):
LGTM!

No strong opinion on whether we should handle both here too or not. Not handling both just means that we'll have to be sure that the RHCOS pipeline has https://github.com/coreos/coreos-assembler/pull/433 backported before the next cut.

lucab:
I'm going to hold this a bit until we sort out the approach in https://github.com/coreos/coreos-assembler/pull/433. I'm fine either ways, I thought that an overlap window here would make easier the lockstep transition on RHCOS side.

lucab:
https://github.com/coreos/coreos-assembler/pull/433 landed adding both flags. As requested by @bgilbert, I reworked this PR to drop the overlapping/legacy fallback. PTAL, thanks!

jlebon (approved):
This looks good to me!

bgilbert (approved):
LGTM

PR #190, opened 2019-03-28
providers/gcp: drop unused import in tests

bgilbert:


lucab (approved):

lucab:
I guess we can hold this a bit, proceed with the release at https://github.com/coreos/afterburn/pull/191 (which does not include this), and then rebase after that.

PR #192, opened 2019-03-29
Release coreos-metadata 3.1.2 to add deprecation notice to crates.io

bgilbert:
I'll push the tag after the PR merges, but I'm not planning to create a GitHub Release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,7 +176,7 @@ dependencies = [
 
 [[package]]
 name = "coreos-metadata"
-version = "3.1.1-alpha.0"
+version = "3.1.2-alpha.0"
 dependencies = [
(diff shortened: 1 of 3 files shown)


lucab (approved):

bgilbert:
[Published to crates.io](https://crates.io/crates/coreos-metadata).

lucab:
It looks like I didn't pay enough attention, and we ended up plainly skipping `3.1.1`.

bgilbert:
That was intentional.  3.1.2 doesn't include any of the commits from the 3.1.1-alpha period, so I thought it would be better to avoid confusion.

lucab:
I see. It makes sense from git-log point of view; I didn't notice initially and just saw the tag progression gap.

PR #201, opened 2019-04-09
Add checkin service files for Azure and Packet

ashcrow:
Adds the service unit for on boot check in.

/cc @imcleod @crawford @jlebon @lucab

### DIFF
--- a/systemd/afterburn-checkin.service
+++ b/systemd/afterburn-checkin.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Afterburn (Check In)
+ConditionKernelCommandLine=|ignition.platform.id=azure
+After=boot-complete.target
+
+[Service]
+ExecStart=/usr/bin/afterburn --cmdline --check-in
+Type=oneshot
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
--- a/systemd/afterburn-firstboot-checkin.service
+++ b/systemd/afterburn-firstboot-checkin.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Afterburn (Firstboot Check In)
+ConditionKernelCommandLine=|ignition.platform.id=packet
+ConditionFirstBoot=yes
+After=boot-complete.target
+
+[Service]
+ExecStart=/usr/bin/afterburn --cmdline --check-in
+Type=oneshot
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target


systemd/afterburn-checkin.service
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Descitpion=Afterburn Check In
```
bgilbert: `Description`

nit: For consistency with the other units, should be e.g. `Afterburn (Check In)`.

systemd/afterburn-checkin.service
```diff
@@ -0,0 +1,8 @@
...
+After=network-online.target
```
bgilbert: This can be dropped.  `network-online.target` is documented to be unreliable, and Afterburn should be automatically retrying requests if the network isn't up yet.

systemd/afterburn-checkin.service
```diff
@@ -0,0 +1,8 @@
...
+
```
bgilbert, line 5: We'll also want:

1. `After` greenboot is happy.
2. `ConditionKernelCommandLine=|ignition.platform.id=azure`

The first is harmless on distros that don't have greenboot.  Afterburn already reads `ignition.platform.id` so the second shouldn't further affect distro-independence.

systemd/afterburn-checkin.service
ashcrow: :+1: will update

systemd/afterburn-checkin.service
ashcrow: :+1: will update

systemd/afterburn-checkin.service
ashcrow: :+1: will add

ashcrow: After speaking with @LorbusChris I put in:

```
After=boot-complete.target
```

systemd/afterburn-checkin.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Descitpion=Afterburn (Check In)
```
jlebon: `Description`

systemd/afterburn-checkin.service
LorbusChris: We can`After=boot-complete.target` (and probably `Requires=boot-complete.target` here, which is what greenboot does, when installed)
IIRCthe boot-complete.target unit is usually not activated - this will activate it however, and by default there is nothing blocking its activation then (like a health-check runner a la greenboot would do on failure)

systemd/afterburn-checkin.service
```diff
@@ -0,0 +1,9 @@
...
+ConditionKernelCommandLine=|ignition.platform.id=azure
```
jlebon, line 3: Also `ConditionKernelCommandLine=|ignition.platform.id=packet`, right?

bgilbert: @jlebon Packet only needs a checkin on first boot.  We could have a separate `afterburn-checkin-firstboot.service` with:

```
ConditionFirstBoot=true
ConditionKernelCommandLine=|ignition.platform.id=packet
```

I'd probably go that route, but there's also an argument that the separate unit isn't worth the trouble.

ashcrow: Added

crawford (commented):
Do we need an `[Install]` section and ConditionFirstBoot?

bgilbert:
Whoops, we do need an `[Install]` section.

Azure needs to run on every boot (well, on some boots).  I'm still partial to a separate `ConditionFirstBoot` unit for Packet, but @ashcrow, it looks like you decided to go the other way?

ashcrow:
> I'm still partial to a separate ConditionFirstBoot unit for Packet, but @ashcrow, it looks like you decided to go the other way?

I'm ok with creating a secondary unit for packet. Will update with `[Install]` and splitting the unit.

ashcrow:
Updated based on latest reviews.

jlebon:
Maybe let's name the services `afterburn-firstboot-checkin.service` and `afterburn-checkin.service`? One could imagine growing the list in either of those as we teach Afterburn about more platforms that need check-ins.

ashcrow:
Updated.

bgilbert (approved):
:100:

jlebon (approved):
LGTM!

crawford (commented):
Does Azure need to check in on every boot? It probably doesn't hurt to do it that way, but I was under the impression that it only needed to report success one time.

arithx:
@crawford Azure needs to check in every boot that is on a different host iirc. But there is no harm in checking in every boot so it's easier to check-in every boot than to detect that.

crawford:
@arithx Ahhh, I forgot about that restriction. LGTM

ashcrow:
*bump* :-)

ashcrow:
This is ready for merge.

PR #204, opened 2019-04-17
cargo: switch to 2018 edition

lucab:
This runs `cargo fix --edition` on the whole project, then switches it to 2018 edition.


jlebon (approved):
Hmm, might need a pass through `cargo fmt` since some the `use` orders likely changed, otherwise LGTM!

lucab:
@jlebon indeed, some imports were out of order. Fixed by a `cargo fmt` run, thanks for catching that.

PR #224, opened 2019-06-06
dracut: add afterburn dracut module

yuqi-zhang:
Add dracut module to run afterburn in the initramfs, and add an
initial service to fetch hostname on necessary providers (Azure).

### DIFF
--- a/dracut/30afterburn/afterburn-hostname.service
+++ b/dracut/30afterburn/afterburn-hostname.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=Afterburn Hostname
+# Azure needs to fetch the hostname in the initramfs
+ConditionKernelCommandLine=|ignition.platform.id=azure
+
+# We order this service after sysroot has been mounted
+# but before ignition-files stage has run, so ignition can
+# overwrite the configured hostname with e.g. a static one
+# Also order after ignition-remount-sysroot just for
+# safety for systems that mount /sysroot ro
+After=initrd-root-fs.target
+Before=ignition-files.service
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+ExecStart=/usr/bin/afterburn --cmdline --hostname=/sysroot/etc/hostname
+Type=oneshot
--- a/dracut/30afterburn/module-setup.sh
+++ b/dracut/30afterburn/module-setup.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+check() {
+    return 0
+}
+
+depends() {
+    echo systemd network
+}
+
+install() {
+    inst_multiple afterburn
+
+    inst_simple "$moddir/afterburn-hostname.service" \
+        "$systemdutildir/system/afterburn-hostname.service"
+
+    # We want the afterburn-hostname to be firstboot only, so Ignition-provided
+    # hostname changes do not get overwritten on subsequent boots
+
+    mkdir -p "$initdir/$systemdsystemunitdir/ignition-complete.target.requires"
+    ln -s "../afterburn-hostname.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-hostname.service"
+}


yuqi-zhang:
Also see discussion in: https://github.com/yuqi-zhang/afterburn-dracut/pull/1

There is the question of whether we want this to run on the firstboot or all boots. I think the hybrid approach to conditionally track Azure's hostname iif no Ignition hostname is specified is possibly the most clean, but I'm not sure what the best way to detect that would be.

cc @jlebon @ajeddeloh @lucab

ashcrow (commented):
Looks sane to me

jlebon:
LGTM modulo the pending discussions around ordering & firstboot semantics.

lucab:
@jlebon has a good point, so I'm backtracking on my initial comment. Let's have the initramfs part first-boot only.

yuqi-zhang:
I read around a bit and I think it mostly makes sense to keep this module as a firstboot initramfs hostname fetch. As for future boots/runtime hostname changes, I don't really foresee dhcp leased hostnames changes (and if we do, I think Azure handles it from what I can tell?), and static configurations are ignition based.

As for ordering, the `ignition-files` will be dropping static confs if we want to overwrite, so I think the present setup is correct.

Please let me know if either of those two understandings sound incorrect.

dracut/30afterburn/afterburn-hostname.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Afterburn Hostname
+# Azure needs to fetch the hostname in the initramfs
+ConditionKernelCommandLine=|ignition.platform.id=azure
+ConditionFirstBoot=yes
+After=ignition-disks.service
+Before=ignition-files.service
```
jlebon, line 12: Minor: maybe a comment there that we're purposely ordering before `ignition-files.service` so that the latter can still override the hostname if wanted?

jlebon (commented):
Just one small comment, otherwise LGTM!

dracut/30afterburn/afterburn-hostname.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Afterburn Hostname
+# Azure needs to fetch the hostname in the initramfs
+ConditionKernelCommandLine=|ignition.platform.id=azure
+ConditionFirstBoot=yes
+After=ignition-disks.service
```
lucab: @yuqi-zhang did you test this service unit on an actual Azure machine? Because I think I made a mistake and this should be `After=ignition-mount.service`.

jlebon: I guess since it only writes to `/sysroot/etc`, technically it just needs `After=initrd-root-fs.target`. That'll work on both FCOS and RHCOS (which doesn't have `ignition-mount.service` yet).

dracut/30afterburn/afterburn-hostname.service
yuqi-zhang: Added some comments

dracut/30afterburn/afterburn-hostname.service
yuqi-zhang: I added `After=initrd-root-fs.target`
`Wants=initrd-root-fs.target` based on https://github.com/coreos/ignition-dracut/blob/spec2x/dracut/30ignition/coreos-static-network.service

lucab: Ack. And do we care about `ignition-remount-sysroot.service` here? (I'm not so familiar with the new Ignition flow in initramfs)

jlebon: > And do we care about `ignition-remount-sysroot.service` here?

It doesn't hurt, but shouldn't make a difference either, see: https://github.com/coreos/ignition-dracut/pull/38#issuecomment-451965981.

yuqi-zhang: I suppose for safety we should? I'm also not sure on which systems that one actually runs.
And to your earlier point, I didn't test on Azure since I didn't have creds at the time, I made a mock unit with the same dependency tree to run on metal, and it worked as expected.

dracut/30afterburn/afterburn-hostname.service
```diff
@@ -0,0 +1,24 @@
...
+
+After=initrd-root-fs.target
+Wants=initrd-root-fs.target
+
+# We order this service after disks have been provisioned
+# but before ignition-files stage has run, so ignition can
+# overwrite the configured hostname with e.g. a static one
+After=ignition-disks.service
+Before=ignition-files.service
+
```
jlebon, line 13: This is maybe more personal preference, but I like to cut down the `After`/`Before` directives to what is actually necessary to make it clearer what the dependencies really are. So e.g. here, we should *only* need `After=initrd-root-fs.target` and `Before=ignition-files.service`. (We don't need `After=ignition-disks.service`, since clearly that must happen before `/sysroot` is mounted, and we don't need `Wants=initrd-root-fs.target` since that's already part of the transaction.)

That said, if you've already validated this works, I'm cool rolling with it like that as well!

yuqi-zhang: I agree, I just quickly added those in. Let me clean it up a bit, thanks for the suggestion!

yuqi-zhang: On that note, do you know where we're using https://github.com/coreos/ignition-dracut/blob/spec2x/dracut/30ignition/ignition-remount-sysroot.service ? Is that something we should look out for

dracut/30afterburn/afterburn-hostname.service
yuqi-zhang: Added

dracut/30afterburn/afterburn-hostname.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Afterburn Hostname
+# Azure needs to fetch the hostname in the initramfs
+ConditionKernelCommandLine=|ignition.platform.id=azure
+
+# This will be firstboot only to set up the correct
+# hostname for kube.
+ConditionFirstBoot=yes
```
jlebon: One question, does this actually work correctly in the initrd? For the real root, systemd keys off of `/etc/machine-id`, but here it has to be able to make that determination before `/sysroot` is even mounted. So does it key off of the *initrd*'s `/etc/machine-id`? Except that check will never change result unless we regen the initrd.

An alternative here is to hook into `ignition-complete.target`, which was made for this: https://github.com/coreos/ignition-dracut/blob/df88988f2f0791ca9ea1e14298c8523501d980c5/dracut/30ignition/ignition-complete.target#L3.

For RHCOS, since it doesn't have that target yet, you could do e.g. `ignition-files.service.requires`?

dracut/30afterburn/afterburn-hostname.service
jlebon: We don't need it in RHCOS and FCOS, see https://github.com/coreos/ignition-dracut/commit/3ec0b39106a0724982da64dc348a38da7a471467. It was added back for other distros, see https://github.com/coreos/ignition-dracut/issues/37 and https://github.com/coreos/ignition-dracut/pull/38#issuecomment-451965981.

dracut/30afterburn/afterburn-hostname.service
ajeddeloh: Afterburn does not depend on Ignition though, so we probably shouldn't make it use that target. I agree though that using `ConditionFirstBoot` from the initramfs seems sketchy. Not sure the best path forward here. I think this is a case where the distro-dependence is showing.

yuqi-zhang: > Afterburn does not depend on Ignition though, so we probably shouldn't make it use that target.

We explicitly reference ignition targets, I don't really see a harm in cross-linking per se? Unless you want them to be as separate as possible.

I did some testing with a metal-bios image. If I `rd.break` into the initramfs on the first boot, I see the unit as inactive, and a message that: `Condition: start condition failed` (and same with subsequent boots). Interesting if I try to `systemctl start` the process manually, it will add a line `CondtionFirstBoot=yes was not met`. I'm not sure if that's the expected behaviour or not (i.e. which condition is not being met).

In either case, this does not seem to be very robust... should we

jlebon: > Afterburn does not depend on Ignition though, so we probably shouldn't make it use that target.

I'd say this unit is inherently dependent on Ignition already by simply having `ConditionKernelCommandLine=|ignition.platform.id=azure`. :)

We could do something fancy here and "connect" the two in the overlay. Though unless there's demand for running Afterburn without Ignition, I'm not sure if it's worth the investment.

ajeddeloh: I think this is fine now that I think about it. Part of me still just wants one giant fcos-dracut repo, but I think that ship has sailed.

yuqi-zhang:
Ok I've updated it to depend on ignition and tested it to work (firstboot only)

For RHCOS I think we want to carry a patch to make it:
```
mkdir -p "$initdir/$systemdsystemunitdir/ignition-files.service.requires"
    ln -s "../afterburn-hostname.service" "$initdir/$systemdsystemunitdir/ignition-files.service.requires/afterburn-hostname.service"
```
for now, since I don't really see a reason to branch.
PTAL.

dracut/30afterburn/module-setup.sh
```diff
@@ -0,0 +1,24 @@
...
+}
+
+install() {
+    inst_multiple afterburn
+
+    inst_simple "$moddir/afterburn-hostname.service" \
+        "$systemdutildir/system/afterburn-hostname.service"
+
+    # We want the afterburn-hostname to be firstboot only, so ignition-provided
+    # hostname changes do not get overwritten on subsequentboots
```
jlebon: Minor: `subsequent boots`
Super minor: `Ignition`

jlebon (commented):
Just some trivial nits, otherwise LGTM!

dracut/30afterburn/module-setup.sh
yuqi-zhang: Fixed up, thanks for the review!

yuqi-zhang:
@jlebon Should be good to go. I also opened https://src.fedoraproject.org/rpms/rust-afterburn/pull-request/1 to show how to pull it in once we release a new crate

jlebon (approved):
Yup this LGTM!
I don't have merge access, will leave it to someone else to have a final look and merge.

lucab (approved):

PR #244, opened 2019-07-15
systemd: schedule checkin after network target

lucab:
Cloud checkin requires networking to be already started, thus this
schedules checkin units after network target.
This adds no additional guarantees on network configuration status,
and afterburn still retries (with backoff) multiple times internally.

### DIFF
--- a/systemd/afterburn-checkin.service
+++ b/systemd/afterburn-checkin.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Afterburn (Check In)
 ConditionKernelCommandLine=|ignition.platform.id=azure
+After=network.target
 After=boot-complete.target
 
 [Service]
--- a/systemd/afterburn-firstboot-checkin.service
+++ b/systemd/afterburn-firstboot-checkin.service
@@ -2,6 +2,7 @@
 Description=Afterburn (Firstboot Check In)
 ConditionKernelCommandLine=|ignition.platform.id=packet
 ConditionFirstBoot=yes
+After=network.target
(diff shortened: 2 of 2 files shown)


systemd/afterburn-checkin.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Afterburn (Check In)
 ConditionKernelCommandLine=|ignition.platform.id=azure
+After=network.target
```
yuqi-zhang, line 4: I think it should depend on After=network-online.target instead

yuqi-zhang (commented):
Makes sense, and shouldn't affect anything either way since I doubt if the network isn't up during the 4 minute timeout, it will ever come up successfully. Just one question on the target

systemd/afterburn-checkin.service
ajeddeloh: https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

Maybe not. If afterburn retries, `network.target` is probably correct.

yuqi-zhang: Hmm fair, then LGTM

lucab: Indeed, this is only to give a change to NM/sd-networkd to start before afterburn.
There is no way to guarantee that the network is stable before running the checkin logic.
Afterburn interally retries (with backoff) to be resilient to that.

yuqi-zhang (approved):
LGTM

PR #245, opened 2019-07-16
dracut: relabel the hostname file

yuqi-zhang:
Mark the file as needing relabelling, otherwise the hostname
file written to by the afterburn-hostname.service will get
SELinux denials.

This is somewhat of a hack. Eventually we want to relabel in the initrd. See: https://github.com/coreos/coreos-assembler/pull/551#issuecomment-500852247

cc @jlebon

### DIFF
--- a/dracut/30afterburn/afterburn-hostname.service
+++ b/dracut/30afterburn/afterburn-hostname.service
@@ -16,4 +16,9 @@ OnFailureJobMode=isolate
 
 [Service]
 ExecStart=/usr/bin/afterburn --cmdline --hostname=/sysroot/etc/hostname
+# Add hack to mark the file as needing relabelling, as the hostname
+# file dropped by afterburn will be unlabelled causing SELinux denials.
+# see: https://github.com/coreos/ignition/issues/635
+ExecStart=/bin/sh -c 'mkdir -p /run/tmpfiles.d'
+ExecStart=/bin/sh -c 'echo "z /etc/hostname - - -" > /run/tmpfiles.d/hostname-relabel.conf'
 Type=oneshot


dracut/30afterburn/afterburn-hostname.service
```diff
@@ -16,4 +16,6 @@ OnFailureJobMode=isolate
 
 [Service]
 ExecStart=/usr/bin/afterburn --cmdline --hostname=/sysroot/etc/hostname
+ExecStart=/bin/sh -c 'mkdir -p /sysroot/run/tmpfiles.d'
+ExecStart=/bin/sh -c 'echo "Z /etc/hostname - - -" > /sysroot/run/tmpfiles.d/hostname-relabel.conf'
```
jlebon: Minor: just needs `z` there. `Z` is for recursive labeling.

dracut/30afterburn/afterburn-hostname.service
```diff
@@ -16,4 +16,6 @@ OnFailureJobMode=isolate
 
 [Service]
 ExecStart=/usr/bin/afterburn --cmdline --hostname=/sysroot/etc/hostname
+ExecStart=/bin/sh -c 'mkdir -p /sysroot/run/tmpfiles.d'
```
jlebon: This should be just `/run/tmpfiles.d`. The `/run` tmpfs is carried over from the initrd at switchroot.

jlebon: Also, maybe throw a comment to e.g. https://github.com/coreos/ignition/issues/635? (Or a more targeted issue if there is one).

dracut/30afterburn/afterburn-hostname.service
jlebon: You'll likely also need to order whatever problematic services to *after* `systemd-tmpfiles-setup.service`. See e.g. https://github.com/coreos/fedora-coreos-config/blob/1842e9e309c2dfc89863705917c19f8eb85be245/overlay/usr/lib/dracut/modules.d/40coreos-var/coreos-populate-var.sh#L39. Depending on the service, that might be tricky to do.

jlebon (commented):
At least for Fedora, you could use systemd's `/run/systemd/relabel-extra.d/`: https://github.com/systemd/systemd/pull/11131. Though there's value in keeping the same workaround for FCOS and RHCOS too.

dracut/30afterburn/afterburn-hostname.service
yuqi-zhang: Shouldn't be an issue since AFAIK only NM should be using this

yuqi-zhang:
Pushed some changes and tested to be in a working state. Thanks for the fast review!

dracut/30afterburn/afterburn-hostname.service
yuqi-zhang: Will do

yuqi-zhang:
@jlebon is this ok to merge? I'd like to patch this into RHCOS for Azure

dracut/30afterburn/afterburn-hostname.service
jlebon: Hmm, does NM have an explicit ordering to `systemd-tmpfiles-setup.service`?

dracut/30afterburn/afterburn-hostname.service
jlebon: Heh, I meant a comment in this file to that GitHub issue, but consider it optional.

dracut/30afterburn/afterburn-hostname.service
jlebon: Well, none of the NM units use `DefaultDependencies=no`, so it should already be ordered after `sysinit.target` (which pulls in `systemd-tmpfiles-setup.service`), so yeah, I think we're fine.

jlebon (approved):
LGTM. Optional suggestion to add a comment, but good as is too.

yuqi-zhang:
Ah my bad, added comment.
(I also don't have merge access to the repo, but feel free to merge)

lucab:
LGTM, deferring to @jlebon for the final stamp+merge.

Issue #251, opened 2019-07-19
docs: mention platform naming skew


lucab:
Afterburn usually follows Ignition platform naming convention, in order to auto-detect the platform.

However there are several platforms where the metadata component is only optional. On those platforms, in case of failures, Afterburn cannot distinguish a known-good state (i.e. this infra lacks the metadata component) from a failed state (i.e. network or logic error).

As such, Afterburn knows about the following platform which are more specific than the Ignition ones, cannot be auto-detected, and require user-provided hints (as cmdline flags):
 * `cloudstack-metadata`
 *  `cloudstack-configdrive`
 * `openstack-metadata`
 * `vagrant-virtualbox`

lucab:
This follows up to @cgwalters's doubt from https://github.com/openshift/machine-config-operator/pull/964#issuecomment-511036266.

bgilbert:
`openstack-metadata` has since been consolidated into a platform ID `openstack` that does match Ignition.  `vagrant-virtualbox` has been removed, and https://github.com/coreos/afterburn/issues/856 tracks consolidation of the CloudStack providers into a `cloudstack` one.  Closing this in favor of fixing CloudStack to use the Ignition platform ID.

PR #253, opened 2019-07-19
systemd: unify provider overriding via env

lucab:
This unifies all non-initramfs units so that the provider flag
can be always overridden via `${AFTERBURN_OPT_PROVIDER}`.


lucab:
For reference, see https://github.com/coreos/afterburn/issues/251.

jlebon (approved):

PR #278, opened 2019-10-01
providers: fetch instance types for all providers

### LINKED ISSUE #277: providers: Fetch instance-type metadata
# Feature Request #
Since in `fedora-coreos-pinger`, we need to fetch the instance type of cloud providers and I am trying to avoid too much duplicate code between fedora-coreos-pinger, Afterburn and Zincati, it would be great if we could fetch cloud instance type here.

## Environment ##

Apply to all cloud providers.

## Desired Feature ##

Fetch `cloud instance type` for all providers.

## Other Information ##
Link to related fedora-coreos-pinger issue: https://github.com/coreos/fedora-coreos-pinger/issues/30

zonggen:
Fetch instance-type from cloud providers.

Closes: https://github.com/coreos/afterburn/issues/277
Signed-off-by: Zonggen (Allen) Bai <abai@redhat.com>

### DIFF
--- a/src/providers/aws/mock_tests.rs
+++ b/src/providers/aws/mock_tests.rs
@@ -1,13 +1,7 @@
 use crate::errors::*;
 use crate::providers::aws;
 use mockito;
-use std::collections::HashMap;
 use crate::providers::MetadataProvider;
-
-#[cfg(not(feature = "cl-legacy"))]
-static ENV_PREFIX: &str = "AWS";
-#[cfg(feature = "cl-legacy")]
-static ENV_PREFIX: &str = "EC2";
 
 #[test]
 fn test_aws_basic() {
@@ -35,40 +29,46 @@ fn test_aws_basic() {
 
 #[test]
 fn test_aws_attributes() {
-    let ep_instance_id = "/meta-data/instance-id";
     let instance_id = "test-instance-id";
-
-    let ep_instance_type = "/meta-data/instance-type";
     let instance_type = "test-instance-type";
-
-    let ep_ipv4_local = "/meta-data/local-ipv4";
     let ipv4_local = "test-ipv4-local";
-
-    let ep_ipv4_public = "/meta-data/public-ipv4";
     let ipv4_public = "test-ipv4-public";
-
-    let ep_availability_zone = "/meta-data/placement/availability-zone";
     let availability_zone = "test-availability-zone";
-
-    let ep_hostname = "/meta-data/hostname";
     let hostname = "test-hostname";
-
-    let ep_public_hostname = "/meta-data/public-hostname";
     let public_hostname = "test-public-hostname";
-
-    let ep_region = "/dynamic/instance-identity/document";
-    let instance_id_doc = "{\"region\": \"test-region\"}";
+    let instance_id_doc = r#"{"region": "test-region"}"#;
     let region = "test-region";
 
-    let mut attributes:HashMap<String, String> = HashMap::new();
-    attributes.insert(format!("{}_INSTANCE_ID", ENV_PREFIX), String::from(instance_id));
-    attributes.insert(format!("{}_INSTANCE_TYPE", ENV_PREFIX), String::from(instance_type));
-    attributes.insert(format!("{}_IPV4_LOCAL", ENV_PREFIX), String::from(ipv4_local));
-    attributes.insert(format!("{}_IPV4_PUBLIC", ENV_PREFIX), String::from(ipv4_public));
-    attributes.insert(format!("{}_AVAILABILITY_ZONE", ENV_PREFIX), String::from(availability_zone));
-    attributes.insert(format!("{}_HOSTNAME", ENV_PREFIX), String::from(hostname));
-    attributes.insert(format!("{}_PUBLIC_HOSTNAME", ENV_PREFIX), String::from(public_hostname));
-    attributes.insert(format!("{}_REGION", ENV_PREFIX), String::from(region));
+    let endpoints = maplit::btreemap! {
+        "/meta-data/instance-id" => instance_id,
+        "/meta-data/instance-type" => instance_type,
+        "/meta-data/local-ipv4" => ipv4_local,
+        "/meta-data/public-ipv4" => ipv4_public,
+        "/meta-data/placement/availability-zone" => availability_zone,
+        "/meta-data/hostname" => hostname,
+        "/meta-data/public-hostname" => public_hostname,
+        "/dynamic/instance-identity/document" => instance_id_doc,
+    };
+
+    let mut mocks = Vec::with_capacity(endpoints.len());
+    for (endpoint, body) in endpoints {
+        let m = mockito::mock("GET", endpoint)
+            .with_status(200)
+            .with_body(body)
+            .create();
+        mocks.push(m);
+    }
+
+    let attributes = maplit::hashmap! {
+        format!("{}_INSTANCE_ID", aws::ENV_PREFIX) => String::from(instance_id),
+        format!("{}_INSTANCE_TYPE", aws::ENV_PREFIX) => String::from(instance_type),
+        format!("{}_IPV4_LOCAL", aws::ENV_PREFIX) => String::from(ipv4_local),
+        format!("{}_IPV4_PUBLIC", aws::ENV_PREFIX) => String::from(ipv4_public),
+        format!("{}_AVAILABILITY_ZONE", aws::ENV_PREFIX) => String::from(availability_zone),
+        format!("{}_HOSTNAME", aws::ENV_PREFIX) => String::from(hostname),
+        format!("{}_PUBLIC_HOSTNAME", aws::ENV_PREFIX) => String::from(public_hostname),
+        format!("{}_REGION", aws::ENV_PREFIX) => String::from(region),
+    };
 
     let client = crate::retry::Client::try_new()
         .chain_err(|| "failed to create http client")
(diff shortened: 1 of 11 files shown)


zonggen:
Ready for review!

src/providers/azure/mod.rs
```diff
@@ -418,12 +418,28 @@ impl Azure {
             .chain_err(|| "failed to get hostname")?;
         Ok(name)
     }
+
+    fn fetch_sku (&self) -> Result<Option<String>> {
```
lucab: Why is there an inner `Option<T>` here? Can the SKU be missing/empty?
I think we can either return a plain `Result<T>` (if the SKU should always be present) or an `Option<T>` (if the SKU can be missing/empty).

zonggen: The logic I took is the same as `fetch_host` on line# 407.
From my understanding, the return type is coming from https://github.com/coreos/afterburn/blob/59d3c56ccefe093ec98d5aecb2579b88b7b5e9b5/src/retry/client.rs#L195 which also depends on https://github.com/coreos/afterburn/blob/59d3c56ccefe093ec98d5aecb2579b88b7b5e9b5/src/retry/client.rs#L238 the request could fail so the outer `Result<_>` and if response is 404 then return `Ok<None>` hence the inner `Option<String>`.. Am I missing something?

lucab: No, you are not missing anything. But the client API you are quoting is a generic HTTP client with retries, which also supports pages with empty bodies (thus the `Result<Option<T>>` return type).
In this specific case, I think you can map the result to a simpler `Result<String>`.

src/providers/azure/mod.rs
```diff
@@ -418,12 +418,28 @@ impl Azure {
...
+        const SKU_URL: &str = "metadata/instance/compute/sku?api-version=2017-08-01&format=text";
```
lucab: For your task, I think you want to use the `vmSize` instead of the `sku` here.

lucab:
@zonggen several of these providers have dedicated mock-tests. It would be helpful if you could cover the "fetch instance type" logic for the existing ones.

src/providers/azure/mod.rs
zonggen: Ah, yup I think you'are right, for future reference:
 - [vmSize_example](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/sizes)
 - [sku_example](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/cli-ps-findimage#table-of-commonly-used-windows-images)

zonggen:
[>>>>>>>>................] Updating according to comments and adding tests
EDIT:
[>>>>>>>>>>>>>>> ] Updated! Added test to existing `mock-tests` files!

src/providers/azure/mod.rs
zonggen: I see, thanks for explaining!

zonggen:
Ready for another round of review!

src/providers/aws/mock_tests.rs
```diff
@@ -1,6 +1,13 @@
 use crate::errors::*;
 use crate::providers::aws;
 use mockito;
+use std::collections::HashMap;
+use crate::providers::MetadataProvider;
+
+#[cfg(not(feature = "cl-legacy"))]
+static ENV_PREFIX: &str = "AWS";
```
lucab: I think (but I didn't try) that you can directly use those from `crate::providers::aws`. If that doesn't directly work, additionally marking them as `pub(crate)` probably would help.

src/providers/aws/mock_tests.rs
```diff
@@ -25,3 +32,85 @@ fn test_aws_basic() {
...
+        .with_body(public_hostname)
+        .create();
+    let _m = mockito::mock("GET", ep_region)
+        .with_status(200)
+        .with_body(instance_id_doc)
+        .create();
+
+    let v = provider.attributes().unwrap();
+
+    assert_eq!(v, attributes);
```
lucab, line 84: This is a very extensive test, thanks for that!

src/providers/azure/mod.rs
```diff
@@ -17,7 +17,9 @@
 mod crypto;
 
 use std::collections::HashMap;
-use std::net::{IpAddr, SocketAddr};
+use std::net::IpAddr;
+#[cfg(not(test))]
+use std::net::SocketAddr;
```
lucab: As it is used only once, you could also directly move this inside the relevant `get_attributes`. As you prefer.

src/providers/gcp/mock_tests.rs
```diff
@@ -1,7 +1,13 @@
 use crate::providers::gcp;
 use crate::providers::MetadataProvider;
+use std::collections::HashMap;
 use mockito;
 
+#[cfg(not(feature = "cl-legacy"))]
+static ENV_PREFIX: &str = "GCP";
```
lucab: Same comment as the `aws` one.

src/providers/packet/mock_tests.rs
```diff
@@ -34,3 +35,41 @@ fn test_boot_checkin() {
     mock.assert();
     r.unwrap();
 }
+
+#[test]
+fn test_packet_attributes() {
+    let metadata = "{
```
lucab: You can define a raw string as `let metadata = r#"YOUR_CONTENT"#;` so that you don't have to manually escape inner quotes.

src/providers/aws/mock_tests.rs
zonggen: It works without `pub(crate)`.
I think it works because they are from the same module and variables in `mod.rs` is visible to `mock-tests.rs`

src/providers/aws/mock_tests.rs
zonggen: No problem! :+1:

src/providers/azure/mod.rs
zonggen: Sure, will update to see how it looks.

src/providers/packet/mock_tests.rs
zonggen: :+1: Lesson learnt..
side note: manually escaping quotes in json file is not fun :(

zonggen:
Use of `maplit` makes testing so much cleaner!

Reference:
https://github.com/coreos/afterburn/blob/b421699817cb21605ac2a856cdb31d9b7fd12417/src/providers/aliyun/mock_tests.rs#L81

zonggen:
Ready for review!

lucab:
@zonggen I think you are hitting a race/flake on GCP. https://github.com/coreos/afterburn/pull/286 should help resolving that.

src/providers/packet/mock_tests.rs
```diff
@@ -34,3 +34,42 @@ fn test_boot_checkin() {
...
+    let _m = mockito::mock("GET", "/metadata")
+        .with_status(200)
+        .with_body(metadata)
+        .create();
+
+    let provider = packet::PacketProvider::try_new().unwrap();
+    let v = provider.attributes().unwrap();
+
+    assert_eq!(v, attributes);
+}
```
lucab, line 77: Let's throw a `mockito::reset()` at the bottom here, and a `try_new` + `attributes().unwrap_err()` too.

src/providers/packet/mock_tests.rs
```diff
@@ -34,3 +34,42 @@ fn test_boot_checkin() {
...
+        },
+        "phone_home_url": "test-url"
+    }"#;
+
+    let hostname = "test-hostname";
+    let phone_home_url = "test-url";
+    let plan = "test-plan";
+
+    let attributes = maplit::hashmap! {
+        format!("PACKET_HOSTNAME") => String::from(hostname),
```
lucab: You can use `xyz.to_string()` on both sides of the arrow (here and below).

src/providers/packet/mock_tests.rs
```diff
@@ -34,3 +34,42 @@ fn test_boot_checkin() {
...
+            "interfaces": [],
+            "addresses": [],
+            "bonding": { "mode": 0 }
+        },
+        "phone_home_url": "test-url"
+    }"#;
+
+    let hostname = "test-hostname";
+    let phone_home_url = "test-url";
+    let plan = "test-plan";
```
lucab: These three are only used within the `hashmap!`, let's directly move them inline there.

src/providers/azure/mock_tests.rs
```diff
@@ -113,3 +113,28 @@ fn test_hostname() {
...
+    let attributes = provider.unwrap().attributes().unwrap();
+    let r = attributes.get("AZURE_VMSIZE");
+
+    m_version.assert();
+    m_goalstate.assert();
+
+    m_vmsize.assert();
+    let vmsize = r.unwrap();
+    assert_eq!(vmsize, testvmsize);
+}
```
lucab, line 149: Let's throw a `mockito::reset()` here at the bottom.

src/providers/packet/mock_tests.rs
zonggen: `try_new()` and `unwrap_err()` might be sufficient since in packet we are fetching the data during `try_new`, and `attributes()` would just read the struct fields.

src/providers/packet/mock_tests.rs
zonggen: :+1: 
Updating!

src/providers/packet/mock_tests.rs
zonggen: Sure

zonggen:
Updated!
EDIT: and rebased on master

src/providers/packet/mock_tests.rs
lucab: Ack (that behavior is quite terrible but not in scope for this PR).

lucab (approved):
LGTM!

PR #283, opened 2019-10-08
providers: add Alibaba Cloud (aliyun)

### LINKED ISSUE #279: providers: add support for Alibaba Cloud
Fedora CoreOS is targeting Alibaba Cloud as a new platform: https://github.com/coreos/fedora-coreos-tracker/issues/276.

Afterburn should also support that, with the platform ID `aliyun`.

This cloud platform has a metadata endpoint, documented at https://www.alibabacloud.com/help/doc-detail/49122.htm.

lucab:
This adds support for Alibaba Cloud, with platform ID `aliyun`.

Closes: https://github.com/coreos/afterburn/issues/279

### DIFF
--- a/src/providers/aliyun/mod.rs
+++ b/src/providers/aliyun/mod.rs
@@ -1,0 +1,187 @@
+//! Alibaba Cloud
+//!
+//! This provider is selected via the platform ID `aliyun`.
+//! The metadata endpoint is documented at https://www.alibabacloud.com/help/doc-detail/49122.htm.
+
+#[cfg(test)]
+use mockito;
+use openssh_keys::PublicKey;
+use std::collections::{BTreeSet, HashMap};
+
+use crate::errors::*;
+use crate::network;
+use crate::providers::MetadataProvider;
+use crate::retry;
+
+#[cfg(test)]
+mod mock_tests;
+
+/// Provider prefix for Alibaba Cloud.
+static PROVIDER_PREFIX: &str = "ALIYUN";
+
+#[derive(Clone, Debug)]
+pub struct AliyunProvider {
+    client: retry::Client,
+}
+
+impl AliyunProvider {
+    pub fn try_new() -> Result<AliyunProvider> {
+        let client = retry::Client::try_new()?.return_on_404(true);
+
+        Ok(AliyunProvider { client })
+    }
+
+    #[cfg(test)]
+    fn endpoint_for(name: &str) -> String {
+        let url = mockito::server_url();
+        format!("{}/{}", url, name)
+    }
+
+    #[cfg(not(test))]
+    fn endpoint_for(name: &str) -> String {
+        format!("http://100.100.100.200/latest/meta-data/{}", name)
+    }
+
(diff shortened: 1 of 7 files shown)


src/providers/aliyun/mod.rs
```diff
@@ -0,0 +1,187 @@
...
+        let entries: Option<String> = self
+            .client
+            .get(retry::Raw, AliyunProvider::endpoint_for("public-keys/"))
+            .send()?;
+        let keys_list = entries.unwrap_or_default();
+
+        let mut out = BTreeSet::new();
+        for entry in keys_list.lines() {
+            let key_id = entry.trim_end_matches('/');
+            let ep = format!("public-keys/{}/openssh-key", key_id);
```
ashcrow, line 83: Just curious: Did you find the single key endpoint via trial and error or was it documented? A quick look at the docs and using search didn't turn it up.

ashcrow (approved):
Makes sense :+1:

src/providers/aliyun/mod.rs
lucab: It is not documented so I had to do a bit of experiments.
However it is basically a 1:1 copy of the AWS logic: https://github.com/coreos/afterburn/blob/59d3c56ccefe093ec98d5aecb2579b88b7b5e9b5/src/providers/aws/mod.rs#L88
For that reason, I only documented the "duplicated entry" quirk.

lucab:
I did some basic manual testing of this on an `aliyun` plain Debian VM, both attributes and ssh-keys logic worked there.

src/providers/aliyun/mod.rs
jlebon: It's also "documented" in the cloud-init implementation :)
https://github.com/cloud-init/cloud-init/commit/4f8ceffb2e3a9feefcb718bda7a7f0f21ef7ab7c

jlebon (approved):
Looks sane to me!

ashcrow:
@lucab is this ready for merge?

zonggen:
LGTM! Pulled locally and tests passed (as expected)!

Another documentation on cloud-init about `AliYun` metadata: https://cloudinit.readthedocs.io/en/latest/topics/datasources/aliyun.html

zonggen (approved):
LGTM! Pulled locally and tests passed (as expected)!

Another documentation on cloud-init about `AliYun` metadata: https://cloudinit.readthedocs.io/en/latest/topics/datasources/aliyun.html

zonggen:
Also I could add basic testing on attribute fetching method, much like this commit https://github.com/coreos/afterburn/pull/278/commits/103f694404290306c111d2b3d053268bb6bad367

zonggen (approved):

lucab:
@zonggen good point, I added a fixup commit (and introduced `maplit`); how does it look now?

ashcrow (approved):
Looks even better now to me :smile:

zonggen:
Nice! Works locally. Will update the testing to my commit too :+1:

PR #295, opened 2019-10-28
*: drop extern crate declarations

bgilbert:
They're not required in Rust 2018.

The last commit is arguably not a net win, so we may want to drop it.


lucab (approved):
LGTM (including the third commit, for consistency)

PR #297, opened 2019-10-29
cloudstack-configdrive: cleanup mounting and umounting logic

lucab:
This improves cloudstack-configdrive logic in order to retry and avoid
panics on errors.


ashcrow (approved):
LGTM. I'd love to see tests for some items but that can be done at a later date.

PR #299, opened 2019-11-01
retry: cleanup and test max-retries

lucab:
This cleans up the max-retries logic, and add some tests to ensure it
works correctly.
In particular, it clarifies that the initial attempt is always performed
and caps the maximum number of retries to `u8::MAX`.

### DIFF
--- a/src/providers/cloudstack/network.rs
+++ b/src/providers/cloudstack/network.rs
@@ -2,7 +2,6 @@
 
 use std::collections::HashMap;
 use std::net::IpAddr;
-use std::time::Duration;
 
 use openssh_keys::PublicKey;
 use slog_scope::warn;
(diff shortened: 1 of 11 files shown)


src/providers/cloudstack/network.rs
```diff
@@ -24,10 +23,7 @@ pub struct CloudstackNetwork {
 impl CloudstackNetwork {
     pub fn try_new() -> Result<CloudstackNetwork> {
         let server_address = CloudstackNetwork::get_dhcp_server_address()?;
-        let client = retry::Client::try_new()?
-            .initial_backoff(Duration::from_secs(1))
-            .max_backoff(Duration::from_secs(5))
-            .max_attempts(10);
```
jlebon, line 30: Are we dropping the backoffs here on purpose?

lucab: These lines were in fact just setting again the default values:
https://github.com/coreos/afterburn/blob/2fc09f15f48908d2b28ff3e3d898e910eea60403/src/retry/mod.rs#L35-L37

jlebon (approved):

PR #302, opened 2019-11-07
providers: add metadata support for IBM Cloud (VPC Gen2)

lucab:
This adds initial metadata support for IBM Cloud (VPC Gen2), under
the provider ID `ibmcloud`.
Right now it only supports extracting attributes and exposing the
hostname.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -21,6 +21,7 @@ use crate::providers::cloudstack::configdrive::ConfigDrive;
 use crate::providers::cloudstack::network::CloudstackNetwork;
 use crate::providers::digitalocean::DigitalOceanProvider;
 use crate::providers::gcp::GcpProvider;
+use crate::providers::ibmcloud;
 use crate::providers::openstack::network::OpenstackProvider;
 use crate::providers::packet::PacketProvider;
 use crate::providers::vagrant_virtualbox::VagrantVirtualboxProvider;
@@ -31,9 +32,11 @@ macro_rules! box_result {
     };
 }
 
-/// `fetch_metadata` is the generic, top-level function that is used by the main
-/// function to fetch metadata. The configured provider is passed in and this
-/// function dispatches the call to the correct provider-specific fetch function
+/// Fetch metadata for the given provider.
+///
+/// This is the generic, top-level function to fetch provider metadata.
+/// The configured provider is passed in and this function dispatches the call
+/// to the provider-specific fetch logic.
 pub fn fetch_metadata(provider: &str) -> errors::Result<Box<dyn providers::MetadataProvider>> {
     match provider {
         "aliyun" => box_result!(AliyunProvider::try_new()?),
@@ -49,6 +52,8 @@ pub fn fetch_metadata(provider: &str) -> errors::Result<Box<dyn providers::Metad
         "gce" => box_result!(GcpProvider::try_new()?),
         #[cfg(not(feature = "cl-legacy"))]
         "gcp" => box_result!(GcpProvider::try_new()?),
+        // ibmcloud has multiple infrastructure types, so this internally tries to auto-detect.
+        "ibmcloud" => ibmcloud::try_autodetect(),
         "openstack-metadata" => box_result!(OpenstackProvider::try_new()?),
         "packet" => box_result!(PacketProvider::try_new()?),
         "vagrant-virtualbox" => box_result!(VagrantVirtualboxProvider::new()),
--- a/src/providers/ibmcloud/gen2.rs
+++ b/src/providers/ibmcloud/gen2.rs
@@ -1,0 +1,199 @@
+//! Metadata fetcher for IBMCloud (VPC Gen2) instances.
+//!
+//! This provider supports the "VPC Generation 2" infrastructure type
+//! on IBMCloud.
+//! It provides a config-drive as the only metadata source, whose layout
+//! is very similar to `cloud-init NoCloud` [datasource][nocloud], with
+//! a few variations:
+//!  - disk label is `cidata` (lowercase)
+//!  - filesystem is `iso9660`
+//!
+//! nocloud: https://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html
+
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::{BufRead, BufReader, Read};
+use std::path::{Path, PathBuf};
+
+use openssh_keys::PublicKey;
+use slog_scope::warn;
+use tempdir::TempDir;
+
+use crate::errors::*;
+use crate::network;
+use crate::providers::MetadataProvider;
+
+const CONFIG_DRIVE_LABEL: &str = "cidata";
+
+/// IBMCloud provider (VPC Gen2).
+#[derive(Debug)]
+pub struct G2Provider {
+    /// Path to the top directory of the mounted config-drive.
+    drive_path: PathBuf,
+    /// Temporary directory for own mountpoint (if any).
+    temp_dir: Option<TempDir>,
+}
+
+impl G2Provider {
+    /// Try to build a new provider client.
+    ///
+    /// This internally tries to mount (and own) the config-drive.
+    pub fn try_new() -> Result<Self> {
+        let target =
+            TempDir::new("afterburn").chain_err(|| "failed to create temporary directory")?;
+        crate::util::mount_ro(
+            &Path::new("/dev/disk/by-label/").join(CONFIG_DRIVE_LABEL),
+            target.path(),
+            "iso9660",
+            3, // maximum retries
+        )?;
+
+        let provider = Self {
+            drive_path: target.path().to_owned(),
+            temp_dir: Some(target),
+        };
+        Ok(provider)
+    }
+
+    /// Return the path to the metadata directory.
+    fn metadata_dir(&self) -> PathBuf {
+        self.drive_path.clone()
+    }
+
+    /// Read metadata file and parse attributes.
+    fn read_metadata(&self) -> Result<HashMap<String, String>> {
+        let filename = self.metadata_dir().join("meta-data");
+        let file =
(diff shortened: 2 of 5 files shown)


lucab:
:heavy_check_mark:  Manually tested `--hostname` and `--attributes` logic (on a vanilla Debian OS).

lucab:
@ashcrow @jlebon may I ask for a review pass on this?

src/providers/ibmcloud/gen2.rs
```diff
@@ -0,0 +1,196 @@
...
+    ///
+    /// This internally tries to mount (and own) the config-drive.
+    pub fn try_new() -> Result<Self> {
+        let target =
+            TempDir::new("afterburn").chain_err(|| "failed to create temporary directory")?;
+        crate::util::mount_ro(
+            &Path::new("/dev/disk/by-label/").join(CONFIG_DRIVE_LABEL),
+            target.path(),
+            "iso9660",
+            3,
```
jlebon: Aside: for magic number params like this, I like to add a comment beside it like: `// retries`. Or alternatively, name the function something more descriptive so it's obvious from context, e.g. `mount_ro_with_retries`.

src/providers/ibmcloud/gen2.rs
```diff
@@ -0,0 +1,196 @@
...
+        )?;
+
+        let provider = Self {
+            drive_path: target.path().to_owned(),
+            temp_dir: Some(target),
+        };
+        Ok(provider)
```
jlebon, line 55: Minor: could squash this with previous statement?

src/providers/ibmcloud/gen2.rs
```diff
@@ -0,0 +1,196 @@
...
+        warn!("boot check-in requested, but not supported on this platform");
+        Ok(())
+    }
+}
+
+impl Drop for G2Provider {
+    fn drop(&mut self) {
+        if self.temp_dir.is_some() {
+            if let Err(e) = crate::util::unmount(&self.metadata_dir(), 3) {
+                error!("failed to cleanup IBM Cloud (Gen2) config-drive: {}", e);
```
jlebon: Minor: `failed to unmount ...` to be clearer?

src/providers/ibmcloud/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+            Err(e) => {
+                slog_scope::debug!("VPC-Gen2 autodetection failed: {}", e);
+            }
+        };
+
+        error_chain::bail!("unable to find ibmcloud metadata");
+    }
+}
+
+impl providers::MetadataProvider for IbmCloudProvider {
```
jlebon: Hmm, do we need to switch and dispatch here? Why not just return the `G2Provider` directly to the caller in the `try_autodetect()`? Or do you expect common code to appear here when other cloud variants are added?

jlebon (commented):
Looks sane to me! Just some minor comments.

src/providers/ibmcloud/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+impl IbmCloudProvider {
+    /// Build a new client for IBMCloud.
+    ///
+    /// This internally tries to autodetect the infrastructure type
+    /// and mount the relevant config-drive.
+    pub fn try_autodetect() -> Result<Self> {
+        // Auto-detection order: Gen2 (possibly later: classic, Gen1).
+        slog_scope::trace!("trying to autodetect ibmcloud infrastructure type");
+
+        match gen2::G2Provider::try_new() {
```
zonggen: Have a question here:
Is it that if we failed to create a `gen2::G2Provider` (failed to create a temp dir or mount the file system), then the program failed to auto-detect the infrastructure type? I'm having this question because the 'detection' step is not obvious to me.
Otherwise LGTM! :+1:

src/providers/ibmcloud/gen2.rs
lucab: I share your feeling.

An alternative would be to have its own simple wrapper type, which means in this line we would pass something like `RetriesNum(3)`.

I don't do that by default as more types introduce more cognitive load, but it looks like here you are lamenting the opposite due to ambiguity. A plain newtype or a [shrinkwrap](https://docs.rs/shrinkwraprs/0.2.1/shrinkwraprs/) are the language way to avoid that.

src/providers/ibmcloud/gen2.rs
lucab: We can, but I don't in general as I prefer an additional line over an additional nesting level. Do you maybe know why your eyes picked this one specifically? How strong is your feeling on this?

src/metadata.rs
```diff
@@ -49,6 +52,7 @@ pub fn fetch_metadata(provider: &str) -> errors::Result<Box<dyn providers::Metad
         "gce" => box_result!(GcpProvider::try_new()?),
         #[cfg(not(feature = "cl-legacy"))]
         "gcp" => box_result!(GcpProvider::try_new()?),
+        "ibmcloud" => box_result!(IbmCloudProvider::try_autodetect()?),
```
ashcrow: It may be worth noting in a comment why this provider uses a different function than the rest.

src/providers/ibmcloud/mod.rs
lucab: The latter. There is already a TODO at the top of this file, but I fear you missed it. Where should I move that to be more visible?

src/providers/ibmcloud/mod.rs
lucab: The detection logic is meant to short-circuit (i.e. return an `Ok(...)`) in the positive case, and otherwise log an error and continue by trying other sub-types in order.
That is, there will be other `match` statements after this one for the other sub-types.

Overall detection only fails when all types have failed, and we reach the `bail!()` statement at the bottom.

Self-note: we'll need to have some scatter-gather parallelism here, otherwise auto-detection will take a linear amount of time with the number of sub-types.

ashcrow (commented):
Only one nit :smile:

src/providers/ibmcloud/gen2.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    /// Parse metadata attributes.
+    ///
+    /// Metadata file contains one attribute per line, in the form of
+    /// `key: value\n`.
+    fn parse_metadata<T: Read>(input: BufReader<T>) -> Result<HashMap<String, String>> {
+        let mut output = HashMap::new();
+
+        for line in input.lines().filter_map(|l| l.ok()) {
+            let parts: Vec<&str> = line.splitn(2, ':').collect();
```
vrutkovs, line 80: `splitn(2, ':')` would always return 2 parts, wouldn't it? In this case the following `parts.len()` is useless.

Perhaps just `split` should be used instead - do we how IBMCloud treats lines like `foo:bar:baz`?

src/providers/ibmcloud/gen2.rs
jlebon: Yeah, I essentially had to dig into the codebase here to find out what that param was. Agreed there's a cognitive load involved in trying to embed a solution into the code. Comments go a long way though. See e.g. https://github.com/coreos/rpm-ostree/blob/2589cd1f927c519cd040485c37881e29cf09f0d2/src/app/rpmostree-builtin-upgrade.c#L178.

src/providers/ibmcloud/gen2.rs
jlebon: > How strong is your feeling on this?

Not very. :)  Agreed it comes down to personal preference.

src/providers/ibmcloud/gen2.rs
lucab: Not always; it returns at most 2 parts, see https://doc.rust-lang.org/std/primitive.str.html#method.splitn.

I don't have any way to test the "embedded :" case, as I can only directly test/control the hostname case and their API reject any name with a `:` inside (with `Use lowercase alphanumeric characters and hyphens only (without spaces)`).

lucab: Not always; it returns at most 2 parts, see https://doc.rust-lang.org/std/primitive.str.html#method.splitn.

I don't have any way to test the "embedded :" case, as I can only directly test/control the hostname case and their API rejects any name with a `:` inside (with `Use lowercase alphanumeric characters and hyphens only (without spaces)`).

vrutkovs: Ah, that's protecting from `foo:` case, got it.

Lets leave it like it is then

src/providers/ibmcloud/mod.rs
jlebon: The top TODO just details that the code is extensible for other cloud types, but doesn't necessarily imply that there would be a lot of shared business logic between implementations. This is fine too though, just trying to avoid boilerplate if we can!

src/providers/ibmcloud/gen2.rs
lucab: @jlebon so your suggestion, across all the alternatives, is to stick to comments in all consuming place, am I reading it right?

jlebon: I think that'd be better than nothing, yeah. Another near-0 cognitive overload approach is renaming it. Don't want to grow the scope here though. We can circle back to this!

src/providers/ibmcloud/mod.rs
lucab: Ah, I see your point only now, sorry.
Yes I think I can change `try_autodetect()` to return a generic boxed `MetadataProvider`.

That would avoid the enum and the matching boilerplate. I don't think we have further logic to share here, as you pointed out above.

I'll try to see if the type-checker is happy with that direction.

lucab:
@jlebon @ashcrow I've added a fixup commit which should address all your comments, PTAL.

jlebon (approved):
Nice, LGTM!

ashcrow (approved):
:+1:

PR #306, opened 2019-11-20
providers/ibmcloud: add support for Classic instance types

lucab:
This adds initial metadata support for IBM Cloud Classic instance
types.
Right now it only supports extracting attributes and exposing the
hostname.


lucab:
:heavy_check_mark:  Manually tested `--hostname` and `--attributes` logic (on a vanilla Debian OS).

jlebon (approved):

PR #309, opened 2019-11-21
network: clean up virtual netdev logic

lucab:
This performs a cleanup of the network logic related to virtual interfaces,
drawing a more clean line between metadata and `systemd.netdev` details.
It also makes the "virtual" detail more explicit to avoid confusion with
physically attached devices.

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -171,8 +171,15 @@ pub trait MetadataProvider {
     fn hostname(&self) -> Result<Option<String>>;
     fn ssh_keys(&self) -> Result<Vec<PublicKey>>;
     fn networks(&self) -> Result<Vec<network::Interface>>;
-    fn network_devices(&self) -> Result<Vec<network::Device>>;
     fn boot_checkin(&self) -> Result<()>;
+
+    /// Return a list of virtual network devices for this machine.
+    ///
+    /// This is used to setup virtual interfaces, e.g. via [systemd.netdev][netdev]
+    /// configuration fragments.
+    ///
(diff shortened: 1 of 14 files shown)


src/providers/mod.rs
```diff
@@ -171,9 +171,16 @@ pub trait MetadataProvider {
     fn hostname(&self) -> Result<Option<String>>;
     fn ssh_keys(&self) -> Result<Vec<PublicKey>>;
     fn networks(&self) -> Result<Vec<network::Interface>>;
-    fn network_devices(&self) -> Result<Vec<network::Device>>;
     fn boot_checkin(&self) -> Result<()>;
 
+    /// Return a list of virtual network devices for this machine.
+    ///
+    /// This is used to setup virtual interfaces, e.g. via [systemd.netdev][netdev]
+    /// configuration fragments.
```
jlebon, line 179: OK, I totally missed in the past that Afterburn also took care of this. Are there plans to support NM too? Should the interface we're setting up here be network-provider independent in that case?

lucab: Yes. This is part of a larger cleanup I'm doing to support NM. That itself is part of larger task to configure RHCOS/FCOS network on ibmcloud classic.

I still don't know how many pieces I can share between NM<->networkd (and how the datastructures would look like), so for the moment I just started by better labeling the things that are clearly networkd-specific.

I guess I'll have a better idea (and make an additional pass on this) once I get closer to synthesizing NM keyfiles.

jlebon (approved):
Looks sane to me!

PR #322, opened 2019-12-02
cli: introduce sub-commands

lucab:
This refactors command-line arguments parsing, introducing sub-commands.
This will in turn make adding new logic and service units less invasive.

Current command-less mode is translated into a new `multi` command,
which can perform multiple actions at once depending on the flags
specified.

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -1,0 +1,198 @@
+//! Command-line arguments parsing.
+
+use crate::errors::*;
+use clap::{crate_version, App, Arg, ArgMatches, SubCommand};
+use slog_scope::trace;
+
+mod multi;
+
+/// Path to kernel command-line (requires procfs mount).
+const CMDLINE_PATH: &str = "/proc/cmdline";
+
+/// CLI sub-commands configuration.
+#[derive(Debug)]
+pub(crate) enum CliConfig {
+    Multi(multi::CliMulti),
+}
+
+impl CliConfig {
+    /// Parse CLI sub-commands into configuration.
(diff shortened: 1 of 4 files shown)


lucab:
I'm queuing this one up, but holding until the release at https://github.com/coreos/afterburn/issues/320 is complete.

src/cli/mod.rs
```diff
@@ -0,0 +1,216 @@
...
+/// In legacy mode there are not sub-commands, and single-dash (Golang-style)
+/// arguments are allowed too.
+fn translate_legacy_args(cli: impl IntoIterator<Item = String>) -> impl Iterator<Item = String> {
+    // Process the first two arguments and check whether there is a sub-command (normal mode)
+    // or not (legacy mode).
+    let mut argv = cli.into_iter();
+    let argv_0 = argv.next().unwrap_or_else(|| "afterburn".to_string());
+    let argv_1 = argv.next();
+    let legacy_mode = match argv_1 {
+        Some(ref arg) => arg.starts_with('-'),
```
jlebon, line 116: Not sure if I follow this logic. The "non-legacy" mode could also fit this, no? E.g. `afterburn --cmdline [subcmd]` ?

lucab: Ah, you are right. I think I'll need to remove the global status from `--cmdline` and `--provider` and push them down as duplicates in each sub-command. Or do you see a better approach to this?

lucab:
@jlebon good catch on my mistake! I've reworked this a bit in order to get rid of global arguments, and left a note for that. PTAL.

jlebon (approved):
LGTM!

PR #338, opened 2020-01-20
[WIP] providers: add experimental initrd network bootstrap

lucab:
This adds initial/experimental support for bootstrapping the network
in the initrd. It is meant to support weird cloud providers where
DHCP is not available or not enough.
This feature is currently reachable as a dedicated `exp rd-net-bootstrap`
subcommand.

PR #339, opened 2020-01-20
cargo: update to reqwest 0.10 and new API 

lucab:
This updates to reqwest 0.10. The default API is now async, but I simply updated and sticked to the blocking API which is good enough for our usecases.

Closes: #337


jlebon (approved):
Seems reasonable!

lucab:
@jlebon FYI I plan to merge this tomorrow and publish a new release, before starting to land further networking logic (e.g. #338).

PR #362, opened 2020-02-20
[WIP] providers: add experimental initrd network bootstrap

lucab:
DRAFT: it is unclear whether we want to provide RHCOS/FCOS on Classic instances, due to overall limitations of Classic infrastructure. Implementation here is complete though, and manually tested.

---

This adds initial/experimental support for bootstrapping the network
in the initrd. It is meant to support weird cloud providers where
DHCP is not available or not usable.
This feature is currently reachable as a dedicated `exp rd-net-bootstrap`
subcommand.
The first provider where this logic is required is `ibmcloud-classic`.

### DIFF
--- a/dracut/30afterburn/afterburn-net-bootstrap.service
+++ b/dracut/30afterburn/afterburn-net-bootstrap.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Afterburn network bootstrapping
+# IBM Cloud (Classic) has no DHCP.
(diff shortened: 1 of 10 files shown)


lucab:
Status update: we aren't shipping RHCOS/FCOS to ibmcloud-classic in the end, so this PR is likely not going anywhere.

There is still the larger topic of bringing up the network from a config-drive. I'm not fully convinced Afterburn should go there, I'm currently feeling that NM may be able to take care of that better.

PR #364, opened 2020-02-22
cargo: switch from deprecated tempdir crate to tempfile

bgilbert:


src/providers/cloudstack/configdrive.rs
```diff
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 
 use openssh_keys::PublicKey;
 use slog_scope::{error, warn};
-use tempdir::TempDir;
+use tempfile::{self, TempDir};
```
lucab: Is the `self` needed here (and other patched files too)?

lucab:
@bgilbert thanks for cleaning up the Option bits too!

src/providers/cloudstack/configdrive.rs
bgilbert: Apparently not.  Fixed.

lucab (approved):

PR #376, opened 2020-03-12
ci: hook up to CoreOS CI

jlebon:
Run a basic FCOS build and kola run.

This doesn't really stress afterburn functionality itself, since CI only
uses qemu right now. Though it's prep for if/when we add testing on
`$cloud`.

(Or I could also imagine some kind of "mock" test harness where we're
still testing on QEMU, but we have some override knob to make afterburn
think it's on `$cloud` and check that it sets up e.g. hostnames or
metadata right.)

For now at least, it still sanity checks that we can compose it fine and
that e.g. the dracut module installs correctly.


cgwalters (approved):
LGTM!

lucab (approved):

PR #379, opened 2020-03-20
[WIP] providers/vmware: add an experimental rd-net-kargs

lucab:
Temporarily based on https://github.com/coreos/afterburn/pull/362 for quicker development.

### DIFF
--- a/src/providers/vmware/mod.rs
+++ b/src/providers/vmware/mod.rs
@@ -1,0 +1,91 @@
+//! VMWare provider.
+//!
+//! This uses the guest->host backdoor protocol for introspection.
+
+use std::collections::HashMap;
+
+use error_chain::bail;
+use openssh_keys::PublicKey;
+use slog_scope::warn;
+
+use crate::errors::*;
+use crate::network;
+use crate::providers::MetadataProvider;
+
+/// Guestinfo key for network kargs.
+static KARGS_NM_INITRD: &str = "guestinfo.exp.org.freedesktop.NetworkManager.initrd";
+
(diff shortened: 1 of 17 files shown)


cgwalters:
In prior discussion we debated whether or not these networking parameters should be "apply once" or be applied every boot.  I thought we'd leaned towards "apply once" - which argues for doing this more as part of Ignition, right?

But, I guess nothing technically stops us from doing "configure once" stuff in afterburn too, it just blurs the architecture diagrams.

src/providers/vmware/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+        fp.write_all(kargs.as_bytes())?;
+        fp.write(&[b'\n'])?;
+        Ok(())
+    }
+
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        Ok(HashMap::new())
+    }
+
+    fn hostname(&self) -> Result<Option<String>> {
```
cgwalters, line 70: We can implement this right?  See https://github.com/openshift/machine-config-operator/pull/1579

lucab: I've replied there. The TLDR is that we can but we shouldn't.
This is for platforms where either:
 * network auto-configuration is performed via DHCP and a cloud-channel for the hostname
 * network auto-configuration happens via some custom protocol, which includes the hostname

Other cases are owned already by 1) dracut kargs and 2) Ignition.

lucab:
Closing this stale WIP in favor of https://github.com/coreos/afterburn/pull/404.

PR #390, opened 2020-04-14
providers: setup network kernel arguments in initrd

lucab:
This adds initial/experimental support for providing supplemental
network kernel arguments in the initrd.
It is meant to support environment where custom network kargs may
be provided via a back-channel.
Such arguments are only retrieved and applied on first boot, and
their format is defined by dracut [0].
The feature is currently reachable as a dedicated subcommand,
but this does not contain any platform-specific implementation.

[0] https://mirrors.edge.kernel.org/pub/linux/utils/boot/dracut/dracut.html#_network

### DIFF
--- a/dracut/30afterburn/afterburn-network-kargs.service
+++ b/dracut/30afterburn/afterburn-network-kargs.service
@@ -1,0 +1,20 @@
+[Unit]
+Description=Afterburn Initrd Setup Network Kernel Arguments
+
+# This service may produce additional kargs fragments,
+# which are then consumed by dracut-cmdline(8).
+DefaultDependencies=no
+Before=dracut-cmdline.service
+Before=ignition-fetch.service
+After=systemd-journald.socket
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Environment=AFTERBURN_OPT_PROVIDER=--cmdline
+# AFTERBURN_NETWORK_KARGS_DEFAULT must be set externally by distributions.
+# If unset, variable expansion results in a missing argument and service
+# hard-failure, on purpose.
+ExecStart=/usr/bin/afterburn exp rd-network-kargs ${AFTERBURN_OPT_PROVIDER} --default-value $AFTERBURN_NETWORK_KARGS_DEFAULT
+Type=oneshot
--- a/src/cli/exp.rs
+++ b/src/cli/exp.rs
@@ -1,0 +1,74 @@
+//! `exp` CLI sub-command.
+
+use crate::errors::*;
+use crate::{initrd, util};
+use clap::ArgMatches;
+use error_chain::bail;
+
+/// Experimental subcommands.
+#[derive(Debug)]
+pub enum CliExp {
+    RdNetworkKargs(CliRdNetworkKargs),
+}
+
+impl CliExp {
+    /// Parse sub-command into configuration.
+    pub(crate) fn parse(app_matches: &ArgMatches) -> Result<super::CliConfig> {
+        if app_matches.subcommand_name().is_none() {
+            bail!("missing subcommand for 'exp'");
+        }
+
+        let cfg = match app_matches.subcommand() {
+            ("rd-network-kargs", Some(matches)) => CliRdNetworkKargs::parse(matches)?,
+            (x, _) => unreachable!("unrecognized subcommand for 'exp': '{}'", x),
+        };
+
+        Ok(super::CliConfig::Exp(cfg))
+    }
+
+    // Run sub-command.
+    pub(crate) fn run(&self) -> Result<()> {
+        match self {
+            CliExp::RdNetworkKargs(cmd) => cmd.run()?,
+        };
+        Ok(())
+    }
+}
+
+/// Sub-command for network kernel arguments.
+#[derive(Debug)]
+pub struct CliRdNetworkKargs {
+    platform: String,
+    default_kargs: String,
+}
+
+impl CliRdNetworkKargs {
+    /// Parse sub-command into configuration.
+    pub(crate) fn parse(matches: &ArgMatches) -> Result<CliExp> {
+        let platform = super::parse_provider(matches)?;
+        let default_kargs = matches
+            .value_of("default-value")
+            .ok_or_else(|| "missing network kargs default value")?
+            .to_string();
+
+        let cfg = Self {
+            platform,
+            default_kargs,
+        };
+        Ok(CliExp::RdNetworkKargs(cfg))
+    }
+
+    /// Run the sub-command.
+    pub(crate) fn run(&self) -> Result<()> {
+        if util::has_network_kargs(super::CMDLINE_PATH)? {
+            slog_scope::warn!("kernel cmdline already specifies network arguments, skipping");
+            return Ok(());
+        };
+
+        let provider_kargs = initrd::fetch_network_kargs(&self.platform)?;
+        let kargs = provider_kargs
+            .as_ref()
+            .unwrap_or_else(|| &self.default_kargs);
+        initrd::write_network_kargs(kargs)
+    }
+}
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -2,8 +2,10 @@
 
 use crate::errors::*;
 use clap::{crate_version, App, Arg, ArgMatches, SubCommand};
+use error_chain::bail;
 use slog_scope::trace;
 
+mod exp;
 mod multi;
 
 /// Path to kernel command-line (requires procfs mount).
@@ -13,6 +15,7 @@ const CMDLINE_PATH: &str = "/proc/cmdline";
 #[derive(Debug)]
 pub(crate) enum CliConfig {
     Multi(multi::CliMulti),
+    Exp(exp::CliExp),
 }
 
 impl CliConfig {
@@ -20,6 +23,7 @@ impl CliConfig {
     pub fn parse_subcommands(app_matches: ArgMatches) -> Result<Self> {
         let cfg = match app_matches.subcommand() {
             ("multi", Some(matches)) => multi::CliMulti::parse(matches)?,
+            ("exp", Some(matches)) => exp::CliExp::parse(matches)?,
             (x, _) => unreachable!("unrecognized subcommand '{}'", x),
         };
 
@@ -30,6 +34,7 @@ impl CliConfig {
     pub fn run(self) -> Result<()> {
         match self {
             CliConfig::Multi(cmd) => cmd.run(),
+            CliConfig::Exp(cmd) => cmd.run(),
         }
     }
 }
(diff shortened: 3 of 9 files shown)


lucab:
This only contains CLI glue, but no actual logic (neither a default nor platform-specific one), in order to first land some scaffolding which I already have in other WIP patches.

/cc @jlebon

lucab:
~~Meh, this needs some kind of catch-all provider otherwise it will fail when 1) there is no provider on kernel cmdline 2) Afterburn does not know the specific provider (e.g. `qemu` in CI here) 3) we introduce a new provider.~~

Edit: due to other early-boot restrictions, I had to go a different path.

lucab:
Temporarily rebased on top of https://github.com/coreos/afterburn/pull/392.

cgwalters:
From the `coreos.auth.verify` test:

```
    3.080761] afterburn[280]: Error: failed to parse command-line arguments
[    3.085988] afterburn[280]: Caused by: error: Found argument '' which wasn't expected, or isn't valid in this context
[    3.092709] afterburn[280]: USAGE:
[    3.095982] afterburn[280]:     afterburn exp rd-network-kargs --cmdline
[    3.099981] afterburn[280]: For more information try --help
[    3.104302] systemd[1]: Startup finished in 2.059s (kernel) + 0 (initrd) + 779ms (userspace) = 2.838s.
```

cgwalters:
Ahh and we need to update ignition in Fedora to pick up https://github.com/coreos/ignition-dracut/pull/146 - that would have made debugging this instant and obvious.

dracut/30afterburn/afterburn-network-kargs.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Afterburn initrd setup network kernel arguments
```
jlebon: Minor: I think the convention is usually Title Case for systemd unit descriptions.

src/cli/mod.rs
```diff
@@ -195,4 +239,22 @@ mod tests {
...
+            "--provider",
+            "gcp",
+            "--default-value",
+            "ip=dhcp",
+        ]
+        .iter()
+        .map(ToString::to_string)
+        .collect();
+
+        parse_args(args).unwrap();
```
jlebon: Shouldn't we check here that the parsed args correspond to the right enum variant and subvariant?

src/cli/exp.rs
```diff
@@ -0,0 +1,69 @@
...
+            platform,
+            default_kargs,
+        };
+        Ok(CliExp::RdNetworkKargs(cfg))
+    }
+
+    /// Run the sub-command.
+    pub(crate) fn run(&self) -> Result<()> {
+        let provider_kargs = initrd::fetch_network_kargs(&self.platform)?;
+        let kargs = provider_kargs.unwrap_or_else(|| self.default_kargs.clone());
```
jlebon: Minor: we're passing a reference anyway to `write_network_kargs`, so we could just do something like `provider_kargs.as_ref().unwrap_or_else(|| self.default_kargs)` here to avoid cloning.

jlebon (commented):
LGTM overall, just a few comments!

lucab:
@jlebon ack, good suggestions. I've rebased this on master, it should contain all the basic logic required for the default-juggling in https://github.com/coreos/fedora-coreos-tracker/issues/460.
I'm going to give this a bunch of manual spins with the reduced grub config.

jlebon (approved):
LGTM!

src/initrd/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+/// This is run early-on in initrd, possibly before networking and other
+/// services are configured, so it may not be able to use all usual metadata
+/// fetcher.
+use crate::errors::*;
+
+use std::fs::File;
+use std::io::Write;
+
+/// Path to cmdline.d fragment for network kernel arguments.
+static KARGS_PATH: &str = "/etc/cmdline.d/50-afterburn-network-kargs";
```
lucab: Self-note: `.conf` suffix was missing here.

lucab:
Spot one silly issue when testing manually with https://github.com/coreos/coreos-assembler/pull/1373. Fixed that, I didn't notice other issues and changing `--default-value=` actually change the behavior on firstboot.

dracut/30afterburn/afterburn-network-kargs.service
```diff
@@ -0,0 +1,18 @@
...
+DefaultDependencies=no
+Before=dracut-cmdline.service
+Before=ignition-fetch.service
+After=systemd-journald.socket
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Environment=AFTERBURN_OPT_DEFAULT_VALUE=--default-value=ip=dhcp,dhcp6
```
cgwalters: The kernel networking defaults are critically important value, and this feels like a rather obscure place for them to live.
How about we add them to fedora-coreos-config in e.g. `/usr/share/coreos-default-network-kargs` or something, and then here do:
`EnvironmentFile=/usr/share/coreos-default-network-kargs`
?

lucab: I agree.
Though, this introduces an hard-requirement on a file which only lives in some -config repo.
Should I keep this current line and just add the `EnvironmentFile=-` afterwards, so that the explicit entry under -config is only *overriding* it if it exists?

cgwalters: But then we have (potentially) two versions; we'd silently use this one if the other one happened to not exist or we typo'd a path.

Perhaps a build-time option is best?  Would help enable other distributions.

lucab: Yes, that's how overriding works by design.
There is also an in-code default [here](https://github.com/coreos/afterburn/pull/390/files#diff-2f2a33f6cbc0b21a19815f89505a9f8cR47) (which can be turned into a build-time option, yes, but I think the unit override is more useful and with lower complexity).

Unless you are actually thinking of making this a required parameter and dropping all the defaults from this repo. In which case, I can rework some bits so that the unit will actually fail if `--default-value` is missing, and inject that **only** via a -config dropin.
Is this UX closer to the flow you had in mind? That is, your preference is for a single definition with hard-fail, not multiple levels of overrides?

cgwalters: > I can rework some bits so that the unit will actually fail if --default-value is missing, and inject that only via a -config dropin.
> Is this UX closer to the flow you had in mind? That is, your preference is for a single definition with hard-fail, not multiple levels of overrides?

I think so yes.  This feels like a global property of the operating system that I wouldn't expect to be hardcoded in afterburn.

Think about how many gyrations we went through with the switch from `dhcp` to `dhcp,dhcp6`, having that be revs to `afterburn.rpm` just feels weird.

So yes, my proposal is that this is a file in some format (text, env foo=bar, JSON, whatever) that we ship in fedora-coreos-config.

(Although, complicating things is that it needs to end up in the initramfs so we'd need a new dracut "module" just to copy the file, but eh.)

This all said...I'm not opposed to landing this as is either to unblock further work and iterating on this as a followup.

lucab: No worries, I'd rather get the integration/design part right upfront. I've pushed a fixup for the flow above. Thanks for the feedback!

lucab:
@cgwalters @jlebon I've pushed a fixup to align the default-flow as suggested in review. It goes together with https://github.com/coreos/fedora-coreos-config/pull/362. Feel free to review either here or there.

dracut/30afterburn/afterburn-network-kargs.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Afterburn Initrd Setup Network Kernel Arguments
+
+# This service may produce additional kargs fragments,
+# which are then consumed by dracut-cmdline(8).
+DefaultDependencies=no
+Before=dracut-cmdline.service
```
dustymabe, line 7: is there anything complicated that needs to happen before this service runs (i.e. device loading or anything to query information (like VMware guestinfo)? If so this might be pretty early.

dustymabe:
is the case where a user provides their own `ip=` arguments handled here? A simple "yes" will suffice.

dustymabe:
> is the case where a user provides their own `ip=` arguments handled here? A simple "yes" will suffice, I just want to make sure.

dracut/30afterburn/afterburn-network-kargs.service
lucab: For VMware specifically, no (it goes via an IOPort, the logic here does not require the vm-tool daemon running).

In general yes, this is very-painfully-early. But this needs to run before the cmdline is assembled by dracut and then fed into the initqueue networking.

I didn't find any better locations given the current status of things, do you have alternative places?

lucab:
@dustymabe if the bootloader command line already contains an `ip=` parameter, the exclusion logic at https://github.com/coreos/afterburn/pull/390/files#diff-2f2a33f6cbc0b21a19815f89505a9f8cR63-R66 short-circuits the rest of the logic.

dracut/30afterburn/afterburn-network-kargs.service
dustymabe: This is fine for now. If we hit a constraint in the future we can do what I do in the copy_firstboot_network code where we [run after cmdline, but before initqueue](https://github.com/coreos/fedora-coreos-config/blob/a8c84a3e055500ce8119535240f17e9e78fb1f4e/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.service#L7-L21) and then just [blank out the results of the earlier call to nm-initrd-generator](https://github.com/coreos/fedora-coreos-config/blob/a8c84a3e055500ce8119535240f17e9e78fb1f4e/overlay.d/05core/usr/lib/dracut/modules.d/15coreos-copy-firstboot-network/coreos-copy-firstboot-network.sh#L52-L54). If we switch to that model we'll need to add some more ordering to the copy_firstboot_network service to run after afterburn.

dracut/30afterburn/afterburn-network-kargs.service
```diff
@@ -0,0 +1,18 @@
...
+Before=ignition-fetch.service
+After=systemd-journald.socket
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Environment=AFTERBURN_OPT_PROVIDER=--cmdline
+# AFTERBURN_NETWORK_KARGS_DEFAULT must be set externally by distributions.
+ExecStart=/usr/bin/afterburn exp rd-network-kargs ${AFTERBURN_OPT_PROVIDER} --default-value $AFTERBURN_NETWORK_KARGS_DEFAULT
```
jlebon, line 19: Shouldn't this be using `${}` too since we want it passed as a single argument?

dustymabe: i.e. to handle the case where the default is `ip=foo ip=bar nameserver=host`?

lucab: Almost, but no. It's both tricky and subtle, so I'll note here all the details. The key part is in https://github.com/systemd/systemd/pull/15450.

`${FOO}` always expands to something (e.g. an empty string), even if the argument is missing. But @cgwalters wants an explicit failure if the default is missing/typoed on the distro side. And an empty string is a totally valid default. So we resort to `$FOO` which expands to zero or more values. Getting one value out of a whitespaced string is achieved by enclosing the value in quotes at the injection point (i.e. the distro dropin).

dustymabe: Looks like a good comment to put in the unit file so someone doesn't try to change it in the future.

dracut/30afterburn/afterburn-network-kargs.service
cgwalters: (Or, rework NM to run as a unit and not in the initqueue)

lucab:
Bump. May I ask for another review round and making sure that everybody is happy with the new approach to default kargs? If so, I think we can start merging https://github.com/coreos/fedora-coreos-config/pull/362 to make CI happy here.

jlebon (approved):

lucab:
The default config landed in both FCOS and RHCOS -config repos. I manually tested this on both with static and DHCP configuration, it works as designed. Merging.

PR #391, opened 2020-04-14
providers: add vmware

lucab:
This adds a skeleton provider for the `vmware` platform. It brings
in library code to access the hypervisor back-channel, but it does
not currently retrieve any specific guestinfo property.


jlebon (approved):
LGTM!

PR #392, opened 2020-04-15
providers: add a default catch-all provider

lucab:
This adds a default provider for otherwise unknown platforms, which
prevents Afterburn from hard-failing but is otherwise intended to
be free of side-effects.


lucab:
This is not actually needed, see https://github.com/coreos/afterburn/pull/390#issuecomment-613707009. Closing.

PR #393, opened 2020-04-15
util/cmdline: add helpers for detecting network kargs

lucab:
This adds some logic for detecting the presence of network kargs
(currently, only `ip=`) on the kernel command-line.


zonggen (commented):
Ran locally without issue and code LGTM!
Though I think the `has_network_kargs` is not used anywhere else..

jlebon (approved):
LGTM!

lucab:
@zonggen indeed, this only contains helper logic. There are a bunch of other PRs in-flight, which are going to consume this.

PR #397, opened 2020-04-20
sshkeys: send structured info to journald

### LINKED ISSUE #395: Add support for sending ssh_authorized_keys info into the journald log
Tagging https://github.com/coreos/fedora-coreos-tracker/issues/279 issue for more visibility here.

In order to solve the above issue, there's one bullet point that talks about displaying information related to ssh_authorized_keys that were added by the `afterburn` process. After having some discussion around this topic(https://github.com/coreos/fedora-coreos-config/pull/344#pullrequestreview-393234733), it appears that sending a journal entry makes more sense here. Ideally, we need to use https://github.com/lucab/libsystemd-rs library to support this feature. 

But it looks like (https://github.com/lucab/libsystemd-rs/pull/34) isn't present in the latest release. It would be really helpful if we cut a new release of that library to support this feature.

sohankunkerkar:
Fixes: https://github.com/coreos/afterburn/issues/395
Until https://github.com/lucab/libsystemd-rs/pull/44 gets merged, I'll be using the [feature-branch](https://github.com/coreos/afterburn/compare/master...sohankunkerkar:send_ssh_keys_info?expand=1#diff-80398c5faae3c069e4e6aa2ed11b28c0R33) to update the `libsystemd` package.
https://github.com/coreos/fedora-coreos-config/pull/344

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ clap = "^2.33"
 error-chain = { version = "^0.12", default-features = false }
 hostname = "^0.3"
 ipnetwork = "^0.16"
+libsystemd = "^0.2.0"
 maplit = "^1.0"
 mime = "^0.3"
 nix = "^0.17"
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -37,12 +37,13 @@ pub mod packet;
 pub mod vagrant_virtualbox;
 pub mod vmware;
 
+use libsystemd::logging;
+use openssh_keys::PublicKey;
+use slog_scope::warn;
 use std::collections::HashMap;
 use std::fs::{self, File};
 use std::io::prelude::*;
 use std::path::Path;
-
-use openssh_keys::PublicKey;
 use users::{self, User};
 
 use crate::errors::*;
@@ -52,6 +53,8 @@ use crate::network;
 const ENV_PREFIX: &str = "AFTERBURN_";
 #[cfg(feature = "cl-legacy")]
 const ENV_PREFIX: &str = "COREOS_";
+/// Message ID marker for authorized-keys entries in journal.
+const AFTERBURN_SSH_AUTHORIZED_KEYS_MESSAGEID: &str = "0f7d7a502f2d433caa1323440a6b4190";
 
 fn create_file(filename: &str) -> Result<File> {
     let file_path = Path::new(&filename);
@@ -64,6 +67,20 @@ fn create_file(filename: &str) -> Result<File> {
     File::create(file_path).chain_err(|| format!("failed to create file {:?}", file_path))
 }
 
+/// Add a message to the journal logging SSH key additions; this
+/// will be used by at least Fedora CoreOS to display in the console
+/// if no ssh keys are present.
+fn write_ssh_key_journal_entry(log: logging::Priority, name: &str, path: &str) {
+    let message = format!("wrote ssh authorized keys file for user: {}", name);
+    let map = maplit::hashmap! {
+        "AFTERBURN_USER_NAME" => name.as_ref(),
+        "AFTERBURN_PATH" => path.as_ref(),
+        "MESSAGE_ID" => AFTERBURN_SSH_AUTHORIZED_KEYS_MESSAGEID,
+    };
+    if let Err(e) = logging::journal_send(log, &message, map.iter()) {
+        warn!("failed to send information to journald: {}", e);
+    }
+}
 #[cfg(feature = "cl-legacy")]
 fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
     use update_ssh_keys::{AuthorizedKeyEntry, AuthorizedKeys};
@@ -95,6 +112,8 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
         authorized_keys_dir
             .sync()
             .chain_err(|| "failed to update authorized keys")?;
+        let path = authorized_keys_dir.ssh_dir.display().to_string();
+        write_ssh_key_journal_entry(logging::Priority::Info, &user_name, &path);
     }
 
     Ok(())
@@ -155,6 +174,10 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
         }
         .chain_err(|| format!("failed to remove file {:?}", file_path.display()))?;
     }
+
+    let username = user.name().to_string_lossy();
+    let path = file_path.to_string_lossy();
+    write_ssh_key_journal_entry(logging::Priority::Info, &username, &path);
 
     // sync parent dir to persist updates
     match File::open(&dir_path) {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,6 +10,7 @@ dependencies = [
  "error-chain",
  "hostname",
  "ipnetwork 0.16.0",
+ "libsystemd",
  "maplit",
  "mime",
  "mockito",
@@ -53,9 +54,9 @@ dependencies = [
 
 [[package]]
 name = "arc-swap"
-version = "0.4.5"
+version = "0.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d663a8e9a99154b5fb793032533f6328da35e23aac63d5c152279aa8ba356825"
+checksum = "b585a98a234c46fc563103e9278c9391fde1f4e6850334da895d27edb9580f62"
 
 [[package]]
 name = "arrayref"
@@ -182,9 +183,9 @@ checksum = "130aac562c0dd69c56b3b1cc8ffd2e17be31d0b6c25b61c96b76231aa23e39e1"
 
 [[package]]
 name = "cc"
-version = "1.0.50"
+version = "1.0.51"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "95e28fa049fda1c330bcf9d723be7663a899c4679724b34c81e9f5a326aab8cd"
+checksum = "9c9384ca4b90c0ea47e19a5c996d6643a3e73dedf9b89c65efb67587e34da1bb"
 
 [[package]]
 name = "cfg-if"
@@ -270,6 +271,16 @@ dependencies = [
  "autocfg",
  "cfg-if",
  "lazy_static",
+]
+
+[[package]]
+name = "crypto-mac"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4434400df11d95d556bac068ddfedd482915eb18fe8bea89bc80b6e4b1c179e5"
+dependencies = [
+ "generic-array",
+ "subtle",
 ]
 
 [[package]]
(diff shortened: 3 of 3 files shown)


src/providers/mod.rs
```diff
@@ -51,6 +52,7 @@ use crate::network;
 const ENV_PREFIX: &str = "AFTERBURN_";
 #[cfg(feature = "cl-legacy")]
 const ENV_PREFIX: &str = "COREOS_";
+const AFTERBURN_SSH_ATHORIZED_KEYS_MESSAGEID: &str = "0f7d7a502f2d433caa1323440a6b4190";
```
lucab: Typo: `AUTHORIZED`.

Also, please add a docstring to record what's the purpose of this.

src/providers/mod.rs
```diff
@@ -130,7 +132,18 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
                 format!("failed to write to file {:?}", temp_file.path().display())
             })?;
         }
-
+        let name = user.name().to_string_lossy();
+        let path = file_path.to_string_lossy();
+        let mut map: HashMap<&str, &str> = HashMap::new();
+        let messge = format!(
+            "wrote ssh authorized keys file for user:{user}",
```
lucab: You can just use `{}` here and remove the `user` assignment in the line below.

src/providers/mod.rs
```diff
@@ -130,7 +132,18 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
                 format!("failed to write to file {:?}", temp_file.path().display())
             })?;
         }
-
+        let name = user.name().to_string_lossy();
+        let path = file_path.to_string_lossy();
+        let mut map: HashMap<&str, &str> = HashMap::new();
```
lucab: See https://docs.rs/maplit/1.0.2/maplit/macro.hashmap.html for a cleaner way of doing this.

src/providers/mod.rs
```diff
@@ -130,7 +132,18 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
...
+        let messge = format!(
+            "wrote ssh authorized keys file for user:{user}",
+            user = &name
+        );
+        map.insert("AFTERBURN_USER_NAME", &name);
+        map.insert("AFTERBURN_PATH", &path);
+        map.insert("MESSAGE_ID", AFTERBURN_SSH_ATHORIZED_KEYS_MESSAGEID);
+        let res = logging::journal_send(logging::Priority::Info, &messge, map.iter());
+        res.chain_err(|| format!("failed to send journald entry"))?;
```
lucab: You aren't further using `res`, so you can merge this line with the one above and drop the variable assignment altogether.

lucab (changes requested):

cgwalters:
Ah, this project uses conditional compilation; see the
`#[cfg(not(feature = "cl-legacy"))]`.  So you're getting "unused code" errors when building with `cl-legacy`.  Basically you need to move the imports inside the function; see what is there now like:

```
#[cfg(not(feature = "cl-legacy"))]
fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
    use std::io::ErrorKind::NotFound;
    use users::os::unix::UserExt;
```

src/providers/mod.rs
```diff
@@ -52,6 +52,10 @@ use crate::network;
 const ENV_PREFIX: &str = "AFTERBURN_";
 #[cfg(feature = "cl-legacy")]
 const ENV_PREFIX: &str = "COREOS_";
+// AFTERBURN_SSH_AUTHORIZED_KEYS_MESSAGEID keeps track of
```
lucab: Docstrings are marked with `///`; to need to repeat the constant name:
`/// Message ID marker for authorized-keys entries in journal.`

src/providers/mod.rs
```diff
@@ -52,6 +52,10 @@ use crate::network;
...
+// the journald log related to ssh_authorized_keys information.
+#[cfg(not(feature = "cl-legacy"))]
```
lucab: I think we can add this logic in all cases, i.e. with and without the legacy feature enabled.

src/providers/mod.rs
```diff
@@ -131,7 +136,16 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
                 format!("failed to write to file {:?}", temp_file.path().display())
             })?;
         }
-
+        let name = user.name().to_string_lossy();
+        let path = file_path.to_string_lossy();
+        let messge = format!("wrote ssh authorized keys file for user:{}", name.as_ref());
```
lucab: Typo: `message`.

src/providers/mod.rs
```diff
@@ -131,7 +136,16 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
...
+        let map = maplit::hashmap! {
+            "AFTERBURN_USER_NAME" => name.as_ref(),
+            "AFTERBURN_PATH" => path.as_ref(),
+            "MESSAGE_ID" => AFTERBURN_SSH_AUTHORIZED_KEYS_MESSAGEID,
+        };
+        logging::journal_send(logging::Priority::Info, &messge, map.iter())
+            .chain_err(|| format!("failed to send journald entry:{}", path.as_ref()))?;
```
lucab: I don't think we want to fail if we can't write to the journal, just logging a warning is enough.

Also, this logic should be moved further below, after we have written and renamed the file.

You can even move this to its own function, and re-use it in the two (legacy and non-legacy) cases.

src/providers/mod.rs
```diff
@@ -64,6 +67,23 @@ fn create_file(filename: &str) -> Result<File> {
...
+    name: &str,
+    path: &str,
+) -> Result<()> {
+    let map = maplit::hashmap! {
+        "AFTERBURN_USER_NAME" => name.as_ref(),
+        "AFTERBURN_PATH" => path.as_ref(),
+        "MESSAGE_ID" => AFTERBURN_SSH_AUTHORIZED_KEYS_MESSAGEID,
+    };
+    let result = logging::journal_send(log, msg, map.iter());
+    if let Err(e) = result {
```
jlebon:
```suggestion
    if let Err(result) = logging::journal_send(log, msg, map.iter()) {
```

src/providers/mod.rs
```diff
@@ -156,6 +175,14 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
         .chain_err(|| format!("failed to remove file {:?}", file_path.display()))?;
     }
 
+    let name = user.name().to_string_lossy();
+    let path = file_path.to_string_lossy();
+    let message = format!("wrote ssh authorized keys file for user:{}", name.as_ref());
```
jlebon:
```suggestion
    let message = format!("wrote ssh authorized keys file for user: {}", name.as_ref());
```

Cargo.toml
```diff
@@ -30,6 +30,7 @@ clap = "^2.33"
 error-chain = { version = "^0.12", default-features = false }
 hostname = "^0.3"
 ipnetwork = "^0.16"
+libsystemd = { git = "https://github.com/lucab/libsystemd-rs", branch = "master"}
```
jlebon: I think we can use `0.2.0` now?

src/providers/mod.rs
```diff
@@ -131,7 +151,6 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
                 format!("failed to write to file {:?}", temp_file.path().display())
             })?;
         }
-
```
jlebon: Minor: was this diff hunk on purpose?

src/providers/mod.rs
```diff
@@ -156,6 +175,14 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
         .chain_err(|| format!("failed to remove file {:?}", file_path.display()))?;
     }
 
+    let name = user.name().to_string_lossy();
+    let path = file_path.to_string_lossy();
+    let message = format!("wrote ssh authorized keys file for user:{}", name.as_ref());
+    let sent = write_structured_journald_entry(logging::Priority::Info, &message, &name, &path);
+    if let Err(e) = sent {
+        warn!("failed to write structured journald entry: {}", e);
+    }
```
jlebon: We already warn in `write_structured_journald_entry`, right? So we could just call it and use `?` here instead. (I guess we could change that function to be infallible now, though meh... nice to just leave it like that too in case we do add errors in the future.)

lucab: Let's make it infallible, we don't plan to fail the pubkeys flow because of journald.

src/providers/mod.rs
```diff
@@ -64,6 +67,23 @@ fn create_file(filename: &str) -> Result<File> {
...
+    path: &str,
+) -> Result<()> {
+    let map = maplit::hashmap! {
+        "AFTERBURN_USER_NAME" => name.as_ref(),
+        "AFTERBURN_PATH" => path.as_ref(),
+        "MESSAGE_ID" => AFTERBURN_SSH_AUTHORIZED_KEYS_MESSAGEID,
+    };
+
+    if let Err(e) = logging::journal_send(log, msg, map.iter()) {
+        warn!("failed to send information to journald: {}", e);
```
cgwalters, line 81: Of course, with this message we'll also try to send it to the journal...something will have had to go catastrophically wrong for it to fail.

src/providers/mod.rs
```diff
@@ -64,6 +67,23 @@ fn create_file(filename: &str) -> Result<File> {
     File::create(file_path).chain_err(|| format!("failed to create file {:?}", file_path))
 }
 
+fn write_structured_journald_entry(
```
cgwalters:
```suggestion
/// Add a message to the journal logging SSH key additions; this
/// will be used by at least Fedora CoreOS to display in the console
/// if no ssh keys are present.
fn write_ssh_key_journal_entry(
```

cgwalters: I think this is worth a comment.

Also, right now it's a bit odd because it takes the message as a parameter, but hardcodes the message ID.  Either we're making a generic function, or we're doing something just for SSH.  I think the latter makes more sense, so that's the rationale for renaming the method.

src/providers/mod.rs
cgwalters: Hm, I would have said let's only touch the FCOS path so we don't need to think about handling both, but eh.

src/providers/mod.rs
```diff
@@ -73,13 +91,13 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
     if !ssh_keys.is_empty() {
         // open the user's authorized keys directory
         let user_name = user.name().to_string_lossy().into_owned();
+        let path = AuthorizedKeys.ssh_dir.display();
```
jlebon: This should be `let path = authorized_keys_dir.ssh_dir.display();`. So then you'd put it lower down. Could just be right before the call to `write_ssh_key_journal_entry`.

src/providers/mod.rs
```diff
@@ -64,6 +67,21 @@ fn create_file(filename: &str) -> Result<File> {
     File::create(file_path).chain_err(|| format!("failed to create file {:?}", file_path))
 }
 
+/// Add a message to the journal logging SSH key additions; this
+/// will be used by at least Fedora CoreOS to display in the console
+/// if no ssh keys are present.
+fn write_ssh_key_journal_entry(log: logging::Priority, name: &str, path: &str) -> Result<()> {
```
lucab: We don't want this function to fail the ssh-keys flows, thus we are not interested in this returning a result.

You can make this infallible, dropping the `Result` here and the `Ok` at the bottom.

Cargo.toml
```diff
@@ -30,6 +30,7 @@ clap = "^2.33"
 error-chain = { version = "^0.12", default-features = false }
 hostname = "^0.3"
 ipnetwork = "^0.16"
+libsystemd = { version = "^0.2.0"}
```
jlebon: Minor: this could just be `libsystemd = "^0.2.0"`, right?

src/providers/mod.rs
```diff
@@ -79,7 +96,6 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
                 user_name
             )
         })?;
-
```
jlebon: Minor: stray change

jlebon (approved):
Some minor nits, but LGTM overall!

lucab (approved):

PR #398, opened 2020-04-21
bump rust version from 1.39.0 to 1.40.0 

sohankunkerkar:
Current CI is unable to handle `#[cfg(doctest)]` tag in the code and throwing the following error:
```
error[E0658]: `cfg(doctest)` is experimental and subject to change
```
For more information: (https://travis-ci.org/github/coreos/afterburn/jobs/677790013#L412)

[This](https://github.com/rust-lang/rust/pull/63803) PR fixes the above issue which is present in the rust  `v1.40.0` release.

PR #401, opened 2020-04-23
doc: update minimum toolchain version

lucab:
This updates the Minimum Supported Rust Version (MSRV) in README,
matching CI setup.


lucab:
Followup to https://github.com/coreos/afterburn/pull/398.

/cc @sohankunkerkar @cgwalters

jlebon:
From a quick look at the el8 side, I think we can take it all the way to 1.42. No real rush though.

PR #404, opened 2020-04-28
providers/vmware: support injecting custom network kargs

lucab:
This implements fetching logic specific to the `vmware` provider
in order to support injecting custom network kargs, which are meant
to override the default ones.

In particular, this checks for an optional guestinfo property named
`guestinfo.afterburn.initrd.network-kargs` which can be used to provide
a string containing VM-specific overrides.

### DIFF
--- a/src/initrd/mod.rs
+++ b/src/initrd/mod.rs
@@ -1,10 +1,12 @@
-/// Agent logic running at early boot.
-///
-/// This is run early-on in initrd, possibly before networking and other
-/// services are configured, so it may not be able to use all usual metadata
-/// fetcher.
-use crate::errors::*;
+//! Agent logic running at early boot.
+//!
+//! This is run early-on in initrd, possibly before networking and other
+//! services are configured, so it may not be able to use all usual metadata
+//! fetcher.
 
+use crate::errors::*;
+use crate::providers::vmware::VmwareProvider;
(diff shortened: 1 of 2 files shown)


jcpowermac:
@lucab I was following the closed PR.
I would assume that vmtoolsd would need to be running before trying to query for guestinfo?
Could we use this to set the hostname?


cc: 
@mtnbikenc
@patrickdillon

lucab:
> I would assume that vmtoolsd would need to be running before trying to query for guestinfo?

No, this has no dependency on `vmtoolsd`. Like Ignition, it queries the hypervisor back-channel directly. At the time when this (and Ignition) runs in initramfs, it is quite too early for vmtoolsd.

> Could we use this to set the hostname?

Possibly, if you already have it as part of your `ip=` and if that works generally.
However, I'd recommend **against** approaching it that way as it is fragile (as we keep finding out, see e.g. https://github.com/coreos/fedora-coreos-tracker/issues/466)
Hostname should be coming from DHCP. If that is missing or needs to be overridden, static hostname configuration should be performed via Ignition configuration (like all other OS customizations).

lucab:
Rebased against master, this is ready for review. CI is green and I manually tested on a vSphere setup with static networking.

/cc @jlebon

cgwalters (approved):

PR #414, opened 2020-05-17
Remove cl-legacy feature

bgilbert:
CoreOS Container Linux will be EOL soon, and Flatcar Container Linux ships a fork of this repo.

cc @pothos, @iaguis, @dongsupark for comments.


pothos:
Hi,
thanks for asking. If you are willing to keep the legacy feature it made it easier to follow upstream. We are not on your latest `master` currently but that is hopefully to change soon. We have a few minor changes regarding the OEM ID which we could even upstream.
But we have no timeline yet when to switch to the new `AFTERBURN_` naming scheme because it is a breaking change; so if you want the `legacy` feature to be gone soon, you could remove it now as well instead of delaying this decision. I think it's better to share Afterburn, not only because I don't like to resolve git conflicts but because it enables us to contribute upstream.

bgilbert:
I'm not opposed to leaving the existing legacy code in place if it's useful.  In that case it would make sense to upstream the Flatcar-specific changes, since hopefully no other distro will be using legacy mode.  Regarding migrating to `AFTERBURN_`, Afterburn could write both the old and new attributes to different files to avoid the need for a flag day.

On the other hand, we'd like to add support for configuring networking via NetworkManager on cloud platforms that require that (e.g. DigitalOcean and Packet).  It'd be a lot easier to do that if we don't need to abstract over both NetworkManager and networkd.

@lucab, any thoughts?

lucab:
Yes, I think we should come up with a plan to drop the feature soon-ish, as that would simplify code & CI.

https://github.com/coreos/afterburn/pull/322 introduced CLI sub-commands/verbs, so assuming that service units can be auto-updated together with the binary, we can consider isolating the legacy logic under its own `afterburn legacy` scope while pulling out bits that are still common (e.g. hostname handling).

ashcrow:
Needs a rebase.

bgilbert:
Rebased.  We're now 9 months past CL EOL, so it seems like a good time to revisit this.

It appears that [Flatcar's fork](https://github.com/kinvolk/afterburn) isn't tracking upstream, and I'm not aware of any additional upstreaming of compat code.  At this point I'd propose dropping the `cl-legacy` feature to ease maintenance.  We can always revisit the compatibility issue with downstream's participation, though I'd argue that a migration plan should be part of that process.

lucab:
Agreed, I'm also looking forward to drop these parts at this point. I think we are going to end up with a cleaner table to accept any future compat logic, if that materializes at some point.

lucab:
Agreed, I'm also looking forward to drop these parts at this point. I think we are going to end up with a cleaner table to accept any future compat logic, if that materializes at some point.

bgilbert:
(deleted duplicate comment)

jlebon (approved):
Agreed with all the above. Patch itself LGTM!

jlebon (approved):
Agreed with all the above. Patch itself LGTM!

travier (approved):

PR #421, opened 2020-05-29
aws: change metadata version from 2009-04-04 to 2019-10-01

### LINKED ISSUE #418: Update AWS metadata version
# Feature Request #

## Environment ##

AWS

## Desired Feature ##
https://github.com/coreos/afterburn/blob/d8f74d5b6fd95dc201a406a68f33eb74b267dcbb/src/providers/aws/mod.rs#L65 references a fairly old AWS metadata version. To avoid confusion during support cases, consider using 2016-09-02 which is the most recent version specified at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html.

zonggen:
This changes updates the instance metadata version for AWS instances,
as mentioned in https://github.com/coreos/afterburn/issues/418.

The original proposed version is 2016-09-02, same as the latest version on official documentation: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html. However, after running `curl http://169.254.169.254` on an FCOS instance on AWS:

```
[core@ip-172-31-48-79 ~]$ curl http://169.254.169.254
1.0
2007-01-19
2007-03-01
2007-08-29
2007-10-10
2007-12-15
2008-02-01
2008-09-01
2009-04-04
2011-01-01
2011-05-01
2012-01-12
2014-02-25
2014-11-05
2015-10-20
2016-04-19
2016-06-30
2016-09-02
2018-03-28
2018-08-17
2018-09-24
2019-10-01
latest
```

The current latest version is 2019-10-01, and I've verified this version currently works with Afterburn. So this PR will update the version to 2019-10-01 instead. Also if we want to track the latest version we could change to `const URL: &str = "http://169.254.169.254/latest";`, but that has the possibility to break Afterburn if the latest version changes the API endpoints. Mock tests are not updated because the API items remain the same as 2009-04-04.

Closes: https://github.com/coreos/afterburn/issues/418

### DIFF
--- a/src/providers/aws/mod.rs
+++ b/src/providers/aws/mod.rs
@@ -62,7 +62,7 @@ impl AwsProvider {
 
     #[cfg(not(test))]
     fn endpoint_for(key: &str) -> String {
-        const URL: &str = "http://169.254.169.254/2009-04-04";
+        const URL: &str = "http://169.254.169.254/2019-10-01";
         format!("{}/{}", URL, key)
     }
 


bgilbert:
Have you confirmed from documentation that there are no behavior changes that will affect us?

zonggen:
I have done the following:
 - Checked IMDSv1 metadata version list on documentation
 - Checked IMDSv1 API items have not changed since 2009-04-04
 - Ran `curl http://169.254.169.254` to confirm version list on FCOS instance on AWS
 - Ran `curl http://169.254.169.254/2019-10-01` to confirm API items have not changed
 - Ran `./afterburn --provider aws --attributes output` with this PR added

So far there's no compatibility issues and the local build has run without issue on AWS. Is there anything else on documentation I should check specifically?

cgwalters:
One concern I have is whether this new API is available in the "other" AWS regions like GovCloud and the other ones.

zonggen:
The only thing I've found related to GovCloud instance metadata is https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/govcloud-ec2.html. It has some restrictions on "ITAR-Regulated Data" but I _think_ other functionalities remain the same as normal EC2 instances?

lucab:
This seems sane to me to merge (@zonggen and thanks for recording here the additional manual testing you did).
So far I have never heard of discrepancies across AWS partitions regarding the metadata endpoint (but I'm always open to be surprised on cloud API :).

@cgwalters @bgilbert unless you have standing concerns, I think this can be merged as-is.

cgwalters:
Not blocking this PR, but one important thing to note is that some Kubernetes components including kubelet (I believe) and the OpenShift router talk directly to the metadata service too.

So we can't turn on the "only IMDSv2" flag on instances unless they get migrated too (or we do something like have afterburn dump the metadata into e.g. `/run/aws-metadata.json` (owned by `0600`) so they don't need to access the network.

lucab:
@cgwalters I think your comment above was meant for https://github.com/coreos/afterburn/pull/426, right? There is already some discussion there (and at https://github.com/coreos/ignition/pull/980 too) about compatibility matrix.

bgilbert (approved):

lucab (approved):

PR #426, opened 2020-06-02
aws: support IMDSv2 for AWS metadata service

### LINKED ISSUE #305: aws: add support for metadata endpoint v2 (IMDSv2)
AWS recently announced a new generation of their metadata endpoint, called IMDSv2.

It is a session-based service, meant to plug a few holes in the v1 design, most notably [SSRF](https://en.wikipedia.org/wiki/Server-side_request_forgery).
See the announcement at https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/.

For the moment both v1 and v2 are active by default, and instances can be configured to disable the v1 endpoint.

In the medium term, Afterburn should move to v2 logic.

zonggen:
Adds the logic for supporting AWS IMDSv2. By default, Afterburn will use v2 logic by making a PUT request to the endpoint to fetch the session token. If anything happens and the request failed, then Afterburn switches to v1 logic.

Note that the only difference between v2 and v1 logic is that v2 makes use of the session token and set the token in the subsequent GET request headers, whereas v1 logic sends GET request without additional information in the header.

Also adds tests for the IMDS version testing. Test cases are:
 1) successful IMDSv2 and 
 2) failed IMDSv2 so use IMDSv1 instead.

Closes: #305

### DIFF
--- a/src/providers/aws/mock_tests.rs
+++ b/src/providers/aws/mock_tests.rs
@@ -86,3 +86,98 @@ fn test_aws_attributes() {
     mockito::reset();
     provider.attributes().unwrap_err();
 }
+
+#[test]
+fn test_aws_imds_versions() {
+    let instance_id = "test-instance-id";
+    let instance_type = "test-instance-type";
+    let ipv4_local = "test-ipv4-local";
+    let ipv4_public = "test-ipv4-public";
+    let availability_zone = "test-availability-zone";
+    let hostname = "test-hostname";
+    let public_hostname = "test-public-hostname";
+    let instance_id_doc = r#"{"region": "test-region"}"#;
+    let region = "test-region";
+
+    let attributes = maplit::hashmap! {
+        format!("{}_INSTANCE_ID", aws::ENV_PREFIX) => instance_id.to_string(),
+        format!("{}_INSTANCE_TYPE", aws::ENV_PREFIX) => instance_type.to_string(),
+        format!("{}_IPV4_LOCAL", aws::ENV_PREFIX) => ipv4_local.to_string(),
+        format!("{}_IPV4_PUBLIC", aws::ENV_PREFIX) => ipv4_public.to_string(),
+        format!("{}_AVAILABILITY_ZONE", aws::ENV_PREFIX) => availability_zone.to_string(),
+        format!("{}_HOSTNAME", aws::ENV_PREFIX) => hostname.to_string(),
+        format!("{}_PUBLIC_HOSTNAME", aws::ENV_PREFIX) => public_hostname.to_string(),
+        format!("{}_REGION", aws::ENV_PREFIX) => region.to_string(),
+    };
+
+    let endpoints = maplit::btreemap! {
+        "/meta-data/instance-id" => instance_id,
+        "/meta-data/instance-type" => instance_type,
+        "/meta-data/local-ipv4" => ipv4_local,
+        "/meta-data/public-ipv4" => ipv4_public,
+        "/meta-data/placement/availability-zone" => availability_zone,
+        "/meta-data/hostname" => hostname,
+        "/meta-data/public-hostname" => public_hostname,
+        "/dynamic/instance-identity/document" => instance_id_doc,
+    };
+
+    let client = crate::retry::Client::try_new()
+        .chain_err(|| "failed to create http client")
+        .unwrap()
+        .max_retries(0)
+        .return_on_404(true);
+
+    // first test imdsv1
+    {
+        let mut mocks = Vec::with_capacity(endpoints.len());
+        for (endpoint, body) in endpoints.clone() {
+            let m = mockito::mock("GET", endpoint)
+                .with_status(200)
+                .with_body(body)
+                .create();
+            mocks.push(m);
+        }
+
+        let _m = mockito::mock("PUT", "/api/token")
+            .match_header("X-aws-ec2-metadata-token-ttl-seconds", "21600")
+            .with_status(403)
+            .with_body("Forbidden")
+            .create();
+
+        let provider = aws::AwsProvider::with_client(client.clone()).unwrap();
+
+        let v = provider.attributes().unwrap();
+        assert_eq!(v, attributes);
+
+        mockito::reset();
+        provider.attributes().unwrap_err();
+    }
+
+    {
+        // then test imdsv2
+        let token = "test-api-token";
+        let mut mocks = Vec::with_capacity(endpoints.len());
+        for (endpoint, body) in endpoints.clone() {
+            let m = mockito::mock("GET", endpoint)
+                .match_header("X-aws-ec2-metadata-token", token)
+                .with_status(200)
+                .with_body(body)
+                .create();
+            mocks.push(m);
+        }
+
+        let _m = mockito::mock("PUT", "/api/token")
+            .match_header("X-aws-ec2-metadata-token-ttl-seconds", "21600")
+            .with_status(200)
+            .with_body(token)
+            .create();
+
+        let provider = aws::AwsProvider::with_client(client.clone()).unwrap();
+
+        let v = provider.attributes().unwrap();
+        assert_eq!(v, attributes);
+
+        mockito::reset();
+        provider.attributes().unwrap_err();
+    }
+}
--- a/src/providers/aws/mod.rs
+++ b/src/providers/aws/mod.rs
@@ -20,6 +20,7 @@ use std::collections::HashMap;
 #[cfg(test)]
 use mockito;
 use openssh_keys::PublicKey;
+use reqwest::header;
 use serde_derive::Deserialize;
 use slog_scope::warn;
 
@@ -50,20 +51,66 @@ pub struct AwsProvider {
 impl AwsProvider {
     pub fn try_new() -> Result<AwsProvider> {
         let client = retry::Client::try_new()?.return_on_404(true);
+        AwsProvider::with_client(client)
+    }
+
+    fn with_client(client: retry::Client) -> Result<AwsProvider> {
+        let mut client = client;
+        let token = AwsProvider::fetch_imdsv2_token(client.clone());
+
+        // If IMDSv2 token is fetched successfully, set the header.
+        // Otherwise, proceed with IMDSv1 mechanism.
+        match token {
+            Ok(t) => {
+                client = client.header(
+                    header::HeaderName::from_bytes(b"X-aws-ec2-metadata-token")
+                        .chain_err(|| "setting header name for aws imdsv2 metadata")?,
+                    header::HeaderValue::from_bytes(t.as_bytes())
+                        .chain_err(|| "setting header value for aws imdsv2 metadata")?,
+                );
+            }
+            Err(err) => {
+                warn!("failed to fetch aws imdsv2 session token with: {}", err);
+            }
+        }
 
         Ok(AwsProvider { client })
     }
 
     #[cfg(test)]
-    fn endpoint_for(key: &str) -> String {
+    fn endpoint_for(key: &str, _use_latest: bool) -> String {
         let url = mockito::server_url();
         format!("{}/{}", url, key)
     }
 
     #[cfg(not(test))]
-    fn endpoint_for(key: &str) -> String {
+    fn endpoint_for(key: &str, use_latest: bool) -> String {
         const URL: &str = "http://169.254.169.254/2019-10-01";
-        format!("{}/{}", URL, key)
+        const URL_LATEST: &str = "http://169.254.169.254/latest";
+        if use_latest {
+            format!("{}/{}", URL_LATEST, key)
+        } else {
+            format!("{}/{}", URL, key)
+        }
+    }
+
+    fn fetch_imdsv2_token(client: retry::Client) -> Result<String> {
+        let token: String = client
+            .header(
+                header::HeaderName::from_bytes(b"X-aws-ec2-metadata-token-ttl-seconds")
+                    .chain_err(|| "setting header name for aws imdsv2 token")?,
+                header::HeaderValue::from_bytes(b"21600")
+                    .chain_err(|| "setting header value for aws imdsv2 token")?,
+            )
+            .put(
+                retry::Raw,
+                // NOTE(zonggen): Use `latest` here since other versions would return "403 - Forbidden"
+                AwsProvider::endpoint_for("api/token", true),
+                None,
+            )
+            .dispatch_put()?
+            .chain_err(|| "unwrapping aws imdsv2 token")?;
+        Ok(token)
     }
 
     fn fetch_ssh_keys(&self) -> Result<Vec<String>> {
@@ -71,7 +118,7 @@ impl AwsProvider {
             .client
             .get(
                 retry::Raw,
-                AwsProvider::endpoint_for("meta-data/public-keys"),
+                AwsProvider::endpoint_for("meta-data/public-keys", false),
             )
             .send()?;
 
@@ -86,10 +133,10 @@ impl AwsProvider {
                     .client
                     .get(
                         retry::Raw,
-                        AwsProvider::endpoint_for(&format!(
-                            "meta-data/public-keys/{}/openssh-key",
-                            tokens[0]
-                        )),
+                        AwsProvider::endpoint_for(
+                            &format!("meta-data/public-keys/{}/openssh-key", tokens[0]),
+                            false,
+                        ),
                     )
                     .send()?
                     .ok_or("missing ssh key")?;
@@ -107,7 +154,7 @@ impl MetadataProvider for AwsProvider {
         let add_value = |map: &mut HashMap<_, _>, key: &str, name| -> Result<()> {
             let value = self
                 .client
-                .get(retry::Raw, AwsProvider::endpoint_for(name))
+                .get(retry::Raw, AwsProvider::endpoint_for(name, false))
                 .send()?;
 
             if let Some(value) = value {
@@ -157,7 +204,7 @@ impl MetadataProvider for AwsProvider {
             .client
             .get(
                 retry::Json,
-                AwsProvider::endpoint_for("dynamic/instance-identity/document"),
+                AwsProvider::endpoint_for("dynamic/instance-identity/document", false),
             )
             .send()?
             .map(|instance_id_doc: InstanceIdDoc| instance_id_doc.region);
@@ -170,7 +217,10 @@ impl MetadataProvider for AwsProvider {
 
     fn hostname(&self) -> Result<Option<String>> {
         self.client
-            .get(retry::Raw, AwsProvider::endpoint_for("meta-data/hostname"))
+            .get(
+                retry::Raw,
+                AwsProvider::endpoint_for("meta-data/hostname", false),
+            )
             .send()
     }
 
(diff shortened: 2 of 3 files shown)


lucab:
@zonggen great! Just wondering, why are you putting this behind an opt-in CLI flag? My understanding is that the new v2 endpoint is always available (if the whole metadata service has not been disabled), no?

zonggen:
@lucab I see.. I thought we were using v1 by default and using v2 if specified to do so. That was misunderstanding on my part, updating :+1:

lucab:
@zonggen you have freedom of thought on what you think would be best approach here, I'm open to any proposal.
Though if we can avoid having users to opt-in/opt-out and automatically do the right thing safely, that's preferred in general.

dustin-decker:
There are two states an instance can be in:

1. Accept V1 + V2
2. Accept V2 only (this is safest)

Afterburn should support V2 by default.

zonggen:
Updated with the following logic:
 - Removed the option `--aws-imdsv2` and use IMDSv2 by default
 - By default, use IMDSv2, if the token retrieval failed, then use IMDSv1 as backup (since IMDSv2 = token retrieval + IMDSv1 with token header added)
 - Added tests for IMDS versions 

EDIT:
 - Lifted WIP, ready for review

<details open>
<summary>sample run on aws</summary>
<br>

```
[core ~]$ ./afterburn --cmdline --attributes output
Jun 04 16:11:50.977 DEBG logging initialized
Jun 04 16:11:50.978 TRCE found 'ignition.platform.id' flag: aws
Jun 04 16:11:50.979 TRCE cli configuration - Multi(CliMulti { attributes_file: Some("output"), check_in: false, hostname_file: None, network_units_dir: None, provider: "aws", ssh_keys_user: None })
Jun 04 16:11:50.980 DEBG command-line arguments parsed
Jun 04 16:11:50.998 INFO Putting http://169.254.169.254/latest/api/token: Attempt #1
Jun 04 16:11:51.001 INFO Put successful
Jun 04 16:11:51.002 INFO Fetching http://169.254.169.254/2009-04-04/meta-data/instance-id: Attempt #1
Jun 04 16:11:51.005 INFO Fetch successful
Jun 04 16:11:51.006 INFO Fetching http://169.254.169.254/2009-04-04/meta-data/instance-type: Attempt #1
Jun 04 16:11:51.008 INFO Fetch successful
Jun 04 16:11:51.009 INFO Fetching http://169.254.169.254/2009-04-04/meta-data/local-ipv4: Attempt #1
Jun 04 16:11:51.011 INFO Fetch successful
Jun 04 16:11:51.012 INFO Fetching http://169.254.169.254/2009-04-04/meta-data/public-ipv4: Attempt #1
Jun 04 16:11:51.014 INFO Fetch successful
Jun 04 16:11:51.015 INFO Fetching http://169.254.169.254/2009-04-04/meta-data/placement/availability-zone: Attempt #1
Jun 04 16:11:51.018 INFO Fetch successful
Jun 04 16:11:51.018 INFO Fetching http://169.254.169.254/2009-04-04/meta-data/hostname: Attempt #1
Jun 04 16:11:51.021 INFO Fetch successful
Jun 04 16:11:51.021 INFO Fetching http://169.254.169.254/2009-04-04/meta-data/public-hostname: Attempt #1
Jun 04 16:11:51.024 INFO Fetch successful
Jun 04 16:11:51.025 INFO Fetching http://169.254.169.254/2009-04-04/dynamic/instance-identity/document: Attempt #1
Jun 04 16:11:51.027 INFO Fetch successful
Jun 04 16:11:51.035 DEBG all tasks completed

[core@ ~]$ cat output 
AFTERBURN_AWS_AVAILABILITY_ZONE=us-east-1e
AFTERBURN_AWS_REGION=us-east-1
AFTERBURN_AWS_IPV4_PUBLIC=34.202.234.175
AFTERBURN_AWS_HOSTNAME=ip-172-31-55-46.ec2.internal
AFTERBURN_AWS_IPV4_LOCAL=172.31.55.46
AFTERBURN_AWS_PUBLIC_HOSTNAME=ec2-34-202-234-175.compute-1.amazonaws.com
AFTERBURN_AWS_INSTANCE_ID=i-0d69ec3c8e4e92afe
AFTERBURN_AWS_INSTANCE_TYPE=t2.micro
``` 
</details>

src/providers/aws/mod.rs
```diff
@@ -66,6 +68,23 @@ impl AwsProvider {
...
+        let tokendata: Option<String> = self
+            .client
+            .clone()
+            .header(header::HeaderName::from_bytes(b"X-aws-ec2-metadata-token-ttl-seconds")
+                        .chain_err(|| "setting header name for aws imdsv2 token")?,
+                    header::HeaderValue::from_bytes(b"21600")
+                        .chain_err(|| "setting header value for aws imdsv2 token")?)
+            .put(
+                retry::Raw,
+                format!("{}/{}", "http://169.254.169.254/latest", "api/token"),
```
bgilbert: Why are we hardcoding `latest` here?

bgilbert (commented):
It'd be good to squash these two commits together.  Adding and then removing a bunch of code clutters the history.

src/providers/aws/mod.rs
zonggen: Sorry I forgot to add a note on this.

So I've tried to fetch v2 token with different versions for example, 2009-04-04 and 2019-10-01, it seems like only `latest` is accepted and returned with a valid token, and others returned a 403 Forbidden.

zonggen:
Squashed commits and added a note on the hardcoded `latest` endpoint.

src/providers/aws/mock_tests.rs
```diff
@@ -86,3 +86,89 @@ fn test_aws_attributes() {
...
+        mocks.push(m);
+    }
+
+    let _m = mockito::mock("PUT", "/api/token")
+        .match_header("X-aws-ec2-metadata-token-ttl-seconds", "21600")
+        .with_status(403)
+        .with_body("Forbidden")
+        .create();
+
+
```
lucab, line 151: Nit: spurious newline.

src/providers/aws/mod.rs
```diff
@@ -66,6 +78,26 @@ impl AwsProvider {
...
 
+    fn fetch_imdsv2_token(client: retry::Client) -> Result<String> {
+        let token: String = client
+            .header(header::HeaderName::from_bytes(b"X-aws-ec2-metadata-token-ttl-seconds")
+                        .chain_err(|| "setting header name for aws imdsv2 token")?,
+                    header::HeaderValue::from_bytes(b"21600")
+                        .chain_err(|| "setting header value for aws imdsv2 token")?)
+            .put(
+                retry::Raw,
+                #[cfg(test)]
```
lucab: I think you can fold this build-condition back into `endpoint_for()` by adding a new `use_latest: bool` parameter there and passing `true` in the caller here.

Also, small nit, the `XXX: ` here below would better be a `NOTE(zonggen): ` (or whatever nick you prefer to sign yourself).

src/providers/aws/mod.rs
```diff
@@ -49,7 +50,18 @@ pub struct AwsProvider {
 
 impl AwsProvider {
     pub fn try_new() -> Result<AwsProvider> {
-        let client = retry::Client::try_new()?.return_on_404(true);
+        let mut client = retry::Client::try_new()?.return_on_404(true);
+        let token = AwsProvider::fetch_imdsv2_token(client.clone());
+
+        // If IMDSv2 token is fetched successfully, set the header.
+        // Otherwise, proceed with IMDSv1 mechanism.
+        if let Ok(t) = token {
```
lucab: I'd rather turn this into a full `match` and perform some kind of `log::warn!()` in the `Err` arm.

If I've correctly understood the docs, the only way for this to fail is when the metadata is disabled/firewalled completely. If that is the case, having a warning logged here would help detecting fishy setups when inspecting logs.

lucab:
Minor thing, Travis is gently reminding you to `cargo fmt` all your changes ;)

src/providers/aws/mod.rs
zonggen: Will proceed with `use_latest:bool` since that sounds very neat :) 
Updating >>

src/providers/aws/mod.rs
zonggen: I believe that is the case, will update with `log::warn!()`

zonggen: ~~Well I think my original logic failed to catch when `token` is an error and still returned `Ok` even if failed to fetch token.~~

Sorry for confusion

zonggen:
Updated with:
 - Removed extra newline
 - Added `use_latest: bool` to `endpoint_for`
 - Added warning message if failed to fetch imdsv2 session token
 - `cargo fmt` ;)

Rebased on master and ready for another look

src/providers/aws/mock_tests.rs
```diff
@@ -86,3 +86,88 @@ fn test_aws_attributes() {
...
+        "/meta-data/instance-type" => instance_type,
+        "/meta-data/local-ipv4" => ipv4_local,
+        "/meta-data/public-ipv4" => ipv4_public,
+        "/meta-data/placement/availability-zone" => availability_zone,
+        "/meta-data/hostname" => hostname,
+        "/meta-data/public-hostname" => public_hostname,
+        "/dynamic/instance-identity/document" => instance_id_doc,
+    };
+
+    // first test imdsv1
```
lucab, line 130: I think there is something fishy with the test as it takes a long time to complete, but I'm not sure why.

It would be good to start scoping the v1 and v2 sublogic into their own `{ ... }` scopes to make sure there isn't anything leaking from the former to the latter, that is:
```rust
 // first test imdsv1
{
let mut mocks = ...
...
}

// then test imdsv2
{
let token = ...
...
}
```

src/providers/aws/mod.rs
```diff
@@ -49,29 +50,74 @@ pub struct AwsProvider {
...
+            .header(
+                header::HeaderName::from_bytes(b"X-aws-ec2-metadata-token-ttl-seconds")
+                    .chain_err(|| "setting header name for aws imdsv2 token")?,
+                header::HeaderValue::from_bytes(b"21600")
+                    .chain_err(|| "setting header value for aws imdsv2 token")?,
+            )
+            .put(
+                retry::Raw,
+                #[cfg(test)]
+                AwsProvider::endpoint_for("api/token", false),
```
lucab: s/false/true/

But you can drop these two `cfg[]` branches here completely now, as you already moved the conditional part to `endpoint_for` declaration.

src/providers/aws/mod.rs
```diff
@@ -49,29 +50,74 @@ pub struct AwsProvider {
...
+                warn!("failed to fetch aws imdsv2 session token with: {}", err);
+            }
+        }
 
         Ok(AwsProvider { client })
     }
 
     #[cfg(test)]
-    fn endpoint_for(key: &str) -> String {
+    #[allow(unused_variables)]
```
lucab: The warning is because `use_latest` is not used. You can directly silence that by naming the argument as `_use_latest` (see [official docs](https://doc.rust-lang.org/stable/book/ch18-03-pattern-syntax.html#ignoring-an-unused-variable-by-starting-its-name-with-_)).

src/providers/aws/mock_tests.rs
zonggen: I feel like the long time comes from retries on the tested failing PUT requests. By default there will be 10 retries and 5 sec maximum back off time. But still the long running time is :(

Tried to time the test:
 - imdsv1 only: 91 sec
 - imdsv2 only: 42 sec
 - imdsv1 and imdsv2: 132 sec

But from AWS the `afterburn --cmdline --attributes output` finished almost instantaneously.

lucab: Ah, it could be that then. If you have a look at other platform mocks (e.g. `gcp`), you'll see we tweak down the number of retries. We should probably do the same for `aws` (possibly in another PR).

zonggen: Tested with `.max_retries(0)`, the test succeeded immediately.. nice :+1: 
~~Updating that bit as well~~     Updated

zonggen:
```diff
diff --git a/src/providers/aws/mod.rs b/src/providers/aws/mod.rs
index 23e2867..928cd80 100644
--- a/src/providers/aws/mod.rs
+++ b/src/providers/aws/mod.rs
@@ -50,6 +50,9 @@ pub struct AwsProvider {
 
 impl AwsProvider {
     pub fn try_new() -> Result<AwsProvider> {
+        #[cfg(test)]
+        let mut client = retry::Client::try_new()?.max_retries(0).return_on_404(true);
+        #[cfg(not(test))]
         let mut client = retry::Client::try_new()?.return_on_404(true);
         let token = AwsProvider::fetch_imdsv2_token(client.clone());
 
```

zonggen:
Updated addressing the reviews, also added `.max_retries(0)` to make sure tests exit immediately.

src/providers/aws/mock_tests.rs
lucab: Ah, nice. Can we keep the `max_retries` in here instead of making that conditional in `try_new()`?

zonggen: Yep, updating :+1: 

Added a `_try_new_with_client` in the AwsProvider and moved the `max_retries` to `mock_tests.rs`

zonggen:
Updated commit message and top-comment section for better clarification.

src/providers/aws/mod.rs
```diff
@@ -49,29 +50,92 @@ pub struct AwsProvider {
...
+            }
+            Err(err) => {
+                warn!("failed to fetch aws imdsv2 session token with: {}", err);
+            }
+        }
+
+        Ok(AwsProvider { client })
+    }
+
+    fn _try_new_with_client(mut client: retry::Client) -> Result<AwsProvider> {
```
lucab: Nit: `with_client` is probably enough as a name for a private method. Alternatively, `try_from_client` is ok too.

Also, you can drop the `mut` from the argument signature and rebind internally (because this function gets complete ownership of the client in input):
```rust
fn with_client(client: retry::Client) -> Result<AwsProvider> {
  let mut client = client;
  ...
}
```

src/providers/aws/mod.rs
```diff
@@ -49,29 +50,92 @@ pub struct AwsProvider {
 
 impl AwsProvider {
     pub fn try_new() -> Result<AwsProvider> {
-        let client = retry::Client::try_new()?.return_on_404(true);
+        let mut client = retry::Client::try_new()?.return_on_404(true);
+        let token = AwsProvider::fetch_imdsv2_token(client.clone());
```
lucab: Here, you can now use your new method and drop the logic below:
```rust
pub fn try_new() -> Result<AwsProvider> {
  let client = ...
  AwsProvider::with_client(client)
}
```

Doing this also helps making sure that the test-path and the reaworld-path are as close as possible.

src/retry/client.rs
```diff
@@ -208,6 +223,42 @@ where
...
+                .chain_err(|| "failed to build PUT request")?;
+
+            info!("Putting {}: Attempt #{}", req.url(), attempt + 1);
+            let response = self
+                .client
+                .execute(req)
+                .chain_err(|| "failed to PUT request")?;
+            let status = response.status();
+            if status.is_success() {
+                info!("Put successful");
```
lucab: Leftover from debug?

If you want to leave this in place but not visible by default, just make this a `trace!()`. Also, it's uppercase `PUT`.

src/providers/aws/mod.rs
zonggen: Updated, this has also de-dup the header setting portion. :+1:

src/retry/client.rs
zonggen: Added this since the `dispath_request` had `info!("Fetch successful");`, but I'm happy to remove it since `dispatch_post` does not have such `info!`

lucab: Ah yes, we are not very consistent in the way we log stuff. This does not seem to warrant a log output under normal cases though.

zonggen: Emm, actually right now when I run afterburn the log will be:
```
[core@ip-172-31-55-46 ~]$ ./afterburn --cmdline --attributes output 
Jun 10 13:44:10.374 DEBG logging initialized
Jun 10 13:44:10.400 TRCE found 'ignition.platform.id' flag: aws
Jun 10 13:44:10.401 TRCE cli configuration - Multi(CliMulti { attributes_file: Some("output"), check_in: false, hostname_file: None, network_units_dir: None, provider: "aws", ssh_keys_user: None })
Jun 10 13:44:10.401 DEBG command-line arguments parsed
Jun 10 13:44:10.446 INFO Putting http://169.254.169.254/latest/api/token: Attempt #1
Jun 10 13:44:10.450 INFO Put successful
Jun 10 13:44:10.454 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/instance-id: Attempt #1
Jun 10 13:44:10.457 INFO Fetch successful
Jun 10 13:44:10.458 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/instance-type: Attempt #1
Jun 10 13:44:10.460 INFO Fetch successful
Jun 10 13:44:10.461 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/local-ipv4: Attempt #1
Jun 10 13:44:10.463 INFO Fetch successful
Jun 10 13:44:10.464 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/public-ipv4: Attempt #1
Jun 10 13:44:10.467 INFO Fetch successful
Jun 10 13:44:10.468 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/placement/availability-zone: Attempt #1
Jun 10 13:44:10.470 INFO Fetch successful
Jun 10 13:44:10.471 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/hostname: Attempt #1
Jun 10 13:44:10.473 INFO Fetch successful
Jun 10 13:44:10.475 INFO Fetching http://169.254.169.254/2019-10-01/meta-data/public-hostname: Attempt #1
Jun 10 13:44:10.477 INFO Fetch successful
Jun 10 13:44:10.478 INFO Fetching http://169.254.169.254/2019-10-01/dynamic/instance-identity/document: Attempt #1
Jun 10 13:44:10.481 INFO Fetch successful
Jun 10 13:44:10.490 DEBG all tasks completed
```

~~if I remove it or change it to trace, there will be a little bit of disconnection between `Putting ...: Attempt #1` and `Fetching ... Attempt #1` since there are `Jun 10 13:44:10.457 INFO Fetch successful` after the PUT request. What do you think about just leave it here?~~

Now I totally see your point, I think it also makes sense to align the logging for HTTP PUT with POST request request since the `fetch` is not quite the same genre as PUT and POST.

lucab: That's also ok.

zonggen: Proceeding with removing this line, since `dispatch_post` does not `trace!` or `info!` the success POST request.

zonggen:
Updated addressing the reviews

lucab (approved):

PR #431, opened 2020-06-11
cargo: exclude tool configs from crate

bgilbert:


lucab:
This makes sense. Is it a cosmetic-only change or was it actively tripping some alert somewhere?

lucab (approved):

bgilbert:
I noticed it when copying Cargo.toml for ssh-key-dir.

PR #433, opened 2020-06-15
dracut: run hostname service on digitalocean

lucab:
The DHCP Offer on DigitalOcean nodes booted from custom images does
not seem to contain the hostname, which is instead available by
querying the metadata endpoint.
This adds the platform condition to the afterburn-hostname service
in order to let it run on DigitalOcean too.

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/538

### DIFF
--- a/dracut/30afterburn/afterburn-hostname.service
+++ b/dracut/30afterburn/afterburn-hostname.service
@@ -1,8 +1,10 @@
 [Unit]
 Description=Afterburn Hostname
-# Azure needs to fetch the hostname in the initramfs
+# These platforms do not provide the hostname via DHCP
+# options, thus it needs to be fetched from the metadata
+# and statically applied on first-boot.
 ConditionKernelCommandLine=|ignition.platform.id=azure
-# Exoscale needs to fetch the hostname in the initramfs
+ConditionKernelCommandLine=|ignition.platform.id=digitalocean
 ConditionKernelCommandLine=|ignition.platform.id=exoscale
 
 # We order this service after sysroot has been mounted


max3903 (approved):

jlebon (approved):

dustymabe (approved):
LGTM

dustymabe:
If you want to post up a dev build of a digitalocean artifact with this in it I can give it a spin and let you know if it appears to solve the problem.

lucab:
@dustymabe yup, possibly tomorrow. You can even directly hack in the `ConditionKernelCommandLine=|ignition.platform.id=digitalocean` in a dropin in fedora-coreos-config, I think.

dustymabe:
I applied the changes from this PR into my cosa overrides directory and then built an image.

```
$ cat overrides/rootfs/usr/lib/dracut/modules.d/30afterburn/afterburn-hostname.service 
[Unit]
Description=Afterburn Hostname
# These platforms do not provide the hostname via DHCP
# options, thus it needs to be fetched from the metadata
# and statically applied on first-boot.
ConditionKernelCommandLine=|ignition.platform.id=azure
ConditionKernelCommandLine=|ignition.platform.id=digitalocean
ConditionKernelCommandLine=|ignition.platform.id=exoscale

# We order this service after sysroot has been mounted
# but before ignition-files stage has run, so ignition can
# overwrite the configured hostname with e.g. a static one
# Also order after ignition-remount-sysroot just for
# safety for systems that mount /sysroot ro
After=initrd-root-fs.target
Before=ignition-files.service

OnFailure=emergency.target
OnFailureJobMode=isolate

[Service]
ExecStart=/usr/bin/afterburn --cmdline --hostname=/sysroot/etc/hostname
# Add hack to mark the file as needing relabelling, as the hostname
# file dropped by afterburn will be unlabelled causing SELinux denials.
# see: https://github.com/coreos/ignition/issues/635
ExecStart=/bin/sh -c 'mkdir -p /run/tmpfiles.d'
ExecStart=/bin/sh -c 'echo "z /etc/hostname - - -" > /run/tmpfiles.d/hostname-relabel.conf'
Type=oneshot
```

it looks good:


```
[core@bootstrap ~]$ journalctl -u afterburn-hostname.service 
-- Logs begin at Mon 2020-06-15 20:16:41 UTC, end at Mon 2020-06-15 20:17:53 UTC. --
Jun 15 20:16:45 localhost systemd[1]: Starting Afterburn Hostname...
Jun 15 20:16:45 localhost afterburn[660]: Jun 15 20:16:45.234 INFO Fetching http://169.254.169.254/metadata/v1.json: Attempt #1
Jun 15 20:16:45 localhost afterburn[660]: Jun 15 20:16:45.283 INFO Fetch successful
Jun 15 20:16:45 localhost systemd[1]: afterburn-hostname.service: Succeeded.
Jun 15 20:16:45 localhost systemd[1]: Finished Afterburn Hostname.
[core@bootstrap ~]$ 
[core@bootstrap ~]$ cat /etc/hostname 
bootstrap
```

PR #443, opened 2020-06-26
provider/vagrant: do not loop back local system hostname

lucab:
This reworks the vagrant-virtualbox provider in order to avoid
fetching the local system hostname and feeding it back to the
service which sets the local hostname (supposedly from external
metadata).
It also contains a minor cleanup of this module.

### DIFF
--- a/src/providers/vagrant_virtualbox/mod.rs
+++ b/src/providers/vagrant_virtualbox/mod.rs
@@ -30,15 +30,14 @@ use crate::providers::MetadataProvider;
 pub struct VagrantVirtualboxProvider;
 
 impl VagrantVirtualboxProvider {
-    pub fn new() -> VagrantVirtualboxProvider {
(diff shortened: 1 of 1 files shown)


lucab:
@bgilbert I'm also inclined to put this whole `vagrant-virtualbox` behind the legacy feature. My feeling is that this is unused outside of CL/flatcar, and even there it basically has no real logic other than re-exporting local system properties (duh). What's your feeling on that path?

bgilbert (approved):

bgilbert:
I agree with gating it behind `cl-legacy`.  We never shipped distro docs for `vagrant-virtualbox` anyway, and that OEM ID is kind of a hack.

lucab:
@bgilbert ack, thanks for the feedback, I'll follow up with a PR to clean that.

PR #450, opened 2020-06-30
providers: mark vagrant-virtualbox as legacy

lucab:
This demotes the "vagrant-virtualbox" provider to legacy only, as
the whole platform was a Container Linux hack that has not been
carried on.

Ref: https://github.com/coreos/afterburn/pull/443#issuecomment-650130910

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,5 @@ The following platforms are supported, with a different set of features availabl
   - Attributes
   - First-boot check-in
   - SSH Keys
-* vagrant-virtualbox
-  - Attributes
 * vmware
   - Custom network command-line arguments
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -61,7 +61,5 @@ The following platforms are supported, with a different set of features availabl
   - Attributes
   - First-boot check-in
   - SSH Keys
-* vagrant-virtualbox
-  - Attributes
```
bgilbert, line 65: These should be dropped from `docs/usage/attributes.md`.

src/metadata.rs
```diff
@@ -26,6 +26,7 @@ use crate::providers::ibmcloud::IBMGen2Provider;
 use crate::providers::ibmcloud_classic::IBMClassicProvider;
 use crate::providers::openstack::network::OpenstackProvider;
 use crate::providers::packet::PacketProvider;
+#[cfg(feature = "cl-legacy")]
```
bgilbert, line 29: Should the `pub mod` in `src/providers/mod.rs` also be flagged?

lucab: Absolutely yes, and clippy confirmed it too.

bgilbert (approved):

PR #462, opened 2020-07-22
openstack: use config-drive by default and metadata API as fallback

### LINKED ISSUE #422: Modify Afterburn openstack defaults to try config-drive, then metadata service
The afterburn folks think this is "not a bug": https://github.com/coreos/afterburn/issues/373

We should work around it here, and attempt to get SSH keys from MD service, especially if no IGN is provided.

See also about hostname: https://github.com/openshift/machine-config-operator/pull/964#issuecomment-513178768

```
State: idle
AutomaticUpdates: disabled
Deployments:
* ostree://fedora:fedora/x86_64/coreos/testing
                   Version: 31.20200310.2.0 (2020-03-12T01:21:36Z)
                    Commit: fef0f5a226ebf0e9915d8ee03a3bc70d3cc0593f0c22841264b3bc906dff0957
              GPGSignature: Valid signature by 7D22D5867F2A4236474BF7B850CB390B3C3359C4
```

zonggen:
This changes afterburn to use config-drive as default for scraping metadata and then try metadata API if config-drive is not available. Also changes the name of the platform from `openstack-metadata` to `openstack`.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/422


src/metadata.rs
```diff
@@ -62,7 +63,10 @@ pub fn fetch_metadata(provider: &str) -> errors::Result<Box<dyn providers::Metad
         "ibmcloud" => box_result!(IBMGen2Provider::try_new()?),
         // IBM Cloud - Classic infrastructure.
         "ibmcloud-classic" => box_result!(IBMClassicProvider::try_new()?),
-        "openstack-metadata" => box_result!(OpenstackProvider::try_new()?),
+        "openstack" => match OpenstackConfigDrive::try_new() {
```
zonggen: This should allow Afterburn to detect OpenStack and start scraping the metadata on first boot, though it's conflicting with the design decision in https://github.com/coreos/afterburn/issues/251.. Please verify @lucab

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,167 @@
...
+    temp_dir: Option<TempDir>,
+}
+
+impl OpenstackConfigDrive {
+    /// Try to build a new provider client.
+    ///
+    /// This internally tries to mount (and own) the config-drive.
+    pub fn try_new() -> Result<Self> {
+        // Short-circuit if the config-drive is already mounted.
+        let path = Path::new("/media/ConfigDrive/openstack/latest/");
```
zonggen: This path is copied from https://github.com/coreos/afterburn/blob/master/src/providers/cloudstack/configdrive.rs#L34, not entirely sure about the auto mount point of the config-drive. I think I might be missing some context on how the mount point is set up, the only documentation I found is https://coreos.com/os/docs/latest/cloud-config-locations.html

zonggen:
Some potential points of interest:
 - Now Afterburn requires super user permission since it might need to mount/umount the config-drive to read the metadata
 - Not entirely sure if the public key API is always available. Tried to boot up an FCOS instance on OpenStack but there's no public key data in either metadata API or config-drive files.

Except the public keys, other metadata can be successfully fetched:
```
core@ip_addr $ AFTERBURN_OPT_PROVIDER=--cmdline sudo ./afterburn ${AFTERBURN_OPT_PROVIDER} --attributes=./metadata

AFTERBURN_OPENSTACK_INSTANCE_TYPE=m1.small
AFTERBURN_OPENSTACK_IPV4_LOCAL=ip_addr
AFTERBURN_OPENSTACK_INSTANCE_ID=i-022da7a2
AFTERBURN_OPENSTACK_IPV4_PUBLIC=
AFTERBURN_OPENSTACK_HOSTNAME=abai-fcos-afterburn-test

```

src/metadata.rs
cgwalters: I think we still need to support anyone who was previously explicitly injecting `openstack-metadata` as a provider, which includes the MCO since https://github.com/openshift/machine-config-operator/commit/76e932b3c369b5c77e970f56bd74b3361dd9352e

src/providers/openstack/configdrive.rs
cgwalters: I would be surprised if FCOS/RHCOS shipped anything to do this by default, but Flatcar might.

cgwalters (commented):
Hmm...don't we basically want to try them *in parallel*?

zonggen:
> don't we basically want to try them in parallel?

@cgwalters By parallel do you mean literally running both config-drive and metadata API at the same time, or running the metadata API as a fallback if any of the method in config-drive version fails?

I was wondering too, since right now it won't check the metadata API unless something at the initialization step went wrong (like a missing config-drive device). For example, if the config-drive is mounted correctly, Afterburn won't check metadata API for ssh public keys even if public keys is missing in the meta_data.json filr in config-drive.

jlebon:
There's some more details in the chat logs from [the community meeting where this was discussed](https://github.com/coreos/fedora-coreos-tracker/issues/422#issuecomment-607513602), but the TL;DR IIUC was: we keep `openstack-metadata` for backwards compatibility, and we add a new `openstack` provider which does config-drive first, then falls back to metadata service.

The reason we do config drive first is to try to be consistent with Ignition's auto-discovery behaviour (which is not great, but we have to live with for now). If it used config-drive, then Afterburn should use config-drive too. Because we run so much later Ignition, we don't need to wait for the device to show up like Ignition does. We can just check for it and if it doesn't exist, fallback to the metadata server.

src/metadata.rs
lucab: Indeed, let's keep both `openstack` and `openstack-metadata` providers.

src/providers/openstack/configdrive.rs
lucab: It should be fine to drop this part. As long as all entities are mounting the config-drive read-only, it should be ok to have it mounted multiple time by different owners (each one handling the mount lifecycle on their own).

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,166 @@
...
+            }
+            Ok(out)
+        } else {
+            Ok(vec![])
+        }
+    }
+}
+
+impl MetadataProvider for OpenstackConfigDrive {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
```
lucab, line 142: It would be good to have some unit-tests on this (see the ibmcloud-* providers for examples).

src/metadata.rs
```diff
@@ -62,7 +63,10 @@ pub fn fetch_metadata(provider: &str) -> errors::Result<Box<dyn providers::Metad
...
+            Ok(result) => box_result!(result),
+            Err(_) => box_result!(OpenstackProviderNetwork::try_new()?),
+        },
```
lucab: I think this can be written as a simple `or_else` (but beware of matching types). Also, can you please:
 *  log a warning with the error from config-drive probing
 * move this small `try_autodetect()` logic to `openstack/mod.rs`

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        // Otherwise, try and mount with the label.
+        let target = tempfile::Builder::new()
+            .prefix("afterburn-")
+            .tempdir()
+            .chain_err(|| "failed to create temporary directory")?;
+        crate::util::mount_ro(
+            &Path::new("/dev/disk/by-label/").join(CONFIG_DRIVE_LABEL),
+            target.path(),
+            "iso9660",
```
lucab: Let's split out this constant to its own declaration.

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /// Return the path to the metadata directory.
+    fn metadata_dir(&self, platform: &str) -> PathBuf {
+        self.drive_path.clone().join(platform).join("latest")
+    }
+
+    // The metadata is stored as key:value pair in ec2/meta-data.json file
+    fn fetch_value(&self, key: &str) -> Result<Option<String>> {
+        let filename = self.metadata_dir("ec2").join("meta-data.json");
```
lucab: It would be good to have a real-world `meta-data.json` fixture and a test to cover this (see `ibmcloud-classic` for an existing example).

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        if let Some(value) = contents_json[key].as_str() {
+            Ok(Some(value.to_owned()))
+        } else {
+            Ok(None)
+        }
+    }
+
+    fn fetch_publickeys(&self) -> Result<Vec<PublicKey>> {
+        let filename = self.metadata_dir("openstack").join("meta_data.json");
```
lucab, line 122: It would be good to have a real-world `meta-data.json` fixture and a test to cover this (see `ibmcloud-classic` for an existing example).

src/metadata.rs
lucab: `README` and `attributes.md` also need to be updated with this provider.

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,166 @@
...
+            Ok(vec![])
+        }
+    }
+}
+
+impl MetadataProvider for OpenstackConfigDrive {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        let mut out = HashMap::with_capacity(5);
+        let add_value = |map: &mut HashMap<_, _>, key: &str, name| -> Result<()> {
+            let value = self.fetch_value(name)?;
```
lucab: I don't think this makes much sense, as it is going to re-open and re-read the file over and over again.
You can instead read it once, cache the metadata, and just extract the relevant keys here.

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    fn fetch_publickeys(&self) -> Result<Vec<PublicKey>> {
+        let filename = self.metadata_dir("openstack").join("meta_data.json");
+        let mut file =
+            File::open(&filename).chain_err(|| format!("failed to open file '{:?}'", filename))?;
+
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)
+            .chain_err(|| format!("failed to read from file '{:?}'", filename))?;
+        let contents_json: serde_json::Value = serde_json::from_str(&contents)?;
```
lucab: If this has a well-known and fixed schema, it would be better to define a serde-annotated struct and deserialize the content into that.

lucab (changes requested):

src/metadata.rs
zonggen: Moved to `openstack/mod.rs` and added a `warn!("failed to utilize config-drive, using the metadata service API instead");` after config-drive fails and before using metadata service API.

src/providers/openstack/configdrive.rs
zonggen: Added unit tests for attributes and ssh_keys.

src/providers/openstack/configdrive.rs
zonggen: Added `const TARGET_FS: &str = "iso9660";` for the file system type.

src/providers/openstack/configdrive.rs
zonggen: Added `tests/fixtures/openstack-config-drive/ec2/meta_data.json`.

src/providers/openstack/configdrive.rs
zonggen: Added `tests/fixtures/openstack-config-drive/openstack/meta_data.json`.

src/providers/openstack/configdrive.rs
zonggen: Removed the use of anonymous function and directly create the `MetadataEc2JSON` struct once and create the Hash Map on the spot.

src/providers/openstack/configdrive.rs
zonggen: Yes I think that makes sense. Updated with the corresponding structs with necessary fields.

zonggen:
Updated with followings:
 - support both `openstack-metadata` and `openstack` platform id
 - add unit tests for `attributes` and `ssh_keys`
 - create explicit structs for the JSON files
 - add fixtures for the `ec2/meta_data.json` and `openstack/meta_data.json`
 - update docs in `README.md` and `attributes.md`
 - minor adjustments according to comments / reviews

Also ran on OpenStack to make sure it worked as intended.. and it worked as expected

zonggen:
Ready for another round of review!

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,209 @@
...
+        serde_json::from_reader(input).chain_err(|| "failed parse JSON metadata")
+    }
+
+    /// The metadata is stored as key:value pair in ec2/latest/meta-data.json file
+    fn read_metadata(&self) -> Result<MetadataEc2JSON> {
+        let filename = self.metadata_dir("ec2").join("meta-data.json");
+        let file =
+            File::open(&filename).chain_err(|| format!("failed to open file '{:?}'", filename))?;
+        let bufrd = BufReader::new(file);
+        Self::parse_metadata_ec2(bufrd)
```
lucab, line 106: According to the reference doc linked at the top:
> Do not rely on the presence of the EC2 metadata in the metadata API or config drive, because this content might be removed in a future release. For example, do not rely on files in the ec2 directory.

Can we move to get as many details as we can from `parse_metadata_openstack` instead (e.g. hostname and AZ), and also make `parse_metadata_ec2` failures not stopping the whole process?

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,209 @@
...
+
+impl MetadataProvider for OpenstackConfigDrive {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        let metadata: MetadataEc2JSON = self.read_metadata()?;
+        Ok(maplit::hashmap! {
+            "OPENSTACK_HOSTNAME".to_string() => metadata.hostname.unwrap_or_default(),
+            "OPENSTACK_INSTANCE_ID".to_string() => metadata.instance_id.unwrap_or_default(),
+            "OPENSTACK_INSTANCE_TYPE".to_string() => metadata.instance_type.unwrap_or_default(),
+            "OPENSTACK_IPV4_LOCAL".to_string() => metadata.local_ipv4.unwrap_or_default(),
+            "OPENSTACK_IPV4_PUBLIC".to_string() => metadata.public_ipv4.unwrap_or_default(),
```
lucab: I think we can skip all the `None` values instead of producing empty-string values.

src/providers/openstack/configdrive.rs
zonggen: Yep, that makes sense. Was trying to align with the EC2 compatible metadata service we used in `openstack/network.rs`. I've shifted `hostname` to use from openstack config and use EC2 data as a backup.

src/providers/openstack/configdrive.rs
zonggen: Sure, updated with the `is_some()` checking.

src/metadata.rs
```diff
@@ -62,7 +62,7 @@ pub fn fetch_metadata(provider: &str) -> errors::Result<Box<dyn providers::Metad
         "ibmcloud" => box_result!(IBMGen2Provider::try_new()?),
         // IBM Cloud - Classic infrastructure.
         "ibmcloud-classic" => box_result!(IBMClassicProvider::try_new()?),
-        "openstack-metadata" => box_result!(OpenstackProvider::try_new()?),
+        "openstack" | "openstack-metadata" => openstack::try_config_drive_else_network(),
```
lucab: I'd suggest to keep the two providers split and directly wire `openstack-metadata` to `OpenstackProviderNetwork`.

src/providers/openstack/mod.rs
```diff
@@ -14,4 +14,20 @@
...
+use slog_scope::warn;
+
+pub mod configdrive;
 pub mod network;
+
+pub fn try_config_drive_else_network() -> errors::Result<Box<dyn providers::MetadataProvider>> {
+    if let Ok(config_drive) = OpenstackConfigDrive::try_new() {
+        Ok(Box::new(config_drive))
+    } else {
+        warn!("failed to utilize config-drive, using the metadata service API instead");
```
lucab: s/utilize/locate/

src/providers/openstack/mod.rs
```diff
@@ -14,4 +14,20 @@
...
+use crate::errors;
+use crate::providers;
+use configdrive::OpenstackConfigDrive;
+use network::OpenstackProviderNetwork;
+use slog_scope::warn;
+
+pub mod configdrive;
 pub mod network;
+
+pub fn try_config_drive_else_network() -> errors::Result<Box<dyn providers::MetadataProvider>> {
```
lucab, line 29: A short docstring on this would be helpful, in order to mention why we do this serially.

src/providers/openstack/configdrive.rs
```diff
@@ -0,0 +1,273 @@
...
+use crate::network;
+use crate::providers::MetadataProvider;
+
+const CONFIG_DRIVE_LABEL: &str = "config-2";
+
+/// Partial object for ec2 `meta_data.json`
+#[derive(Debug, Deserialize)]
+pub struct MetadataEc2JSON {
+    /// Local hostname.
+    pub hostname: Option<String>,
```
lucab: I think you can drop this from here as this detail is now coming from `MetadataOpenstackJSON`.

src/providers/openstack/configdrive.rs
lucab: What about using `if let Some(val) = metadata.foo { ... }` instead? That would avoid all the `unwrap`s.

zonggen:
@lucab Thanks for the review! Updated regarding the reviews.

lucab (approved):

lucab:
LGTM, thanks for implementing this!

PR #463, opened 2020-07-24
providers: add an AzureStack stub

darkmuggle:
Stub out AzureStack wire-framing in anticipation of working on AzureStack in the near future. This stub out should allow exploration on the AzureStack platform. 

As a platform AzureStack does not provide either a meta-data service nor does it provide a useful OVF XML the utility of Afterburn is a pretty much a noop for AzureStack.

Signed-off-by: Ben Howard <ben.howard@redhat.com>

### DIFF
--- a/src/providers/azurestack/mod.rs
+++ b/src/providers/azurestack/mod.rs
@@ -1,0 +1,33 @@
+// Copyright 2020 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This is a stub provider. Right now the AzureStack provider does
+// asbolutely nothing interesting.
+
+//! azurestack/azurestack metadata fetcher
+use crate::providers::MetadataProvider;
+
+use slog_scope::debug;
+
+#[derive(Clone, Copy, Debug)]
+pub struct AzureStack;
+
+impl AzureStack {
+    pub fn new() -> Self {
+        debug!("azure stack provider is a noop stub");
+        Self
+    }
+}
+
+impl MetadataProvider for AzureStack{}
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -17,6 +17,7 @@ use crate::providers;
 use crate::providers::aliyun::AliyunProvider;
 use crate::providers::aws::AwsProvider;
 use crate::providers::azure::Azure;
+use crate::providers::azurestack::AzureStack;
 use crate::providers::cloudstack::configdrive::ConfigDrive;
 use crate::providers::cloudstack::network::CloudstackNetwork;
 use crate::providers::digitalocean::DigitalOceanProvider;
@@ -48,6 +49,7 @@ pub fn fetch_metadata(provider: &str) -> errors::Result<Box<dyn providers::Metad
         #[cfg(not(feature = "cl-legacy"))]
         "aws" => box_result!(AwsProvider::try_new()?),
         "azure" => box_result!(Azure::try_new()?),
+        "azurestack" => box_result!(AzureStack::new()),
         "cloudstack-metadata" => box_result!(CloudstackNetwork::try_new()?),
         "cloudstack-configdrive" => box_result!(ConfigDrive::try_new()?),
         "digitalocean" => box_result!(DigitalOceanProvider::try_new()?),
--- a/src/providers/aliyun/mod.rs
+++ b/src/providers/aliyun/mod.rs
@@ -6,11 +6,10 @@
 #[cfg(test)]
 use mockito;
 use openssh_keys::PublicKey;
-use slog_scope::{error, warn};
+use slog_scope::error;
 use std::collections::{BTreeSet, HashMap};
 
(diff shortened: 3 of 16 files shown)


darkmuggle:
Here is an OVF-XML file from AzureStack. You'll notice that today it treats Linux VM's as Windows.
```
<?xml version="1.0" encoding="utf-8"?>
<Environment xmlns="http://schemas.dmtf.org/ovf/environment/1" xmlns:oe="http://schemas.dmtf.org/ovf/environment/1" xmlns:wa="http://schemas.microsoft.com/windowsazure" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

    <wa:ProvisioningSection><wa:Version>1.0</wa:Version><WindowsProvisioningConfigurationSet xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><ConfigurationSetType>WindowsProvisioningConfiguration</ConfigurationSetType><ComputerName>prf000000</ComputerName><AdminPassword>!!123abc</AdminPassword><EnableAutomaticUpdates>true</EnableAutomaticUpdates><AdminUsername>tenantuser</AdminUsername></WindowsProvisioningConfigurationSet></wa:ProvisioningSection>

  <wa:PlatformSettingsSection><wa:Version>1.0</wa:Version><PlatformSettings xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><KmsServerHostname i:nil="true" /><ProvisionGuestAgent>true</ProvisionGuestAgent><GuestAgentPackageName>WindowsGuestAgent.zip</GuestAgentPackageName><UseAVMA>true</UseAVMA></PlatformSettings></wa:PlatformSettingsSection>
</Environment>
```

The ISO also includes an `unattend.xml`: 
```
<?xml version="1.0" encoding="utf-8"?>
<unattend xmlns="urn:schemas-microsoft-com:unattend">
    <settings pass="generalize">
[REDACTED]
         <component name="Microsoft-Windows-Shell-Setup" processorArchitecture="amd64" publicKeyToken="31bf3856ad364e35" language="neutral" versionScope="nonSxS" xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
            <TimeZone>UTC</TimeZone>
            <ComputerName>prf000000</ComputerName>
        </component>
[REDACTED]
```

I would need word from Microsoft or API documentation, but the example `ovf-env.xml` and `unattend.xml` include references to the `<ComputerName>` which could be used. Regardless the purpose of the stub is simple: provide the wire-framing.

darkmuggle:
FWIW, I'm am happy to leave this a WIP if the Afterburn overseers would prefer a "complete"  provider pending information from MS.

ashcrow:
@darkmuggle thanks for starting this! You noted this is kind of a `noop` and, that makes sense. It's primarily scaffolding and boilerplate to match the trait. I'm curious what folks think about these ideas (which shouldn't block this PR! -- just thinking out loud):

- having a literal `noop` MetadataProvider for reuse in these types of instances (avoiding copy/paste implementations with the same boilerplate)?
- moving these `noop` function implementations into the [trait](https://github.com/coreos/afterburn/blob/master/src/providers/mod.rs#L198-L202) and move to overriding default implementations?

/cc @lucab

darkmuggle:
In regards to the `noop` provider,  https://github.com/darkmuggle/afterburn/commit/6369339b3fd7ebe44e1bbeb8e30f1c45fdf7b6a5 is a novice take on the idea.  And using traits https://github.com/darkmuggle/afterburn/commit/8f77a41ff2337ae1d99a94e0251d9d3d103fe352

src/providers/azurestack/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+            .chain_err(|| "unable to get hostname")?
+            .to_string_lossy()
+            .into_owned();
+        Ok(hostname)
+    }
+}
+
+impl MetadataProvider for AzureStack {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        let hostname = Self::system_hostname()?;
```
lucab: I don't think it is useful to pipe back the local hostname here, we can plainly drop this.

lucab:
@ashcrow @darkmuggle yes we can add default implementation for trait methods, that would help de-duplicating the same empty logic in a bunch of places. I'm not very convinced about a `noop` platform though.

lucab:
> I would need word from Microsoft or API documentation, but the example ovf-env.xml and unattend.xml include references to the <ComputerName> which could be used.

Just for clarity, does the `ComputerName` in there "prf000000" match the name of the instance?

darkmuggle:
> > I would need word from Microsoft or API documentation, but the example ovf-env.xml and unattend.xml include references to the  which could be used.
> 
> Just for clarity, does the `ComputerName` in there "prf000000" match the name of the instance?

TBH, I *think* that is the case, but I'm not sure. The snippets came via Microsoft. In fact, when I did a deep dive on the XML, I have serious questions about whether it was provisioned as Windows or Linux virtual machine; the ISO contains *windows* PowerShell commandlets.

> @ashcrow @darkmuggle yes we can add default implementation for trait methods, that would help de-duplicating the same empty logic in a bunch of places. I'm not very convinced about a noop platform though. 

Yeah, I've not convinced myself. I was learning me some Rust as my Hack-day, and thought that I would do something potentially useful; hence this PR. After filing the PR, I realized that `noop` is kind of pointless given that in CoreOS SystemD units are used. 

I'll update the PR to:
- use traits
- drop the noop provider

lucab:
@darkmuggle did you close this by mistake or in preparation for a further step?

darkmuggle:
I was going to rebase and submit as two sperate PRs:
- Metadata trait refactor
- follow up stub

The diff was pretty ugly, IMHO.

PR #468, opened 2020-07-28
azure: rework ready-state posting

### LINKED ISSUE #459: azure: keep retrying requests a bit longer
On Azure, Afterburn currently sends requests to the metadata/wireserver endpoint with some mildly aggressive retrying logic, which on failures results in sending about 10 requests in ~1 minute.

According to https://github.com/Azure/azure-sdk-for-go/issues/1848, the WireServer routinely goes down for maintenance with a p90 downtime up to 3 minutes (!) for each restart event.

As such, in order to be more lenient toward transient platform issues, we should relax our deadline for retries as to cover extended downtime up to 4 or 5 minutes.

### LINKED ISSUE #461: azure: boot reporting fails if boot diagnostic is enabled after instance creation
If an Azure instance is created with "Boot diagnostics: off" and then later on while booting the setting is turned to "on", Afterburn boot check-in will fail with an output like this:

```
Jul 21 13:21:25.449 INFO Fetching http://168.63.129.16/?comp=versions: Attempt #1
Jul 21 13:21:25.455 INFO Fetch successful
Jul 21 13:21:25.462 INFO Fetching http://168.63.129.16/machine/?comp=goalstate: Attempt #1
Jul 21 13:21:25.468 INFO Fetch successful
Jul 21 13:21:25.506 INFO Posting http://168.63.129.16/machine/?comp=health: Attempt #1
Jul 21 13:21:26.520 INFO Posting http://168.63.129.16/machine/?comp=health: Attempt #2
...
Jul 21 13:22:02.653 INFO Posting http://168.63.129.16/machine/?comp=health: Attempt #10
Error: failed to run
Caused by: checking-in instance boot to cloud provider
Caused by: maximum number of retries (10) reached
Caused by: POST failed: 410 Gone
```

This seems unrelated to https://github.com/coreos/afterburn/issues/459, i.e. it isn't a timeout issue.

I played a guessing game on the root cause for this, and via empirical experimentation it seems to be due to a mismatch in `GoalStateIncarnation`.

In particular, [Azure documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/no-agent#generic-steps-without-using-python) tells us to hardcode `<GoalStateIncarnation>1</GoalStateIncarnation>` in requests.
However, flipping the setting seems to increment the machine incarnation, and querying the goalstate endpoint after that returns an XML document with `<Incarnation>2</Incarnation>`.
Manually tweaking the request to have a matching `Incarnation <-> GoalStateIncarnation` seems to fix the failure, even after further tweaking settings (manually confirmed with incarnations up to 4).

lucab:
This refactors the Azure provider, in order to better report instance
ready-state to the WireServer.
In particular, it looks like the platform expects a matching
`GoalStateIncarnation` to be posted in ready-state requests.
For this reason, this stops caching the goalstate content when creating
the provider. Instead, the goalstate is now fetched just in time,
and the incarnation number is used when posting the ready-state.
This also ensures that on transient errors all the endpoints are retried
long enough.

Closes: https://github.com/coreos/afterburn/issues/459
Closes: https://github.com/coreos/afterburn/issues/461

### DIFF
--- a/src/providers/azure/goalstate.rs
+++ b/src/providers/azure/goalstate.rs
@@ -1,0 +1,120 @@
+//! Logic to interact with WireServer `goalstate` endpoint.
+
+use crate::errors::*;
+use serde_derive::Deserialize;
+
+#[derive(Debug, Deserialize, Clone)]
+pub(crate) struct GoalState {
+    #[serde(rename = "Container")]
+    pub container: Container,
+    #[serde(rename = "Incarnation")]
+    incarnation: String,
+}
+
+impl GoalState {
+    /// Return the certificates endpoint (if any).
+    pub(crate) fn certs_endpoint(&self) -> Option<String> {
+        let role = match self.container.role_instance_list.role_instances.get(0) {
+            Some(r) => r,
+            None => return None,
+        };
+
+        role.configuration.certificates.clone()
+    }
+
+    /// Return this instance `ContainerId`.
+    pub(crate) fn container_id(&self) -> &str {
+        &self.container.container_id
+    }
+
+    /// Return this instance `InstanceId`.
+    pub(crate) fn instance_id(&self) -> Result<&str> {
+        Ok(&self
+            .container
+            .role_instance_list
+            .role_instances
+            .get(0)
+            .ok_or_else(|| "empty RoleInstanceList".to_string())?
+            .instance_id)
+    }
+
+    /// Return the current `Incarnation` count for this instance.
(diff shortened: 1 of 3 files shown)


lucab:
Manually tested, this is now ready for review.

/cc @jlebon @bgilbert

cgwalters (approved):
Read the code, outline looks sane; but didn't verify details.
LGTM

PR #469, opened 2020-07-29
azure: log a warning on network failure if non-root

### LINKED ISSUE #95: azure: log a warning on network failure if non-root
This is related to https://github.com/coreos/coreos-metadata/issues/94. If we experience a network failure on Azure while trying to reach the metadata endpoint and the process is not running as root, we should log a more informative message about possible Azure-specific local firewalling,

bh7cw:
This logs a warning about possible Azure-specific local firewalling if unable to reach Azure endpoints and running as non-root.

Closes: https://github.com/coreos/afterburn/issues/95

### DIFF
--- a/src/providers/azure/mod.rs
+++ b/src/providers/azure/mod.rs
@@ -30,6 +30,7 @@ use crate::errors::*;
 use crate::network;
 use crate::providers::MetadataProvider;
 use crate::retry;
+use nix::unistd::Uid;
 
 #[cfg(test)]
 mod mock_tests;
@@ -216,6 +217,15 @@ impl Azure {
         // make sure the metadata service is compatible with our version
         azure
             .is_fabric_compatible(MS_VERSION)
+            .map_err(|e| {
+                //It may require to run as root in order to reach the metadata endpoint on Azure, more details: https://github.com/coreos/bugs/issues/2468
+                let is_root = Uid::current().is_root();
+                if !is_root {
+                    warn!("unable to reach Azure endpoints, please check whether firewall rules are blocking access to them");
+                }
+
+                e
+            })
             .chain_err(|| "failed version compatibility check")?;
 
         // populate goalstate


bh7cw:
Manually tested on azure, test result is:
```
[AzureUser@rhel test729]$ ./afterburn --provider azure --attributes /dev/stdout
(omit info)
Jul 29 12:23:21.351 INFO Fetching http://168.63.129.16/?comp=versions: Attempt #11
Jul 29 12:23:51.352 INFO Failed to fetch: error sending request for url (http://168.63.129.16/?comp=versions): operation timed out
Jul 29 12:23:51.352 WARN unable to reach Azure endpoints, please check whether firewall rules are blocking access to them
Error: failed to run
Caused by: fetching metadata from provider
Caused by: failed to get shared configuration: not found
```
This is now ready for review.
/cc @lucab

src/providers/azure/mod.rs
```diff
@@ -216,7 +217,15 @@ impl Azure {
         // make sure the metadata service is compatible with our version
         azure
             .is_fabric_compatible(MS_VERSION)
-            .chain_err(|| "failed version compatibility check")?;
+            .map_err(|_| {
+                let user = Uid::current();
+                let root_or_not = Uid::is_root(user);
```
lucab: `let is_root = Uid::current().is_root();`

src/providers/azure/mod.rs
```diff
@@ -216,7 +217,15 @@ impl Azure {
...
+                if !root_or_not {
+                    warn!("unable to reach Azure endpoints, please check whether firewall rules are blocking access to them");
+                }
+
+                "failed to get shared configuration: not found"
```
lucab: I think it would be better to keep the closure parameter and propagate it here (beware of types).
That is, something like:
```
.map_err(|e| { ... ; e})
```

src/providers/azure/mod.rs
```diff
@@ -216,7 +217,15 @@ impl Azure {
         // make sure the metadata service is compatible with our version
         azure
             .is_fabric_compatible(MS_VERSION)
-            .chain_err(|| "failed version compatibility check")?;
```
lucab, line 219: I'd keep this.

src/providers/azure/mod.rs
```diff
@@ -216,7 +217,15 @@ impl Azure {
...
+            .map_err(|_| {
+                let user = Uid::current();
+                let root_or_not = Uid::is_root(user);
+                if !root_or_not {
+                    warn!("unable to reach Azure endpoints, please check whether firewall rules are blocking access to them");
```
lucab, line 224: It would be good to have a short comment on top of this to mention why we specifically care about root, and a link to https://github.com/coreos/bugs/issues/2468.

src/providers/azure/mod.rs
bh7cw: Thank you so much for review. I tested `.map_err(|e| { ... ; e})`, it works and show the clear chain errors.

lucab (approved):

PR #475, opened 2020-08-05
ibmcloud: add support for SSH keys

erlarese:
This PR adds support to the ibmcloud provider for parsing SSH keys.  This applies to the IBM Cloud VPC Gen 2 environment.  It does not apply to other aspects of IBM Cloud including the "classic" environment.

This code looks for the `vendor-data` file in the `cidata` drive and finds the `cloud-config` section within that file.  A YAML parser is used to pull the SSH keys out from this section.

This is my first time writing any Rust code, so I apologize for anything that's obviously wrong.  I have attempted to update the unit tests and verify they still pass.  Please let me know if anything else is needed !


lucab:
@erlarese Thanks for the PR! I currently have a release in progress at https://github.com/coreos/afterburn/issues/473, so I'll queue the review here after that. Also, the CI is on fire today, so don't worry about that right now.

Before I start with a first pass on this, can you please squash everything in a single commit?

erlarese:
I attempted to squash all 13 commits into one, but I appear to have done it poorly and somehow pulled in a bunch of other commits.  I think I'm just going to create a new fork and start over with a new PR.  Oops.

lucab:
@erlarese as you prefer, but from the current branch status I think it is only missing a rebase on master (which should automatically detect and drop the other commits), see https://git-scm.com/book/en/v2/Git-Branching-Rebasing for guidance.

erlarese:
Thanks Luca for the pointer.  I think I hopefully got everything squashed into a single commit and rebased now.

Cargo.toml
```diff
@@ -52,6 +52,7 @@ tempfile = "^3.1"
 update-ssh-keys = { version = "^0.6", optional = true }
 users = "^0.10"
 vmw_backdoor = "^0.1"
+yaml-rust = "^0.4"
```
lucab: Any reason for not using https://docs.rs/serde_yaml directly? It looks like you manually wrote a deserializer, while you could more quickly declare the data-structure you want to handle, and then add `derive(Deserialize)` on it (see examples from other providers in this repo).

src/providers/ibmcloud/mod.rs
```diff
@@ -93,6 +95,21 @@ impl IBMGen2Provider {
         Ok(output)
     }
 
+    /// Return the path to the vendordata directory.
+    fn vendordata_dir(&self) -> PathBuf {
```
lucab: This seems to be same as `metadata_dir()`. If those file are in the same directory, you can omit and just re-use the existing method.

src/providers/ibmcloud/mod.rs
```diff
@@ -112,6 +129,49 @@ impl IBMGen2Provider {
         }
         output
     }
+
+    // Find the SSH keys in the vendordata file
```
lucab: Docstrings needs to be marked with `///`.

src/providers/ibmcloud/mod.rs
```diff
@@ -112,6 +129,49 @@ impl IBMGen2Provider {
...
+    fn fetch_ssh_keys(vendordata: String) -> Result<Vec<String>> {
+        // The ibmcloud vendor data file contains multiple sections, start by finding the section boundary
+        let v: Vec<&str> = vendordata.split("boundary=").collect();
+        let boundary_delimiter = v[1];
```
lucab: Unfortunately this kind of unchecked indexed access (here and everywhere else) will panic as soon as the input does not match your expectations. Manually parsing data is usually not a good idea.

You may try using a proper mail parser instead, for example https://github.com/staktrace/mailparse, pick up the first entry with a `text/cloud-config` type and pass its content to the YAML deserializer.

src/providers/ibmcloud/mod.rs
```diff
@@ -93,6 +95,21 @@ impl IBMGen2Provider {
         Ok(output)
     }
 
+    /// Return the path to the vendordata directory.
+    fn vendordata_dir(&self) -> PathBuf {
+        self.drive_path.clone()
+    }
+
+    /// Read vendordata file into a string
+    fn read_vendordata(&self) -> Result<String> {
```
lucab: This is not really a String, just a buffer of bytes (i.e. `Result<Vec<u8>>`).

src/providers/ibmcloud/mod.rs
```diff
@@ -93,6 +95,21 @@ impl IBMGen2Provider {
...
+        let filename = self.vendordata_dir().join("vendor-data");
+        let mut file =
+            File::open(&filename).chain_err(|| format!("Failed to open vendordata '{:?}'", filename))?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents).expect("Unable to read vendordata");
```
lucab: Please do not expect/unwrap in main logic (i.e. everywhere outside of tests), prefer bubbling up the error instead.

src/providers/ibmcloud/mod.rs
erlarese: Removed `vendordata_dir` in the next commit and used `metadata_dir` instead.

src/providers/ibmcloud/mod.rs
erlarese: Updated in the next commit.

src/providers/ibmcloud/mod.rs
erlarese: Fixed in the next commit.

src/providers/ibmcloud/mod.rs
erlarese: Removed `expect` in next commit.

cgwalters:
> This is my first time writing any Rust code, so I apologize for anything that's obviously wrong. 

No worries, and thanks for submitting this PR!  Personally it took me quite a while to learn Rust but it was also a rewarding process.  Luca already went over some things, but in general don't hesitate to reach out with questions and the like on e.g. #forum-coreos on freenode.

src/providers/ibmcloud/mod.rs
erlarese: I rewrote the code to use mailparse in the next commit.

Cargo.toml
erlarese: Hi Luca, no I didn't have a particular reason for not using serde.  I just googled "rust yaml" and started working with the first library that came up.  I rewrote the code to use serde_yaml and a deserializer with a struct in the next commit.  I attempted to follow the example used in the Azure provider.

erlarese (commented):
Hi Luca, thanks for the feedback!  I have attempted to implement everything.  Please let me know what you think now.  Thanks !

src/providers/ibmcloud/mod.rs
```diff
@@ -23,6 +24,10 @@ use crate::errors::*;
 use crate::network;
 use crate::providers::MetadataProvider;
 
+use mailparse::*;
+
```
lucab: Nit: you can remove this newline and move all the `use` statements together (`rustfmt` will sort them properly).

src/providers/ibmcloud/mod.rs
```diff
@@ -112,6 +127,27 @@ impl IBMGen2Provider {
...
+    /// Find the SSH keys in the vendordata file
+    fn fetch_ssh_keys(vendordata_vec: Vec<u8>) -> Result<Vec<String>> {
+        // Parse MIME format from vendor-data file
+        let vendor_data_mail = parse_mail(&vendordata_vec).unwrap();
+        let mut cloud_config = String::new();
+        for section in vendor_data_mail.subparts {
+            for header in &section.headers {
+                match header.get_value().as_str() {
+                    "text/cloud-config" => {
+                        cloud_config = section.get_body().unwrap().clone();
```
lucab:
```suggestion
                        cloud_config = section
                            .get_body()
                            .chain_err(|| "failed to get cloud-config content")?;
```

src/providers/ibmcloud/mod.rs
```diff
@@ -112,6 +127,27 @@ impl IBMGen2Provider {
...
+                        break;
+                    }
+                    _ => {}
+                }
+            }
+        }
+        // Parse YAML to find SSH keys
+        let deserialized_cloud_config: VendorDataCloudConfig = serde_yaml::from_str(&cloud_config).unwrap();
```
lucab:
```suggestion
        let deserialized_cloud_config: VendorDataCloudConfig = serde_yaml::from_str(&cloud_config)
            .chain_err(|| "failed to deserialize cloud-config content")?;
```

src/providers/ibmcloud/mod.rs
```diff
@@ -159,6 +202,11 @@ impl Drop for IBMGen2Provider {
     }
 }
 
+#[derive(Debug, Deserialize, Clone)]
+struct VendorDataCloudConfig {
```
lucab, line 202: Nit: a docstring on this one would be appreciated.

lucab:
@erlarese thanks, great work! I've left a few minor nits and a bunch of suggestions in order to avoid a couple of unwrapping points.
In the failed CI jobs there are a few more suggestions from clippy and rustfmt (you can run those locally too, check Travis config to see how they run).

src/providers/ibmcloud/mod.rs
erlarese: It looks like the Travis config is using `cargo fmt -- --check -l` and `cargo clippy -- -D warnings` and neither of those seemed to suggest changes to the `use` statements?  I might be looking at the wrong Travis config, or using the wrong version of the linters, or perhaps the wrong rulesets.  I removed this newline manually in the next commit.

src/providers/ibmcloud/mod.rs
erlarese: I added a docstring in the next commit.

erlarese:
I'm not sure if this is candidate for merging yet, but please hold off on merging for now.  I just built an FCOS image with the latest patches and afterburn is failing to run, I'm still debugging what exactly is wrong.

```
afterburn-sshkeys@core.service: Main process exited, code=exited, status=1/FAILURE
afterburn-sshkeys@core.service: Failed with result 'exit-code'.
```

lucab:
@erlarese no hurry. A `journalctl -u afterburn-sshkeys@core.service` should you all the details.
However, when trying to investigate issues, I usually do a debug build (which produces a larger binary, with all symbols and full debugging on) and run it manually on the host.

erlarese:
Hi @lucab , thanks for the tip.  The problem was that the IBM Cloud VPC control plane was generating a MIME `vendor-data` file that actually contained two sub sections with the `cloud-config` header.  The switch to mailparse and serde caused it to match on the second cloud-config section which doesn't contain the SSH keys.  My unit test passed anyway because I had used an abbreviated version of the `vendor-data` file to keep the UT smaller.

I fixed this by adding an additional check for a cloud-config section with `ssh_authorized_keys` inside, and I also expanded the UT to a full actual vendor-data file generated by the IBM Cloud (with SSH keys snipped).

With these updates, it appears the unit tests pass, the linters pass, and an FCOS image built with this boots successfully on the IBM Cloud and I can connect via SSH.

Please re-review and let me know if anything else is needed, thanks !

src/providers/ibmcloud/mod.rs
```diff
@@ -196,4 +249,116 @@ foo:      ba:r
             Some(&"test_instance-vpc-gen2".to_string())
         );
     }
+
+    #[test]
+    fn test_fetch_ssh_keys() {
+        let vendordata = concat!(
```
lucab: I think you are manually escaping a lot of things in here. You may consider using a [raw string](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals) instead:
```rust
let vendordata = 
r#"your
raw
"quotes"
in
here!"#;
```

or directly put this content into a fixture file (under `tests/fixtures/`).

src/providers/ibmcloud/mod.rs
```diff
@@ -112,6 +126,29 @@ impl IBMGen2Provider {
...
+
+    /// Find the SSH keys in the vendordata file
+    fn fetch_ssh_keys(vendordata_vec: Vec<u8>) -> Result<Vec<String>> {
+        // Parse MIME format from vendor-data file
+        let vendor_data_mail = parse_mail(&vendordata_vec).unwrap();
+        let mut cloud_config = String::new();
+        for section in vendor_data_mail.subparts {
+            for header in &section.headers {
+                if let "text/cloud-config" = header.get_value().as_str() {
+                    if section.get_body().unwrap().contains("ssh_authorized_keys") {
```
lucab: I think you can replace this `unwrap` with an `unwrap_or_default`.

src/providers/ibmcloud/mod.rs
```diff
@@ -112,6 +126,29 @@ impl IBMGen2Provider {
...
+                        cloud_config = section
+                            .get_body()
+                            .chain_err(|| "failed to get cloud-config content")?;
+                        break;
+                    }
+                }
+            }
+        }
+        // Parse YAML to find SSH keys
+        let deserialized_cloud_config: VendorDataCloudConfig = serde_yaml::from_str(&cloud_config)
```
lucab, line 154: [Minor UX concern]
There is a chance that `cloud_config` is still an empty string at this point.
I think it would be good to have an `is_empty()` check before this and return an helpful error (something like "no cloud-config section found in vendor-data") before trying to deserialize.

src/providers/ibmcloud/mod.rs
```diff
@@ -112,6 +126,29 @@ impl IBMGen2Provider {
         }
         output
     }
+
+    /// Find the SSH keys in the vendordata file
+    fn fetch_ssh_keys(vendordata_vec: Vec<u8>) -> Result<Vec<String>> {
+        // Parse MIME format from vendor-data file
+        let vendor_data_mail = parse_mail(&vendordata_vec).unwrap();
```
lucab: This `unwrap` should be replaced by a `.chain_err(...)?` or `map_err(...)?` instead.

src/providers/ibmcloud/mod.rs
erlarese: I moved this to a fixture file in the next commit.

src/providers/ibmcloud/mod.rs
erlarese: Added this check in the next commit.

src/providers/ibmcloud/mod.rs
erlarese: Changed in the next commit.

src/providers/ibmcloud/mod.rs
erlarese: Changed in the next commit.

erlarese (commented):
Hi @lucab, I believe I've made the requested changes.

One confusing thing is the linter is failing saying the import of `use std::fs` is unused, but it's pretty clearly used in that file to read the new fixture, and if I remove the import the code doesn't compile because `fs` can't be found.  I tried a few variations of prefixing the `read` call with `std::` and such, but I can't quite figure out why the linter thinks this import isn't used when the compiler seems to require it.  Any idea on how to fix this?

erlarese:
Ah, ignore the earlier comment about the import error, I realized the unit tests actually have a separate dependency declaration and moving the import down there seems to have fixed the issue.

lucab:
@erlarese thanks, code now looks overall fine to me. Can you please squash everything into a single commit, so that I'll do the final pass?

erlarese:
Hi @lucab -- the commits should be squashed back into one again.  Please let me know if anything else needs to be changed!

lucab (approved):

lucab:
LGTM, thanks for the patch!

PR #479, opened 2020-08-13
providers: move trait noops into MetadataProvider

darkmuggle:
Generally in all the providers there are trait(s) that are not
implemented or used by the provider; they all had various methods of
noop'ing. This change unifies the noop handling by implementing the
required fn in the trait definition itself.

Signed-off-by: Ben Howard <ben.howard@redhat.com>

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -195,11 +195,27 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
 }
 
 pub trait MetadataProvider {
-    fn attributes(&self) -> Result<HashMap<String, String>>;
-    fn hostname(&self) -> Result<Option<String>>;
-    fn ssh_keys(&self) -> Result<Vec<PublicKey>>;
-    fn networks(&self) -> Result<Vec<network::Interface>>;
-    fn boot_checkin(&self) -> Result<()>;
+    fn attributes(&self) -> Result<HashMap<String, String>> {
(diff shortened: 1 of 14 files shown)


src/providers/mod.rs
```diff
@@ -195,19 +195,38 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
...
 
 pub trait MetadataProvider {
-    fn attributes(&self) -> Result<HashMap<String, String>>;
-    fn hostname(&self) -> Result<Option<String>>;
-    fn ssh_keys(&self) -> Result<Vec<PublicKey>>;
-    fn networks(&self) -> Result<Vec<network::Interface>>;
-    fn boot_checkin(&self) -> Result<()>;
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        let attributes = maplit::hashmap! {};
+        Ok(attributes)
```
lucab: Nit: `Ok(HashMap::new())` should be enough here (it also avoids a macro-expansion pass).

lucab:
@darkmuggle nice cleanup, thanks!

src/providers/mod.rs
darkmuggle: Indeed.... my rust noob status is showing :) I had wondered about going with anonymous hashmap there. The tip is appreciated.

lucab (approved):

jlebon:
Nice cleanup!

PR #487, opened 2020-09-07
dracut: run hostname service on aliyun

### LINKED ISSUE #482: dracut: run hostname service on aliyun
AliCloud havn't DHCP for set hostname when firstboot, could you add condition for aliyun in afterburn-hostname.service ?
@lucab

robertvolkmann:
Unfortunately, Aliyun does not offer the hostname via DHCP. The recommended way to set the hostname without cloud-init is by a custom shell script, see [Customize Linux images](https://www.alibabacloud.com/help/doc-detail/51138.htm?spm=a2c63.p38356.879954.11.15f81541TyNJQ9#concept-nt5-4km-xdb)

Fixes: #482

### DIFF
--- a/dracut/30afterburn/afterburn-hostname.service
+++ b/dracut/30afterburn/afterburn-hostname.service
@@ -3,6 +3,7 @@ Description=Afterburn Hostname
 # These platforms do not provide the hostname via DHCP
 # options, thus it needs to be fetched from the metadata
 # and statically applied on first-boot.
+ConditionKernelCommandLine=|ignition.platform.id=aliyun
 ConditionKernelCommandLine=|ignition.platform.id=azure
 ConditionKernelCommandLine=|ignition.platform.id=digitalocean
 ConditionKernelCommandLine=|ignition.platform.id=exoscale


robertvolkmann:
@lucab Did you get a chance to look into it?

lucab:
@robertvolkmann not yet, it's in the same bucket as https://github.com/coreos/fedora-coreos-docs/pull/173. The company account I used initially when working on AliCloud is gone. I asked for access to another internal one we have, and I hope to be unblocked next week.

lucab:
@robertvolkmann thanks for your patience, I finally got my dev account re-enabled and cleared. I verified that this change is indeed correct, patch LGTM. Can you please rebase against latest master?

lucab:
Self-note: one minor caveat is that on `aliyun` the hostname is optional. If not specified, the node-id is supposed to be used transparently by the cloud platform. However the metadata endpoint seems to return some kind of maimed value, which isn't exactly equal to the node-id (but awkwardly similar):
```
$ curl http://100.100.100.200/latest/meta-data/instance-id
i-gw856mji2h8aix5w0pos

$ curl http://100.100.100.200/latest/meta-data/hostname    
iZgw856mji2h8aix5w0posZ
```

This isn't directly under our control, so I'm just noting down this quirk here. If AliCloud eventually manages to fix the wrong value in the `/hostname` endpoint, this quirk will go away on its own.

lucab (approved):

PR #488, opened 2020-09-08
[RFC] providers: add qemu provider for boot check-in

### LINKED ISSUE #458: qemu: implement boot-time checkin (via ovirt-guest-agent protocol)
Currently if FCOS fails in very early boot (e.g. in the bootloader or kernel, before switching to the initramfs), it's...hard to detect consistently.  For example, we have a test for Secure Boot, but it turns out that if the [kernel fails to verify](https://bugzilla.redhat.com/show_bug.cgi?id=1856501) then...coreos-assembler today hangs for a really long time.

We can scrape the serial console in *most* cases, but we configure the Live ISO not to log to a serial console...so we'd end up having instead to do something like [OpenQA](http://open.qa/documentation/) and do image recognition on the graphics console :cry: 

Now we debated this somewhat in
https://github.com/coreos/ignition-dracut/pull/170
and I argued strongly that the most important thing was to cover the "failure in initramfs" case, and we could support the "stream journal in general" by injecting Ignition.

In retrospect...I think I was wrong.  It would be extremely useful for us to stream the journal starting from the initramfs at least by default on qemu.

In particular, what we really want is some sort of message from the VM that it has *entered the initramfs*, but before we start processing Ignition.  If we're doing things like reprovisioning the rootfs, it becomes difficult to define a precise "timeout bound".  But I think we can e.g. reliably time out after something quite low (like 10 seconds) if we haven't seen the "entered initramfs" message.

So here's my proposal:

- Add a variant of https://github.com/coreos/ignition-dracut/pull/170 that streams the journal to a *differently named* virtio channel
- Update coreos-assembler to use it
- Eventually drop the old "only dump on failure" channel and code

lucab:
TODO:
 * [ ] figure out at which point in initramfs this should run 

---

This adds a new provider for the "qemu" platform.
In particular, this provider currently supports checking in to
the hypervisor in order to report a successful boot.
This is performed through a VirtIO port, re-using the same console
and protocol used by oVirt guest-agent. Though, this is intentionally
not aimed at providing a full guest-agent implementation.

Fixes https://github.com/coreos/afterburn/issues/458

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,6 +32,8 @@ dependencies = [
  "slog-scope",
  "slog-term",
  "tempfile",
+ "tokio",
+ "tokio-oga",
  "update-ssh-keys",
  "users",
  "vmw_backdoor",
@@ -39,9 +41,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.10"
+version = "0.7.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8716408b8bc624ed7f65d223ddb9ac2d044c0547b6fa4b0d554f3a9540496ada"
+checksum = "043164d8ba5c4c3035fec9bbee8647c0261d788f3474306f93bb65901cae0e86"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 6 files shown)


cgwalters:
There's a lot going on in tokio-oga, will try to give that a sanity check soon too.

>  figure out at which point in initramfs this should run

Wouldn't it be part of `afterburn-checkin.service`?

I assume you've tested this in an actual oVirt/RHV environment - is that correct?  Which version?

Do you have a sense of how hard it'd be to have the server side of this implemented in coreos-assembler/kola?  That would allow us to use it across the board in tests.

PR #493, opened 2020-09-28
sshkeys: activate service on OpenStack

lucab:
This adds triggering condition to SSH-keys service units on OpenStack.


dustymabe (approved):
LGTM

jlebon (approved):

PR #495, opened 2020-10-08
providers/vmware: allow avoiding iopl permission errors

lucab:
This introduces a fallback path for the backdoor-opening logic,
in order to allow Afterburn to keep going when changing
I/O access level is not allowed.
Privileged mode is in theory more reliable but `kernel_lockdown(7)`
may block it due to `iopl()` usage. Thus, we try that first and
fall back if kernel blocks it.

Ref: https://github.com/lucab/vmw_backdoor-rs/issues/6

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,9 +39,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.10"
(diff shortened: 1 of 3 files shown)


arithx (approved):
Code looks good to me. Looks like CI flaked.

PR #496, opened 2020-10-09
cargo: remove carets from semver requirements

lucab:
This removes all carets (`^`) from dependencies, resulting in
semver entries which are semantically equivalent from cargo
point of view.
This repository is starting to accumulate stale dependencies, as
dependabot is having troubles when trying to update them. I think
I'm seeing a pattern where the `^` may be the one thing confusing
the bot, and the changes in here try to verify such theory.

Ref: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio
Ref: https://github.com/dependabot/dependabot-core/issues/2532

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,42 +25,42 @@ path = "src/main.rs"
 lto = true
 
 [dependencies]
-base64 = "^0.12"
-byteorder = "^1.3"
-cfg-if = "^0.1"
-clap = "^2.33"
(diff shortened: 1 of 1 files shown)


jlebon (approved):

lucab:
Indeed, this unlocked https://github.com/coreos/afterburn/pull/499 and https://github.com/coreos/afterburn/pull/498 and https://github.com/coreos/afterburn/pull/497.

PR #505, opened 2020-10-19
systemd/firstboot-checkin: order before first-boot-complete.target

jlebon:
Otherwise, it's possible we'll never run if the system is rebooted after
`systemd-machine-id-commit.service` but before this unit.

This will be recommended by systemd starting from:
https://github.com/systemd/systemd/pull/16939

### DIFF
--- a/systemd/afterburn-firstboot-checkin.service
+++ b/systemd/afterburn-firstboot-checkin.service
@@ -2,6 +2,14 @@
 Description=Afterburn (Firstboot Check In)
 ConditionKernelCommandLine=|ignition.platform.id=packet
 ConditionFirstBoot=yes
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
LGTM

jlebon:
Added `Wants=first-boot-complete.target` as pointed out in https://github.com/jlebon/fedora-coreos-config/commit/8d610b8e0eb92bb43d71152e0caf87af4d0d6f81#commitcomment-43396028.

jlebon:
Let's hold this one. It may result in an ordering cycle like https://github.com/coreos/fedora-coreos-tracker/issues/660. At least let's wait until we have https://github.com/coreos/coreos-assembler/pull/1834.

Issue #509, opened 2020-10-26
dracut/hostname: shorten overlong hostname


lucab:
On a subset of platforms, the `afterburn-hostname.service` tries to retrieve the node hostname from a metadata field and forward any value from there into `/etc/hostname` as the machine static hostname.

According to POSIX.1 "Host names (not including the terminating null byte) are limited to `HOST_NAME_MAX` bytes".  On Linux, `HOST_NAME_MAX` is defined with the value `64`, see https://man7.org/linux/man-pages/man2/gethostname.2.html.

From prior experiences, we know that it's possible that either humans or platforms try to set up names longer than than (e.g. FQDN up to 255 bytes), resulting in much pain.

Afterburn should have some additional logic in `write_hostname()` to truncate overlong hostnames to the first dot or to `HOST_MAX_LEN`, whatever comes earlier.
Note that this will possibly mean a desync between the hostname file and the metadata attributes files.

I wrote similar logic (with tests) for systemd-networkd once (https://github.com/systemd/systemd/pull/7616) and it could be ported from there.

jamescassell:
I've run into some problematic applications that barf if the hostname of the system is not the FQDN rather than the short name... hopefully those are rare. (openshift-ansible for OKD 3.11 had this issue.)

cgwalters:
For Kubernetes it's required that the hostname be routable.  Which means that the "hostname shortening" logic *must* match whatever the cloud provider (GCP) will do.

cgwalters:
OK this came up again recently in https://github.com/openshift/machine-config-operator/pull/2401 and here's my proposal:

We ship `afterburn-transient-hostname.service` that is disabled by default, but can be enabled e.g. by the MCO.  We also move the special GCP hostname truncation into this project.

That way the MCO doesn't need to carry bash code that shleps the hostname to a file, etc.

cgwalters:
It turned out this bit us *badly* in https://bugzilla.redhat.com/show_bug.cgi?id=2008521#c40

Clearly, we should have gone this route from the start.

cgwalters:
@lucab Did you have any further thoughts on this?

It seems like we're all in agreement that afterburn should do this?

lucab:
Yes, I agree Afterburn should more strictly honor `HOST_NAME_MAX` if not doing so causes further bugs in downstream components. That may cause a regression for folks that were previously using overlong hostnames, but I think those should be considered buggy cases.

[...]

PR #512, opened 2020-11-03
Turn on Afterburn hostname support for GCP

nikita-dubrovskii:
On GCP the DHCP server send the FQDN as the shortname in DHCP.
So when the code that truncates the hostname in the kubelet runs,
NM gets ornery that it didn't change the hostname and flips it back to the long name

Signed-off-by: Nikita Dubrovskii <nikita@linux.ibm.com>

### DIFF
--- a/dracut/30afterburn/afterburn-hostname.service
+++ b/dracut/30afterburn/afterburn-hostname.service
@@ -7,6 +7,7 @@ ConditionKernelCommandLine=|ignition.platform.id=azure
 ConditionKernelCommandLine=|ignition.platform.id=digitalocean
 ConditionKernelCommandLine=|ignition.platform.id=exoscale
 ConditionKernelCommandLine=|ignition.platform.id=vultr
+ConditionKernelCommandLine=|ignition.platform.id=gcp
 
 # We order this service after sysroot has been mounted
 # but before ignition-files stage has run, so ignition can


bgilbert (approved):
Patch LGTM, but let's hold for a bit to allow further discussion in https://github.com/coreos/fedora-coreos-tracker/issues/649.

dustymabe:
Note that the systemd fallback hostname change is being reverted in Fedora 33 (https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-739956843). By the time the change is re-introduced there should be NetworkManager code that lands that should handle the hostname truncation in a configurable way such that it's not a problem for us any longer (https://github.com/coreos/fedora-coreos-tracker/issues/649#issuecomment-726884301).

sohankunkerkar:
> Patch LGTM, but let's hold for a bit to allow further discussion in [coreos/fedora-coreos-tracker#649](https://github.com/coreos/fedora-coreos-tracker/issues/649).

It seems like this [issue](https://github.com/coreos/fedora-coreos-tracker/issues/649) was closed, so does this unblock us from proceeding with this PR?

jlebon:
FCOS now still has (and will continue to have) `localhost` as the fallback hostname (see https://github.com/coreos/fedora-coreos-config/pull/868), so IIUC, we don't strictly need this anymore because the dispatcher script shipped in the MCO kick in. Moving forward, I think we need to figure out if to push the hostname truncation logic here or in NM (it looks like the NM MR for this was closed: https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/677).

cgwalters:
This looks related to https://github.com/coreos/afterburn/issues/509 ?

bgilbert:
From the discussion above, it appears to me that this isn't needed any longer, so I'll close.  Feel free to reopen if it's still needed.

PR #561, opened 2021-03-08
Add Azurestack Support

patrickdillon:
Updated as this is no longer WIP.

This PR:

- adds a bare-bones azurestack provider
- copies all of the logic from the azure provider
- adjusts the logic behind hostname fetching to query the identity metadata service (not the new instance metadata endpoint) 
- removes support for attributes (using the logic of azure resulted in incorrect IP addresses)

I tested this and it pulled unique hostnames for a bootstrap node and 3 masters. The SSH keys also appear in the login screen. 

Note: because this uses the identity metadata service, this will work in any azure stack environment.

Further todo on this PR:

- [X] remove changes to dracut/30afterburn/afterburn-hostname.service
- [X] Go through all of the logic from the azure provider and determine whether it is necessary; strip it out if we don't need it
- [X] Determine whether we want to query both the compute endpoint and the identity endpoint, falling back to one if the other doesn't work
- [X] Clean up commits (better messages, squash FIXUPs)

Regarding the compute endpoint: 
From a running VM (in the lab environment) I am able to query it:
```shell
[core@ash-bootstrap ~]$ curl -H Metadata:true --noproxy "*" http://169.254.169.254/metadata/instance?api-version=2019-03-11 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1484  100  1484    0     0   9160      0 --:--:-- --:--:-- --:--:--  9217
{
  "compute": {
    "azEnvironment": "AzureStack",
```
But when I hit that same endpoint during afterburn I get an error. Because I was able to get the vm name from the identity endpoint, I prioritized that. as future work we could add more support for the instance metadata service.


patrickdillon:
/hold
WIP
/cc @jlebon

dracut/30afterburn/afterburn-hostname.service
```diff
@@ -1,14 +1,9 @@
...
 # These platforms do not provide the hostname via DHCP
 # options, thus it needs to be fetched from the metadata
 # and statically applied on first-boot.
-ConditionKernelCommandLine=|ignition.platform.id=aliyun
-ConditionKernelCommandLine=|ignition.platform.id=azure
-ConditionKernelCommandLine=|ignition.platform.id=digitalocean
-ConditionKernelCommandLine=|ignition.platform.id=exoscale
-ConditionKernelCommandLine=|ignition.platform.id=ibmcloud
-ConditionKernelCommandLine=|ignition.platform.id=vultr
+ConditionKernelCommandLine=|ignition.platform.id=azurestack
```
jlebon, line 8: > there is a messy change in afterburn-hostname.service that breaks other platforms, but I think I should be able to just add them back in....

Yeah, there shouldn't be any issue there at all with just adding this line only to the original service unit.

src/providers/azurestack/crypto/mod.rs
```diff
@@ -0,0 +1,90 @@
+// Copyright 2017 CoreOS, Inc.
```
jlebon: `2021`

This came up in discussion in another project as well. For this project at least, we should probably be able to drop all these blocks and just rely on the `Cargo.toml` license field. (To be clear, I mean this as a follow-up, not something you need to do here.)

jlebon:
> But when I hit that same endpoint from ignition, I get a 400. Any ideas?

You mean when running Afterburn from the initrd, right? Hmm, not sure offhand. Can you paste the logs?

patrickdillon:
> You mean when running Afterburn from the initrd, right?

Yes, afterburn! I keep mistakenly saying "from ignition". 

> Hmm, not sure offhand. Can you paste the logs?

Looks like I did not save the logs. I remember there was minimal info in the error, with a 400 status code. I will run again in order to capture. Let me know if you have any suggestions for getting a more detailed debug output.

jlebon:
> I remember there was minimal info in the error, with a 400 status code.

That's odd, I don't know why it would give a 400 in the initrd, but work in the real root. Hmm, wonder if it wants the node to check in first?

If you have interactive access to the console BTW, you can try booting with `rd.break` (which itself might not be possible), and then run commands in the initrd shell. (E.g. I'd start with doing the same `curl` command to make sure that it's not a difference between how `curl` and `afterburn` do the request, then try doing a checkin first, etc...)

src/providers/azurestack/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+        // convert that to the OpenSSH public key format
+        let ssh_pubkey = crypto::p12_to_ssh_pubkey(&p12)
+            .context("failed to convert pkcs12 blob to ssh pubkey")?;
+
+        Ok(ssh_pubkey)
+    }
+
+    fn fetch_hostname(&self) -> Result<Option<String>> {
+        let instance_metadata = AzureStack::fetch_identity();
+        Ok(Some(instance_metadata.unwrap().vm_name))
```
jlebon: The linter is indirectly saying to use `Err` instead of `unwrap()`, e.g.:

```suggestion
        let instance_metadata = AzureStack::fetch_identity()
        	.context("fetching instance identity")?;
        Ok(Some(instance_metadata.vm_name))
```

patrickdillon: Is there a way I should change `fetch_identity`: https://github.com/coreos/afterburn/pull/561/files#diff-34cbce01c9aa769653cb8f06a598911f28d90de5710e9b602560660669d702d1R172-R183 where I already have `context`?

jlebon: The `fetch_identity` function looks good. The `context` stuff is like a human-friendly backtrace where you can add more error-prefixing as the error bubbles up the stack to make it easier to tell where the error came from. It's up to you when to add a `.context(...)?` vs just `?`. If you think e.g. it's repetitive with the context already added by the callee, you can omit it in the caller. In this case for example, yeah it's probably superfluous given that my suggestion is more or less repeating the context provided by `fetch_identity`, and our caller ultimately adds the context that the failure happened while fetching the hostname:

```rust
        // write hostname if configured to do so
        self.hostname_file
            .map_or(Ok(()), |x| metadata.write_hostname(x))
            .context("writing hostname")?;
```

So you can probably just swap out `.context("fetching instance identity")?` with just `?` in my suggestion above.

patrickdillon:
@jlebon All work on this is done and tested. Can you give it a more thorough review or suggest how to get more eyes on it?

I have logs from testing master and bootstrap nodes if that will help. 

Not sure why continuous-integration/jenkins/pr-merge  test is failing. Perhaps this is a test ifnra problem? I cannot replicate locally. Also can you tip me on how to retrigger this test? I can't figure out a way. Do I have to push a new empty commit?

patrickdillon:
The merge test is failing with:
```
[2021-03-25T20:16:35.302Z] + make
[2021-03-25T20:16:35.302Z] sed -e 's,@DEFAULT_INSTANCE@,'core',' < systemd/afterburn-sshkeys@.service.in > systemd/afterburn-sshkeys@.service.tmp && mv systemd/afterburn-sshkeys@.service.tmp systemd/afterburn-sshkeys@.service
[2021-03-25T20:16:35.302Z] cargo build --release
[2021-03-25T20:16:35.302Z] make: cargo: No such file or directory
[2021-03-25T20:16:35.302Z] make: *** [Makefile:23: all] Error 127

script returned exit code 2
```
This looks like a problem running `cargo` so a problem with the test itself.

jlebon:
Yeah, CI was hitting issues yesterday. It should be fine now, so I've restarted another run.

jlebon:
So in a first pass, there seems to be a lot of duplication between `azure` and `azurestack`. From what I can tell, all the crypto and goalstate stuff is exactly the same. We should try to dedupe this. Hmm, how about re-arranging things like this:

```
src/providers/azure/mod.rs
src/providers/azure/goalstate.rs
src/providers/azure/crypto/...
src/providers/azure/azure/mod.rs
src/providers/azure/azure/mock_test.rs
src/providers/azure/azurestack/mod.rs
src/providers/azure/azurestack/mock_test.rs
```

So then the top-level `providers::azure` module would just re-export the `azure` and `azurestack` (using `pub mod`) so that they could be used in the top-level `match` in `metadata.rs`. WDYT?

src/providers/azurestack/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+            .dispatch_post()?;
+        Ok(())
+    }
+}
+
+impl MetadataProvider for AzureStack {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        Ok(HashMap::new())
+    }
+
```
jlebon: That's the default implementation, so you can drop it entirely.

src/providers/azurestack/mod.rs
```diff
@@ -0,0 +1,369 @@
...
+    endpoint: IpAddr,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+struct InstanceMetadata {
+    #[serde(rename = "vmName", default)]
+    pub vm_name: String,
+    #[serde(rename = "subscriptionId", default)]
+    pub subscription_id: String,
+}
```
jlebon: Should be able to simplify to

```suggestion
#[derive(Debug, Deserialize, Clone)]
#[serde(rename_all = "camelCase")]
struct InstanceMetadata {
    pub vm_name: String,
    pub subscription_id: String,
}
```

Do we actually need the `default` here? I.e. do we always expect those fields to exist, or do we want to support them not existing? In case of the former, we should just drop `default`. In case of the latter, it would be cleaner to use an `Option<String>` instead.

jlebon:
OK, updated this based on https://github.com/coreos/afterburn/pull/561#issuecomment-808360832. The only change was that I made the top-level module `microsoft` instead of `azure` to make it less confusing (because otherwise we'd have `azure::azure`).

Still need to test this.

jlebon:
Now with authorship actually preserved! :)

lucab:
@jlebon there seems to be still quite a bit of overlap between `azure` and `azurestack` code (at least in constants/macros/helpers/tests). Is the plan to keep polishing here, or is that for later?

ashcrow:
Other than possible overlap of constants/etc.. is there anything else holding this PR up?

jlebon:
> @jlebon there seems to be still quite a bit of overlap between `azure` and `azurestack` code (at least in constants/macros/helpers/tests). Is the plan to keep polishing here, or is that for later?

> Other than possible overlap of constants/etc.. is there anything else holding this PR up?

To answer both of you at the same time: this PR is being tested in ASH, which is painful to do. If testing comes back green, my suggestion is to get this in as is and tweak things in follow-ups.

patrickdillon:
I tested this in Azure Stack and everything looked good. I did not test ssh key fetching which I am not familiar with. But happy to help further.

jlebon (approved):
I've also finally been able to successfully test this in ASH, including SSH key fetching (which matters more for FCOS):

```
[root@jlebon-vm-bootstrap ~]# journalctl -u afterburn-sshkeys@core.service
-- Logs begin at Thu 2021-04-08 01:43:58 UTC, end at Thu 2021-04-08 03:27:48 UTC. --
Apr 08 02:00:22 jlebon-vm-bootstrap systemd[1]: Starting Afterburn (SSH Keys)...
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: Apr 08 02:00:51.046 WARN Failed to get fabric address from DHCP: maximum number of retries (60) reached
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: Apr 08 02:00:51.046 INFO Using fallback address
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: Apr 08 02:00:51.055 INFO Fetching http://168.63.129.16/?comp=versions: Attempt #1
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: Apr 08 02:00:51.062 INFO Fetch successful
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: Apr 08 02:00:51.062 INFO Fetching http://168.63.129.16/machine/?comp=goalstate: Attempt #1
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: Apr 08 02:00:51.219 INFO Fetch successful
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: Apr 08 02:00:51.270 INFO Fetching http://168.63.129.16/machine/70f676d8-fb84-4921-9716-96cc629c05a7/...
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: Apr 08 02:00:51.577 INFO Fetch successful
Apr 08 02:00:51 jlebon-vm-bootstrap afterburn[1929]: wrote ssh authorized keys file for user: core
Apr 08 02:00:51 jlebon-vm-bootstrap systemd[1]: afterburn-sshkeys@core.service: Succeeded.
Apr 08 02:00:51 jlebon-vm-bootstrap systemd[1]: Started Afterburn (SSH Keys).
[root@jlebon-vm-bootstrap ~]# cat /home/core/.ssh/authorized_keys.d/afterburn
ssh-rsa AAA...
```

So my vote is to get this in and we can do more cleanups afterwards. @lucab Would appreciate a stamp to sanity-check there isn't anything majorly wrong.

lucab (approved):

jlebon:
One thing worth noting is that in the base case, ASH support also requires https://github.com/coreos/fedora-coreos-config/pull/935 so it has networking in the initrd. For OKD/OCP it works without it, because we use pointer Ignition configs and so Ignition itself needs networking and requests it.

PR #567, opened 2021-03-16
Switch from error-chain to anyhow

bgilbert:
`error-chain` is no longer actively maintained.  Switch to a more modern library.  This does lose us backtrace support until `std::backtrace` is stabilized.

Switch to returning a `Result` directly from `main()`, since `anyhow` has reasonable default error rendering.


cgwalters (approved):
Nice!  LGTM.

PR #592, opened 2021-06-14
Add PowerVS provider

Prashanth684:
As part of supporting Power Virtual server as a platform (https://github.com/coreos/fedora-coreos-tracker/issues/817), Afterburn will be used to set the hostname attributes and possibly the ssh-keys.

This provider is very similar to the IBMCloud classic provider with the exception of the filesystem for the config drive which is `iso9660` similar to Openstack. The contents of the config drive are usually the meta_data.json and the network_data.json. This provider includes support for:
- hostname
- ssh keys


src/metadata.rs
```diff
@@ -58,6 +59,7 @@ pub fn fetch_metadata(provider: &str) -> Result<Box<dyn providers::MetadataProvi
         "ibmcloud" => box_result!(IBMGen2Provider::try_new()?),
         // IBM Cloud - Classic infrastructure.
         "ibmcloud-classic" => box_result!(IBMClassicProvider::try_new()?),
+        "powervs" => box_result!(PowerVSProvider::try_new()?),
```
bgilbert: Alphabetical order, please.

src/providers/powervs/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    /// The public key is stored as key:value pair in openstack/latest/meta_data.json file
+    fn fetch_publickeys(&self) -> Result<Vec<PublicKey>> {
+        let filename = self.metadata_dir().join("meta_data.json");
+        let file = File::open(&filename)
+            .with_context(|| format!("failed to open file '{:?}'", filename))?;
+
+        let bufrd = BufReader::new(file);
+        let metadata: MetaDataJSON = Self::parse_metadata(bufrd)
+            .with_context(|| format!("failed to parse file '{:?}'", filename))?;
```
bgilbert: This can be replaced by a call to `read_metadata()`.

src/providers/powervs/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    #[test]
+    fn test_powervs_parse_metadata_json() {
+        let fixture = File::open("./tests/fixtures/powervs/meta_data.json").unwrap();
+        let bufrd = BufReader::new(fixture);
+        let parsed = PowerVSProvider::parse_metadata(bufrd).unwrap();
+
+        assert!(!parsed.instance_id.is_empty());
+        assert!(!parsed.local_hostname.is_empty());
+        assert!(!parsed.public_keys.is_none());
```
bgilbert, line 225: Can we have an explicit test for fetching public keys also?

src/providers/powervs/mod.rs
Prashanth684: ah..yes ..done..

src/metadata.rs
Prashanth684: done.

src/providers/powervs/mod.rs
Prashanth684: added.

src/providers/powervs/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+    fn read_metadata(&self) -> Result<MetaDataJSON> {
+        let filename = self.metadata_dir().join("meta_data.json");
+        let file = File::open(&filename)
+            .with_context(|| format!("failed to open file '{:?}'", filename))?;
+        let bufrd = BufReader::new(file);
+        Self::parse_metadata(bufrd)
+    }
+
+    /// The public key is stored as key:value pair in openstack/latest/meta_data.json file
+    fn fetch_publickeys(&self) -> Result<Vec<PublicKey>> {
```
bgilbert: Hmm, actually, I misled you.  It would be better if this was `fn public_keys(metadata: MetaDataJSON)`, so the tests could verify the output of this function against a fixture.

Prashanth684: yes..makes sense...changed the test

bgilbert (approved):
LGTM.  @lucab, do you have any concerns?

lucab:
I have some slight confusion related to this:

> This provider includes support for:
> * network interfaces

But at https://github.com/coreos/fedora-coreos-tracker/issues/817 it says:

> How can the OS retrieve network configuration? Is DHCP sufficient, or is there some other network-accessible metadata service?
> DHCP. But let me also confirm if any additional config is required.

How does networking on this platform actually work?

Prashanth684:
> I have some slight confusion related to this:
> 
> > This provider includes support for:
> > 
> > * network interfaces
> 
> But at [coreos/fedora-coreos-tracker#817](https://github.com/coreos/fedora-coreos-tracker/issues/817) it says:
> 
> > How can the OS retrieve network configuration? Is DHCP sufficient, or is there some other network-accessible metadata service?
> > DHCP. But let me also confirm if any additional config is required.
> 
> How does networking on this platform actually work?

So for a little context on this - Power VS as of today has only static networking support. As part of developing support for Openshift [1] , the IBM team is working on a DHCP solution. The idea is that when productizing the platform (especially for IPI deployments), DHCP is a hard requirement and is needed. But in the meantime, the team could also use static networking to test other bits. This might not be used in the final product but PowerVS does support configuring networks through the config drive data.

[1] https://github.com/openshift/enhancements/pull/736

Prashanth684:
> > I have some slight confusion related to this:
> > > This provider includes support for:
> > > 
> > > * network interfaces
> > 
> > 
> > But at [coreos/fedora-coreos-tracker#817](https://github.com/coreos/fedora-coreos-tracker/issues/817) it says:
> > > How can the OS retrieve network configuration? Is DHCP sufficient, or is there some other network-accessible metadata service?
> > > DHCP. But let me also confirm if any additional config is required.
> > 
> > 
> > How does networking on this platform actually work?
> 
> So for a little context on this - Power VS as of today has only static networking support. As part of developing support for Openshift [1] , the IBM team is working on a DHCP solution. The idea is that when productizing the platform (especially for IPI deployments), DHCP is a hard requirement and is needed. But in the meantime, the team could also use static networking to test other bits. This might not be used in the final product but PowerVS does support configuring networks through the config drive data.
> 
> [1] [openshift/enhancements#736](https://github.com/openshift/enhancements/pull/736)

@lucab does this provide sufficient clarity ?

bgilbert:
> But in the meantime, the team could also use static networking to test other bits. This might not be used in the final product but PowerVS does support configuring networks through the config drive data.

I think it's okay to hook up the network metadata inside Afterburn, but it won't actually help on RHCOS.  Afterburn only knows how to configure the network via systemd-networkd, not via NetworkManager, so RHCOS doesn't use this piece of Afterburn functionality.

Prashanth684:
> > But in the meantime, the team could also use static networking to test other bits. This might not be used in the final product but PowerVS does support configuring networks through the config drive data.
> 
> I think it's okay to hook up the network metadata inside Afterburn, but it won't actually help on RHCOS. Afterburn only knows how to configure the network via systemd-networkd, not via NetworkManager, so RHCOS doesn't use this piece of Afterburn functionality.

Ok. On further discussions with the IBM team - the plan is to read the data from `network_data.json` and convert the data to network kargs which will then be read by dracut-cmdline. this will be used while a plan will be developed for DHCP in the long term. does that sound like a reasonable approach ?

bgilbert:
Is the idea still to use static networking during development, but to ship with DHCP support?  I'm not sure it makes sense to add kargs support to Afterburn solely for a temporary use case not intended for production deployments.  One hacky option is to write out a networkd config, and then have a shell script in the initrd convert it to kargs.

Prashanth684:
> Is the idea still to use static networking during development, but to ship with DHCP support? I'm not sure it makes sense to add kargs support to Afterburn solely for a temporary use case not intended for production deployments. One hacky option is to write out a networkd config, and then have a shell script in the initrd convert it to kargs.

Ok. so i've dropped the static networking support as the plan is to go with DHCP. if things change later , we can always look to adding static support at a later time.

Prashanth684:
@bgilbert  ok to merge?

bgilbert (approved):
LGTM

Prashanth684:
> LGTM

thanks! can this be merged please?

PR #628, opened 2021-10-11
docs: Do not convert -- & --- to en/em-dash

travier:
'--' is frequently used for command line options and was thus
incorrectly rendered as a special en-dash symbol.


bgilbert (approved):
Thanks!

PR #631, opened 2021-10-21
providers: only log we wrote SSH keys when we actually did

### LINKED ISSUE #632: afterburn writes authorized keys even when it gets a 404 from openstack metadata service
# Bug #

afterburn outputs an extra log message (and possibly writes an authorized keys file) when it shouldn't. It gets a 404 which means it shouldn't write anything out

Screenshot 2021-10-21 at 16-30-17 kola-b56a4bb3-54580fa4af(7c5ba524-c3e2-49f1-816e-a89c3369eec2)

FCOS 35.20211003.1.0 

## Afterburn Version ##

afterburn-5.1.0-1.fc35.x86_64

## Environment ##

OpenStack on Vexxhost

## Reproduction Steps ##

It's a bug in vexxhost infra we think where the metadata service isn't ready. See https://github.com/coreos/fedora-coreos-tracker/issues/994#issuecomment-948978150

jlebon:
We hit an instance in VexxHost where the metadata service was returning
404 and yet Afterburn was still saying that it wrote an SSH key for the
`core` user.

Closes: #632

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -125,6 +125,11 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
                 e.error
             })
             .with_context(|| format!("failed to persist file {:?}", file_path.display()))?;
+
+        // emit journal entry
+        let username = user.name().to_string_lossy();
+        let path = file_path.to_string_lossy();
+        write_ssh_key_journal_entry(logging::Priority::Info, &username, &path);
     } else {
         // delete the file
         match fs::remove_file(&file_path) {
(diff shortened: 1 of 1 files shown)


jlebon:
(Only compile-tested and noticed this by inspection. Didn't test.)

bgilbert (approved):
LGTM as far as it goes.  In the `else` case where we delete an existing SSH fragment file, should we log that fact as well?

jlebon:
Tested working on AWS with and without SSH key pair.

dustymabe:
This most likely fixed https://github.com/coreos/coreos-assembler/issues/1835

PR #668, opened 2022-01-06
Explicitly log the hostname we write

cgwalters:
This would have greatly helped debugging https://bugzilla.redhat.com/show_bug.cgi?id=2008521#c40
because it was not immediately obvious from the logs that
afterburn is just writing a full, untruncated hostname to `/sysroot/etc/hostname`.

We see in the logs that the service started/stopped successfully,
but in this case it happens that systemd truncates it, which obscured
the value.

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -219,7 +219,7 @@ pub trait MetadataProvider {
                     hostname, hostname_file
                 )
             })?;
-            slog_scope::info!("Wrote hostname {} to {}", hostname, hostname_file_path);
+            slog_scope::info!("wrote hostname {} to {}", hostname, hostname_file_path);
         }
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


src/providers/mod.rs
```diff
@@ -211,18 +211,17 @@ pub trait MetadataProvider {
...
-            None => Ok(()),
+        if let Some(hostname) = self.hostname()? {
+            let mut hostname_file = create_file(&hostname_file_path)?;
+            writeln!(&mut hostname_file, "{}", hostname).with_context(|| {
+                format!(
+                    "failed to write hostname {:?} to file {:?}",
+                    hostname, hostname_file
+                )
+            })?;
+            slog_scope::info!("Wrote hostname {} to {}", hostname, hostname_file_path);
```
bgilbert: nit: looks like we generally don't capitalize log messages.

bgilbert (approved):
LGTM!

src/providers/mod.rs
cgwalters: Heh, you're right except I happened to find in my code search the one log message that *was* capitalized :smile: - then I used that as a base for this.

Pushed an additional commit to make it all consistent.

bgilbert (approved):

PR #673, opened 2022-01-11
providers: limit hostname file output value to `HOST_NAME_MAX` bytes

### LINKED ISSUE #509: dracut/hostname: shorten overlong hostname

bgilbert:
Some providers can return hostnames larger than the OS's maximum size (64 bytes on Linux), such as an FQDN of up to 255 bytes.  When writing such hostnames to the file specified by `--hostname`, truncate them to the first dot, and additionally to the maximum length if necessary.

Fixes https://github.com/coreos/afterburn/issues/509.

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -41,9 +41,11 @@ pub mod vultr;
 use crate::network;
 use anyhow::{anyhow, Context, Result};
 use libsystemd::logging;
+use nix::unistd;
 use openssh_keys::PublicKey;
 use slog_scope::warn;
 use std::collections::HashMap;
+use std::convert::TryInto;
 use std::fs::{self, File};
 use std::io::prelude::*;
 use std::path::Path;
@@ -152,6 +154,16 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
     Ok(())
 }
 
+fn max_hostname_len() -> Result<Option<usize>> {
+    unistd::sysconf(unistd::SysconfVar::HOST_NAME_MAX)
+        .context("querying maximum hostname length")?
+        .map(|l| {
+            l.try_into()
+                .context("overflow querying maximum hostname length")
+        })
+        .transpose()
+}
+
 pub trait MetadataProvider {
     fn attributes(&self) -> Result<HashMap<String, String>> {
         Ok(HashMap::new())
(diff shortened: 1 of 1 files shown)


src/providers/mod.rs
```diff
@@ -254,3 +284,67 @@ pub trait MetadataProvider {
...
+        provider
+            .write_hostname(temp.path().to_str().unwrap().into())
+            .unwrap();
+        let mut ret = String::new();
+        temp.read_to_string(&mut ret).unwrap();
+        ret.trim_end().into()
+    }
+
+    #[test]
+    fn test_hostname_truncation() {
```
cgwalters, line 314: Maybe add the test case from https://bugzilla.redhat.com/show_bug.cgi?id=2008521 here i.e.
`ci-op-dgnrb143-44266-42t6h-worker-c-ldlk9.com.openshift-gce-devel-ci.internal`
explicitly?

src/providers/mod.rs
```diff
@@ -152,6 +154,16 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
     Ok(())
 }
 
+fn max_hostname_len() -> Result<Option<usize>> {
+    unistd::sysconf(unistd::SysconfVar::HOST_NAME_MAX)
```
cgwalters, line 158: This is all fine as is, but at least systemd does it via compile-time and not runtime constant.  OTOH, I guess querying dynamically is perhaps cleaner.

src/providers/mod.rs
```diff
@@ -152,6 +154,16 @@ fn write_ssh_keys(user: User, ssh_keys: Vec<PublicKey>) -> Result<()> {
...
+        .context("querying maximum hostname length")?
+        .map(|l| {
+            l.try_into()
+                .context("overflow querying maximum hostname length")
+        })
+        .transpose()
```
cgwalters, line 164: I love `.transpose()`

cgwalters (approved):
Nice!

src/providers/mod.rs
bgilbert: Yeah, I'd prefer to use a compile-time constant.  Any idea where I can get one?  Neither `nix` nor `libc` have it.

src/providers/mod.rs
bgilbert: It's possibly too paranoid, but I'm using all synthetic test cases because I didn't mock the value of `HOST_NAME_MAX` and wanted to be robust to other values.

src/providers/mod.rs
cgwalters: I think for libc it's done manually, see e.g. https://github.com/rust-lang/libc/pull/2580
which is a random one I found.  I guess no one else needed it yet.

bgilbert: I did some digging.  At least FreeBSD and Mac OS X don't define `HOST_NAME_MAX`, which is allowed by POSIX if the value is at least 255.  So I think there's not much point in defining it in `libc`, since portable code couldn't assume the constant is available.

systemd doesn't have to be portable, of course.  We could hardcode a value here (or non-portably in `libc`), but it's ambiguous whether Afterburn supports non-Linux, so... sticking with the current approach for now.

PR #704, opened 2022-02-15
Add COPR integration Makefile and add CI test for RPM building

jlebon:
```
commit 50831ebcaafb8162222f57b75411a1f92fd9a036
Date:   Tue Feb 15 13:41:55 2022 -0500

    Add COPR integration Makefile

    I'd like to enable auto-builds of this repo to
    https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/ so it could
    eventually feed into
    https://github.com/coreos/fedora-coreos-tracker/issues/910.
```
```
commit 338c4db6bb80fd81a8e10e1b6ae39552f2d12c40
Date:   Tue Feb 15 13:45:30 2022 -0500

    workflows: add "RPM build" test

    Verify that we can build using purely Fedora packages. This will catch
    earlier any Rust packages we need to update or add in Fedora rather than
    find out at release time.

    It also ensures that the new continuous COPR builds will not break since
    we don't vendor there either.
```

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -1,0 +1,34 @@
+---
+name: RPM build
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+permissions:
+  contents: read
+
+jobs:
+  test-rpm-build:
+    name: "RPM build"
+    runs-on: ubuntu-latest
+    container:
+      image: registry.fedoraproject.org/fedora:latest
+      options: --privileged
+    steps:
+      # need to install git before checkout to get a git repo
+      - name: Install packages
+        run: dnf install -y git make mock
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Build RPM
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -98,3 +97,26 @@ jobs:
         run: cargo build
       - name: cargo test
         run: cargo test
+  test-rpm-build:
```
bgilbert: I'd lean toward putting this in a separate workflow, since it's unrelated to the other jobs in this file.

bgilbert (approved):
We should require the job in branch protection settings after this merges.

jlebon:
Updated!

.github/workflows/rpm.yml
```diff
@@ -0,0 +1,34 @@
+---
+name: Rust
```
bgilbert: Copypaste error.

bgilbert (approved):

Issue #705, opened 2022-02-15
Test `providers::aws::mock_tests::test_aws_basic` is flaky


jlebon:
Just hit this in a COPR build:

```
failures:

---- providers::aws::mock_tests::test_aws_basic stdout ----
thread 'providers::aws::mock_tests::test_aws_basic' panicked at 'called `Result::unwrap()` on an `Err` value: maximum number of retries (0) reached

Caused by:
    failed to fetch: 503 Service Unavailable', src/providers/aws/mock_tests.rs:22:39
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    providers::aws::mock_tests::test_aws_basic

test result: FAILED. 59 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.61s
```

It passed on the next try.

Permalink to the referenced line: https://github.com/coreos/afterburn/blob/7314651de320467a701172d8f8489926a386e188/src/providers/aws/mock_tests.rs#L22

@lucab Have you seen these mocked calls flake before?

lucab:
I have never observed this one before. It sounds like a race in `mockito` when deregistering mocks, however I did a quick reading through its code and didn't spot any obvious issues.
This may be related to the fact that we have multiple conflicting mocks and we rely on shadowing them under the name same. Moving to explicit non-overlapping scope may help here.

I've opened https://github.com/coreos/afterburn/pull/706, which also contains a general lockfile refresh to make sure we are picking up all the bugfixes in transitive dependencies.

jlebon:
We haven't hit this again so far, so let's optimistically close as fixed by #706 for now.

PR #706, opened 2022-02-16
tests: explicitly scope conflicting mocks

lucab:
This moves mocks with conflicting conditions to their own scopes,
without overlapping or implicit shadowing.
It also contains a general lockfile refresh.
Hopefully it should help avoiding test flakes observed on COPR.

Ref: https://github.com/coreos/afterburn/issues/705


jlebon (approved):
Thanks!

PR #721, opened 2022-03-08
github/release-checklist: Remove Windows binaries from vendored sources

travier:
Workaround to remove Windows specific binaries that we don't need from the vendored sources.
    
See: https://github.com/rust-lang/cargo/issues/7058


miabbott (approved):
LGTM

jlebon (approved):

bgilbert:
@travier Could you update coreos-installer and Zincati also?

PR #727, opened 2022-03-27
providers/aws: add `AWS_IPV6` attribute

### LINKED ISSUE #726: Support IPv6 on AWS
# Feature Request #

Currently afterburn only support IPv4 metadata, if the platform is AWS.

https://github.com/coreos/afterburn/blob/6e3f5452437cfcafc8c7abd6594b4e879a8bbc8d/src/providers/aws/mod.rs#L163-L164

Since [AWS announced IPv6 only subnets](https://aws.amazon.com/de/blogs/networking-and-content-delivery/introducing-ipv6-only-subnets-and-ec2-instances/) 

The IPv6 address can recived through meta-data.

```
http://169.254.169.254/latest/meta-data/ipv6
```
See also 

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html

> The IPv6 address of the instance. In cases where multiple network interfaces are present, this refers to the eth0 device (the device for which the device number is 0) network interface and the first IPv6 address assigned. If no IPv6 address exists on network interface[0], this item is not set and results in an HTTP 404 response.

## Environment ##

AWS

## Desired Feature ##

Have `COREOS_EC2_IPV6` available in ignition configs.

## Other Information ##

The IPv6 meta-data is available in metadata version `2021-01-03`.
Consider to bump the default version
https://github.com/coreos/afterburn/blob/6e3f5452437cfcafc8c7abd6594b4e879a8bbc8d/src/providers/aws/mod.rs#L82
or use latest.

bgilbert:
The AWS provider exports `IPV4_LOCAL` and `IPV4_PUBLIC` attributes.  EC2 doesn't make the same distinction for IPv6, and just assigns a single address to the primary interface.  Add it under the attribute name `IPV6`.

Bump the metadata version to 2021-01-03 to enable the new property.

Fixes https://github.com/coreos/afterburn/issues/726.


src/providers/aws/mod.rs
```diff
@@ -79,7 +79,7 @@ impl AwsProvider {
 
     #[cfg(not(test))]
     fn endpoint_for(key: &str, use_latest: bool) -> String {
-        const URL: &str = "http://169.254.169.254/2019-10-01";
+        const URL: &str = "http://169.254.169.254/2021-01-03";
```
jlebon, line 83: Minor: this would be nicer as a separate commit.

jlebon: Hmm, I can't seem to find any documentation about what actually changes in each version (the only section on the topic I found was https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#instancedata-meta-data-retrieval-examples). Was it just trial-and-error that got you to figure out that this was the earliest version that had `meta-data/ipv6`?

jlebon (approved):

src/providers/aws/mod.rs
bgilbert: Those docs are [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html).

Moved the version bump to a prep commit.

jlebon (approved):

jkroepke:
Hi, can I help here unblock the merge of the PR?

Thanks

bgilbert:
@lucab Any concerns about the naming of the attribute?

jkroepke:
Personally recommend the name, since its a 1:1 mapping of the IMDS endpoint

lucab:
I agree, let's stick to the IMDS name even if it looks overly generic. LGTM!

lucab (approved):

PR #728, opened 2022-03-28
network: Encode information for systemd-networkd-wait-online

pothos:
The network-online.target can use systemd-networkd-wait-online.service
to wait for all interfaces to come up. It will fail if the interfaces
didn't came up but sometimes it is actually ok for some interfaces to
be down because they are unused or they are just one of two parts of a
bond. We should encode when interfaces will never come up and when it
is acceptable to have interfaces in a degraded state and which.
Extend the network logic to handle this additional configuration. For
Packet we expect the metadata to specify all interfaces, and any other
physical NICs can be set to "unmanaged" so that we don't wait for them.
Introduce "Path" matching in the networkd unit file for that.
We also allow bonds to operate with only one working link, and we don't
wait for all bonded interfaces to be configured.
This is a port of https://github.com/flatcar-linux/afterburn/pull/10
to afterburn's main branch.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -72,6 +72,8 @@ pub struct Interface {
     pub name: Option<String>,
     /// Interface MAC address.
     pub mac_address: Option<MacAddr>,
+    /// Path as identifier
+    pub path: Option<String>,
     /// Relative priority for interface configuration.
     pub priority: u8,
     pub nameservers: Vec<IpAddr>,
@@ -79,6 +81,8 @@ pub struct Interface {
     pub routes: Vec<NetworkRoute>,
     pub bond: Option<String>,
     pub unmanaged: bool,
+    /// Optional requirement setting instead of the default
+    pub required_for_online: Option<String>,
 }
 
 /// A virtual network interface.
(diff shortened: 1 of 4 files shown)


src/network.rs
```diff
@@ -126,10 +130,11 @@ impl NetDevKind {
...
     pub fn sd_network_unit_name(&self) -> Result<String> {
-        let iface_name = match (&self.name, &self.mac_address) {
-            (Some(ref name), _) => name.clone(),
-            (None, Some(ref addr)) => addr.to_string(),
-            (None, None) => bail!("network interface without name nor MAC address"),
+        let iface_name = match (&self.name, &self.mac_address, &self.path) {
+            (Some(ref name), _, _) => name.clone(),
+            (None, Some(ref addr), _) => addr.to_string(),
+            (None, None, Some(ref path)) => path.to_string(),
+            (None, None, None) => bail!("network interface without name nor MAC address"),
```
bgilbert:
```suggestion
            (None, None, None) => bail!("network interface without name, MAC address, or path"),
```

bgilbert (approved):
LGTM generally, just one nit.

bgilbert (approved):
Thanks for upstreaming this!

PR #730, opened 2022-04-05
Update to Rust 2021; bump MSRV to 1.56.0

bgilbert:
RHEL 8.6 has a new-enough Rust to allow this. Dependencies are starting to require Rust 2021, and clap 3.0 needs Rust 1.54.0.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ permissions:
 env:
   CARGO_TERM_COLOR: always
   # Minimum supported Rust version (MSRV)
-  ACTIONS_MSRV_TOOLCHAIN: 1.49.0
+  ACTIONS_MSRV_TOOLCHAIN: 1.56.0
   # Pinned toolchain for linting
   ACTIONS_LINTS_TOOLCHAIN: 1.59.0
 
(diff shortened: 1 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -11,9 +11,9 @@ permissions:
 env:
   CARGO_TERM_COLOR: always
   # Minimum supported Rust version (MSRV)
-  ACTIONS_MSRV_TOOLCHAIN: 1.49.0
+  ACTIONS_MSRV_TOOLCHAIN: 1.56.0
```
lucab: This can be dropped now.

.github/workflows/rust.yml
```diff
@@ -52,10 +52,16 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+      - name: Detect crate MSRV
+        run: |
+          msrv=$(cargo metadata --format-version 1 --no-deps | \
```
lucab, line 55: This is lovely! :heart_eyes:

.github/workflows/rust.yml
bgilbert: Whoops, good catch.

lucab (approved):

PR #748, opened 2022-05-09
providers/aws: expose instance availability-zone-id as AWS_AVAILABILITY_ZONE_ID

abhinavdahiya:
ec2 instance metadata exposes the physical id for availability zone. Physical zone ids are useful
because they are same across AWS accounts and helpful when trying to decide if instances are actually
in the same zone when reconciling that between instances in different accounts.

See doc:
> https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html

```
placement/availability-zone-id	The static Availability Zone ID in which the instance is launched. The Availability Zone ID is consistent across accounts. However, it might be different from the Availability Zone, which can vary by account.	2020-08-24
```

This is available in api versions starting 2020-08-24, and afterbun already uses 2021-01-03 [1].

[1]: https://github.com/coreos/afterburn/blob/640ce9790c4f2d56aa75d1e3ad63f8011bec4255/src/providers/aws/mod.rs#L83

### DIFF
--- a/src/providers/aws/mock_tests.rs
+++ b/src/providers/aws/mock_tests.rs
@@ -48,6 +48,7 @@ fn aws_get_maps() -> (
     let ipv4_public = "test-ipv4-public";
     let ipv6 = "test-ipv6";
     let availability_zone = "test-availability-zone";
+    let availability_zone_id = "test-availability-zone-id";
     let hostname = "test-hostname";
     let public_hostname = "test-public-hostname";
     let instance_id_doc = r#"{"region": "test-region"}"#;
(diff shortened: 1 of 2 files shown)


bgilbert (approved):
LGTM, thanks!

bgilbert:
Known test failure.  Merging.

abhinavdahiya:
@bgilbert any idea when this rolls out to FCOS ?

bgilbert:
No Afterburn release is currently scheduled.  Is something blocking on this change?

sohankunkerkar:
@abhinavdahiya it should be available in the next `testing` release

https://koji.fedoraproject.org/koji/buildinfo?buildID=1983824

PR #767, opened 2022-06-30
docs: add release notes

bgilbert:
Copied from GitHub releases, with draft notes for unreleased changes.

Drop the crate/vendor hashes, since otherwise there's a circular dependency; we don't know them until after we tag and build.


marmijo (approved):
LGTM

PR #771, opened 2022-07-08
providers: Use inline `format!` in a few places

cgwalters:
It reads more nicely.  No particular motivation, this is just
a drive-by random change.


bgilbert:
Content LGTM.  This would require bumping the MSRV to 1.58, which we could do, since RHEL 8 and 9 both have 1.58 now.

jlebon:
MSRV bump in https://github.com/coreos/afterburn/pull/775.

jlebon:
Can you rebase this?

jlebon (approved):

PR #784, opened 2022-08-02
dependabot: automatically add `dependency` and `skip-notes` labels to PRs

bgilbert:
Dependabot was defaulting to a `dependencies` label, which doesn't match our other repos.  Rename it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,6 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
+  labels:
+  - dependency
+  - skip-notes


.github/dependabot.yml
```diff
@@ -5,3 +5,6 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
+  labels:
+  - dependency
```
prestist, line 9: Did we intentionally add "dependency" label?

prestist: Ah sorry, missed your comment.

prestist (approved):
LGTM

PR #826, opened 2022-11-02
cargo: stop enabling LTO in release builds

bgilbert:
It's not yet the upstream default, and it's caused miscompilations in the past on non-x86 builds.  We don't especially need the extra optimization, so let's be conservative and drop it.


cgwalters (approved):

lucab:
See https://github.com/coreos/zincati/pull/877#issuecomment-1300986124 for the discussion behind this.

PR #861, opened 2023-01-28
Support DHCP option lookup from NetworkManager

### LINKED ISSUE #146: Read lease files from dhclient and NetworkManager on Azure and CloudStack
RHCOS/FCOS uses NetworkManager presently. So the logic to look for option 245 in the systemd-networkd lease file would need to be adapted to read from those lease files (`/var/lib/NetworkManager` on RHEL7, `/var/lib/dhclient` on FCOS and likely RHEL8). Might be worth investigating if NM's D-Bus API exposes this so we don't have to learn dhclient's syntax.

Though note that this is not a blocker for getting those platforms on Azure if we're comfortable relying exclusively on the fallback endpoint (#145) until this is implemented.

Related: #130

bgilbert:
Read DHCP options from NetworkManager and networkd in parallel.  The NetworkManager check works with both dhclient and NM's internal DHCP client.  Avoids a 30-second boot delay on Azure and AzureStack as we time out waiting for a networkd lease.  Doesn't help in the initrd on RHCOS 8, since NetworkManager doesn't daemonize there.

Support zbus 2.3 because Fedora isn't shipping zbus 3 yet.  The two are API-compatible for our purposes.

Tested on Azure on Flatcar and RHEL in the real root, and FCOS in the initrd.  Not tested on CloudStack, except to verify that NetworkManager exposes a `dhcp_server_identifier` DHCP option with the expected format.

Fixes https://github.com/coreos/afterburn/issues/146.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -15,6 +15,7 @@ Minor changes:
 - Add release notes to documentation
 - Fix default dependency ordering on all `checkin` services
 - Fix failure setting SSH keys on IBM Cloud if none are provided
+- Don't ignore network interfaces that appear during DHCP option lookup retry
 
 Packaging changes:
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -8,6 +8,8 @@ nav_order: 8
 
 Major changes:
 
+- Support reading DHCP options from NetworkManager, fixing 30s delay on
+  Azure, Azure Stack, and CloudStack
 
 Minor changes:
 
--- a/src/providers/microsoft/azure/mod.rs
+++ b/src/providers/microsoft/azure/mod.rs
@@ -179,7 +179,7 @@ impl Azure {
 
     #[cfg(not(test))]
     fn get_fabric_address_from_dhcp() -> Result<IpAddr> {
-        let v = crate::util::dns_lease_key_lookup("OPTION_245")?;
+        let v = crate::util::DhcpOption::AzureFabricAddress.get_value()?;
         // value is an 8 digit hex value. convert it to u32 and
         // then parse that into an ip. Ipv4Addr::from(u32)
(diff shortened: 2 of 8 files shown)


bgilbert:
`zbus` appears to require a newer MSRV.

cgwalters (approved):
Looks good to me!

src/providers/microsoft/azure/mod.rs
```diff
@@ -179,7 +179,7 @@ impl Azure {
 
     #[cfg(not(test))]
     fn get_fabric_address_from_dhcp() -> Result<IpAddr> {
-        let v = crate::util::dns_lease_key_lookup("OPTION_245")?;
+        let v = crate::util::DhcpOption::AzureFabricAddress.get_value()?;
```
prestist, line 182: This LN and 197 is way clearer!

docs/release-notes.md
```diff
@@ -15,6 +15,7 @@ Minor changes:
 - Add release notes to documentation
 - Fix default dependency ordering on all `checkin` services
 - Fix failure setting SSH keys on IBM Cloud if none are provided
+- Check network interfaces that show up while querying networkd DHCP options
```
prestist: Hmm, WDYT about these options:

1. While querying networkd DHCP options, continue to expand query with any new found network interfaces

2. Add support for expanding active networkid DHCP options query with new found network interfaces

3. Update networkid DHCP options query to be fluid and expand with new found network interfaces

prestist (approved):
LGTM

docs/release-notes.md
bgilbert: Took another shot at it.

prestist: Thank you; lgtm

PR #866, opened 2023-02-03
cargo: add configuration for cargo-vendor-filterer

bgilbert:
See https://github.com/coreos/cargo-vendor-filterer.

This project only targets Linux so we don't need any other platform dependencies.  Notably this obsoletes the manual step to strip all the static library files for the Windows crates.

```
$ find vendor-old -name '*.a' -delete
$ du -sh vendor-old vendor-new
155M	vendor-old
77M	vendor-new
```


prestist (approved):
LGTM!

PR #883, opened 2023-02-20
Small fixups

bgilbert:


src/util/cmdline.rs
```diff
@@ -62,7 +62,7 @@ fn contains_flag_prefix(cmdline: &str, prefix: &str) -> bool {
     cmdline.split(' ').any(|s| s.starts_with(prefix))
 }
 
-// Find OEM ID flag value in cmdline string.
+// Find argument value in cmdline string.
```
prestist: sorry just a nit, argument's ?

bgilbert: The function argument is `flagname` so I guess I should use `flag`.

bgilbert:
Updated.

prestist (approved):
LGTM

PR #888, opened 2023-03-10
providers/vmware: Process guestinfo.metadata netplan configuration

pothos:
The network environment can be dynamic and thus needs to be provided as
    VM metadata. Since the format should not depend on whether the VM runs
    uses Ignition and Afterburn or Cloud-Init, the idea is to also support
    the guestinfo.metadata variable as used by Cloud-Init which contains
    Netplan YAML network configuration.
    Add a new command to write out netplan configs to a given directory,
    similar as we do with networkd units. While this is currently just used
    for VMware, other providers could also construct the netplan data type
    to provide netplan configurations if the OS rather wants to use
    NetworkManager than systemd-networkd. For backwards compatibility and
    to not need netplan it would be nice to keep the systemd-networkd
    support as long as its used.

What do you think, rather only write out to `/etc/netplan/afterburn.yaml` and rely on the netplan implementation, or do the processing YAML directly in afterburn?

References:
* https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html#walkthrough-of-guestinfo-keys-transport 
* https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v2.html
* https://netplan.io/reference/
* https://linux-on-z.blogspot.com/p/using-netplan-on-ibm-z.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,9 +35,11 @@ debug = true
 
 [dependencies]
 anyhow = "1.0"
+base64 = "0.21"
 cfg-if = "1.0"
 clap = { version = "4", "default_features" = false, "features" = ["std", "cargo", "derive", "error-context", "help", "suggestions", "usage", "wrap_help"] }
 ipnetwork = ">= 0.17, < 0.21"
+libflate = "1.3"
 libsystemd = ">= 0.2.1, < 0.7.0"
 mailparse = ">= 0.13, < 0.15"
 maplit = "1.0"
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -9,6 +9,7 @@ nav_order: 8
 Major changes:
 
 - Add support for Scaleway
+- Add Netplan guestinfo support on VMware
 
 Minor changes:
 
--- a/docs/usage/vmware-netplan-guestinfo-metadata.md
+++ b/docs/usage/vmware-netplan-guestinfo-metadata.md
@@ -1,0 +1,27 @@
+---
+nav_order: 2
+parent: Usage
+---
+
+# VMware Netplan guestinfo metadata
+
+The network environment can vary between VMware servers and instead of leaking these requirements into userdata snippets, a well-known guestinfo metadata field can be used.
+The guestinfo metadata field is OS-independent and supported by cloud-init (spec [here](https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v2.html), example [here](https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html#walkthrough-of-guestinfo-keys-transport)) and Afterburn. When the OS supports this mechanism the user can provide Netplan configs which the OS renders using the backend of choice.
+
+## Specifying the guestinfo metadata
+
+The guestinfo keys are named `guestinfo.metadata` for the content and `guestinfo.metadata.encoding` to specify the encoding of the content.
+The value of the encoding field can be empty to indicate raw string data, or one of `base64` or `b64` to indicate an base64 encoding, or one of `gzip+base64` or `gz+b64` to indicate base64-encoded gzip data.
+
+An example for raw string data is the following:
+```
+network:
+  version: 2
+  ethernets:
+    nics:
+      match:
+        name: ens*
+      dhcp4: yes
+```
+
+The supported config format with examples can be found in the [Netplan specification](https://netplan.readthedocs.io/en/latest/netplan-yaml/).
--- a/src/cli/multi.rs
+++ b/src/cli/multi.rs
@@ -26,6 +26,9 @@ pub struct CliMulti {
     /// The directory into which network units are written
     #[arg(long = "network-units", value_name = "path")]
     network_units_dir: Option<String>,
+    /// The directory into which a netplan config is written
+    #[arg(long = "netplan-config", value_name = "path")]
+    netplan_config_dir: Option<String>,
     /// Update SSH keys for the given user
     #[arg(long = "ssh-keys", value_name = "username")]
     ssh_keys_user: Option<String>,
@@ -41,6 +44,7 @@ impl CliMulti {
 
         if self.attributes_file.is_none()
             && self.network_units_dir.is_none()
+            && self.netplan_config_dir.is_none()
             && !self.check_in
             && self.ssh_keys_user.is_none()
             && self.hostname_file.is_none()
@@ -71,6 +75,11 @@ impl CliMulti {
         self.network_units_dir
             .map_or(Ok(()), |x| metadata.write_network_units(x))
             .context("writing network units")?;
+
+        // write netplan config if configured to do so
+        self.netplan_config_dir
+            .map_or(Ok(()), |x| metadata.write_netplan_config(x))
+            .context("writing netplan config")?;
 
         // perform boot check-in.
         if self.check_in {
--- a/src/providers/vmware/amd64.rs
+++ b/src/providers/vmware/amd64.rs
@@ -4,9 +4,15 @@
 
 use super::VmwareProvider;
 use anyhow::{bail, Context, Result};
+use base64::{engine::general_purpose, Engine as _};
+use libflate::gzip::Decoder;
+use serde_json::json;
+use std::io::Read;
 
 /// Guestinfo key for network kargs.
 static INITRD_NET_KARGS: &str = "guestinfo.afterburn.initrd.network-kargs";
+static METADATA: &str = "guestinfo.metadata";
+static METADATA_ENCODING: &str = "guestinfo.metadata.encoding";
 
 impl VmwareProvider {
     /// Build the VMware provider, fetching and caching guestinfo entries.
@@ -24,11 +30,28 @@ impl VmwareProvider {
             vmw_backdoor::probe_backdoor()
         })?;
 
-        let mut erpc = backdoor.open_enhanced_chan()?;
-        let guestinfo_net_kargs = Self::fetch_guestinfo(&mut erpc, INITRD_NET_KARGS)?;
+        let guestinfo_net_kargs = {
+            // Use a block, otherwise we would have to drop(erpc) manually
+            let mut erpc = vmw_backdoor::EnhancedChan::open(&mut backdoor)?;
+            Self::fetch_guestinfo(&mut erpc, INITRD_NET_KARGS)?
+        };
+
+        let guestinfo_metadata_raw = {
+            let mut erpc = vmw_backdoor::EnhancedChan::open(&mut backdoor)?;
+            Self::fetch_guestinfo(&mut erpc, METADATA)?
+        };
+
+        let guestinfo_metadata_encoding = {
+            let mut erpc = vmw_backdoor::EnhancedChan::open(&mut backdoor)?;
+            Self::fetch_guestinfo(&mut erpc, METADATA_ENCODING)?
+        };
+
+        let guestinfo_metadata =
+            parse_metadata(guestinfo_metadata_encoding, guestinfo_metadata_raw)?;
 
         let provider = Self {
             guestinfo_net_kargs,
+            guestinfo_metadata,
         };
 
         slog_scope::trace!("cached vmware provider: {:?}", provider);
@@ -39,8 +62,144 @@ impl VmwareProvider {
     fn fetch_guestinfo(erpc: &mut vmw_backdoor::EnhancedChan, key: &str) -> Result<Option<String>> {
         let guestinfo = erpc
             .get_guestinfo(key.as_bytes())
-            .context("failed to retrieve network kargs")?
+            .with_context(|| format!("failed to retrieve guestinfo for {}", key))?
             .map(|bytes| String::from_utf8_lossy(&bytes).into_owned());
         Ok(guestinfo)
     }
+
+    pub fn parse_netplan_config(&self) -> Result<Option<String>> {
+        if let Some(metadata) = &self.guestinfo_metadata {
+            // We need to parse the netplan config to remove the non-Netplan keys.
+            // The data can either be JSON or YAML, but since JSON is a subset of
+            // YAML we don't need to try serde_json::from_str here.
+            let netplan_config_unfiltered: serde_json::Value =
+                serde_yaml::from_str(metadata).context("invalid YAML/JSON metadata")?;
+            // Only the "network" key is allowed to be present.
+            // We use the json! macro but this is only about creating a serde value::Value,
+            // even though its name sounds like it would create JSON.
+            let netplan_config_filtered = json!({
+                "network": netplan_config_unfiltered.get("network").context("no 'network' key found")?
+            });
+            Ok(Some(serde_yaml::to_string(&netplan_config_filtered)?))
+        } else {
+            Ok(None)
+        }
+    }
+
+    #[cfg(test)]
+    pub fn new_from_metadata(metadata: String) -> Result<Self> {
+        Ok(Self {
+            guestinfo_net_kargs: None,
+            guestinfo_metadata: Some(metadata),
+        })
+    }
+}
+
+fn parse_metadata(
+    guestinfo_metadata_encoding: Option<String>,
+    guestinfo_metadata_raw: Option<String>,
+) -> Result<Option<String>> {
+    match (
+        guestinfo_metadata_encoding.as_deref(),
+        guestinfo_metadata_raw,
+    ) {
+        (Some("base64" | "b64"), Some(guestinfo_metadata_raw_val)) => {
+            let decoded =
+                general_purpose::STANDARD.decode(guestinfo_metadata_raw_val.as_bytes())?;
+            Ok(Some(String::from_utf8(decoded)?))
+        }
+        (Some("gzip+base64" | "gz+b64"), Some(guestinfo_metadata_raw_val)) => {
+            let decoded =
+                general_purpose::STANDARD.decode(guestinfo_metadata_raw_val.as_bytes())?;
+            let mut decompressor = Decoder::new(decoded.as_slice())?;
+            let mut uncompressed = Vec::new();
+            decompressor.read_to_end(&mut uncompressed)?;
+            Ok(Some(String::from_utf8(uncompressed)?))
+        }
+        (Some(""), guestinfo_metadata_raw) => Ok(guestinfo_metadata_raw),
+        (Some(encoding), _) => bail!("unknown guestinfo.metadata.encoding '{}'", encoding),
+        (None, guestinfo_metadata_raw) => Ok(guestinfo_metadata_raw),
+    }
+}
+
+#[test]
+fn test_netplan_json() {
+    let metadata = r#"{
+      "network": {
+        "ethernets": {
+          "nics": {
+            "match": {
+              "name": "ens*"
+            }
+          }
+        }
+      },
+      "ExcludeNonNetplanField": 0
+    }"#;
+    let provider = VmwareProvider::new_from_metadata(metadata.to_owned()).unwrap();
+    let netplan_config = provider.parse_netplan_config().unwrap().unwrap();
+    let expected = r#"network:
+  ethernets:
+    nics:
+      match:
+        name: ens*
+"#;
+    assert_eq!(netplan_config, expected);
+}
+
+#[test]
+fn test_netplan_dhcp() {
+    let metadata = r#"network:
+  ethernets:
+    nics:
+      match:
+        name: ens*
+"#;
+    let provider = VmwareProvider::new_from_metadata(metadata.to_owned()).unwrap();
+    let netplan_config = provider.parse_netplan_config().unwrap().unwrap();
+    assert_eq!(netplan_config, metadata);
+}
+
+#[test]
+fn test_metadata_plain_1() {
+    let guestinfo_metadata_raw = Some("hello".to_owned());
+    let parsed = parse_metadata(None, guestinfo_metadata_raw)
+        .unwrap()
+        .unwrap();
+    assert_eq!(parsed, "hello");
+}
+
+#[test]
+fn test_metadata_plain_2() {
+    let guestinfo_metadata_raw = Some("hello".to_owned());
+    let parsed = parse_metadata(Some("".into()), guestinfo_metadata_raw)
+        .unwrap()
+        .unwrap();
+    assert_eq!(parsed, "hello");
+}
+
+#[test]
+fn test_metadata_base64() {
+    let guestinfo_metadata_raw = Some("aGVsbG8=".to_owned());
+    let parsed = parse_metadata(Some("base64".into()), guestinfo_metadata_raw.clone())
+        .unwrap()
+        .unwrap();
+    assert_eq!(parsed, "hello");
+    let parsed_b64 = parse_metadata(Some("b64".into()), guestinfo_metadata_raw)
+        .unwrap()
+        .unwrap();
+    assert_eq!(parsed_b64, "hello");
+}
+
+#[test]
+fn test_metadata_gzip_base64() {
+    let guestinfo_metadata_raw = Some("H4sIAAAAAAACA8tIzcnJBwCGphA2BQAAAA==".to_owned());
+    let parsed = parse_metadata(Some("gzip+base64".into()), guestinfo_metadata_raw.clone())
+        .unwrap()
+        .unwrap();
+    assert_eq!(parsed, "hello");
+    let parsed_b64 = parse_metadata(Some("gz+b64".into()), guestinfo_metadata_raw)
+        .unwrap()
+        .unwrap();
+    assert_eq!(parsed_b64, "hello");
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,13 +3,21 @@
 version = 3
 
 [[package]]
+name = "adler32"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
+
+[[package]]
 name = "afterburn"
 version = "5.4.3"
 dependencies = [
  "anyhow",
+ "base64 0.21.2",
  "cfg-if",
  "clap",
  "ipnetwork",
+ "libflate",
  "libsystemd",
  "mailparse",
  "maplit",
@@ -387,6 +395,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3e4c1eaa2012c47becbbad2ab175484c2a84d1185b566fb2cc5b8707343dfe58"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "crc32fast"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b540bd8bc810d3885c6ea91e2018302f68baba2129ab3e88f32389ee9370880d"
+dependencies = [
+ "cfg-if",
 ]
 
 [[package]]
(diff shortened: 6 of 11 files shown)


pothos:
@bgilbert @lucab Any preference? While I think that doing all processing in Afterburn would be nice, the more practical approach would be to write the netplan config file out and let the distro handle it through the regular `netplan` implementation.
Edit: Also an option, don't do it in afterburn but call `vmtoolsd --cmd 'info-get guestinfo.metadata'` directly to write out this file for netplan.

bgilbert:
Thanks for the PR.  Afterburn already has a mechanism for this on VMware, #404, but it isn't very user-friendly (initrd `ip=` syntax is not great).  Afterburn's network configuration support hasn't gotten a lot of attention in recent years, due to the longstanding problem that it's designed for networkd and doesn't map well onto NetworkManager.  I wasn't aware of Netplan and wish I had known about it sooner; it looks like a good fit for Afterburn's needs.

I see an opportunity here to revisit our network configuration design, and reuse existing abstractions rather than maintaining our own. Rather than having the DigitalOcean and Packet providers populate networkd-format structs, they (and the VMware provider) could populate a `netplan-types` struct, and then Afterburn could write that out via a new `--netplan-conf <directory>` option.  The distro would be responsible for configuring Netplan to translate to networkd or NetworkManager as desired.

WDYT?  If you have an immediate need for this, it could be reasonable to add some stub Netplan support for VMware but defer migration of the other providers to a later PR.

pothos:
Thanks for the feedback, ok, I'm actually happy that we generate networkd units directly in afterburn but it's good to know that you are looking for a solution that also covers NetworkManager-only systems. Flatcar doesn't ship netplan (and ok course also doesn't ship [nmstate](https://nmstate.io/)). We could add it to the VMware Flatcar images and limit the use to it at first, and later see if that's something we could use for other providers, too. I'll drop the unit generation then and only parse the file once and emit it again to have a way where other providers could expose a netplan struct as you said.

pothos:
Done but keeping it it draft mode until I test it on a real system

src/providers/vmware/amd64.rs
```diff
@@ -43,4 +54,111 @@ impl VmwareProvider {
...
+            }
+        } else {
+            guestinfo_metadata_raw
+        };
+    Ok(guestinfo_metadata)
+}
+
+#[test]
+fn test_netplan_dhcp() {
+    let metadata = r#"network:
```
akutz, line 152: I am not familiar with Rust, so please pardon my ignorance, but does the parsing code care if the `network` key is not the only top-level value in the metadata? Because it would not have to be. Also, we support sending the metadata either as YAML _or_ JSON, so is it possible to support both?

pothos: Other keys can be present, but are ignored. Yes, JSON support could be added, too.

pothos: Pushed JSON support, please have a look

src/providers/vmware/amd64.rs
```diff
@@ -24,11 +29,24 @@ impl VmwareProvider {
             vmw_backdoor::probe_backdoor()
         })?;
 
-        let mut erpc = backdoor.open_enhanced_chan()?;
+        let mut erpc = vmw_backdoor::EnhancedChan::open(&mut backdoor)?;
         let guestinfo_net_kargs = Self::fetch_guestinfo(&mut erpc, INITRD_NET_KARGS)?;
 
+        drop(erpc);
```
pothos: Multiple calls on a channel don't work:
```
    2: vmware backdoor error: erpc send_command_len failed
```
and the channel holds the mutable reference, hence to create a new channel, I needed to drop. I think the explicit call is better than doing this implicitly through blocks because this will be confusing for the next person to touch.

pothos:
Tested with a Flatcar VM as follows, the VM didn't have the guestinfo.metadata set on provisioning:
```
cargo build
scp -C target/debug/afterburn core@ESXIVM:
ssh core@ESXIVM
# now in the VM:
./afterburn multi --netplan-configs $PWD/ --provider vmware
# no afterburn.yaml will be written to PWD
/usr/share/oem/bin/vmtoolsd --cmd "info-set guestinfo.metadata bmV0d29yazoKICB2ZXJzaW9uOiAyCiAgZXRoZXJuZXRzOgogICAgbmljczoKICAgICAgbWF0Y2g6CiAgICAgICAgbmFtZTogZW5zKgogICAgICBkaGNwNDogdHJ1ZQo="
/usr/share/oem/bin/vmtoolsd --cmd "info-set guestinfo.metadata.encoding base64"
./afterburn multi --netplan-configs $PWD/ --provider vmware
# the afterburn.yaml gets written to PWD
cat afterburn.yaml
# prints this:
# network:
#  version: 2
#  ethernets:
#    nics:
#      match:
#        name: ens*
#      dhcp4: true
```

src/providers/vmware/amd64.rs
akutz: Lgtm, thanks!

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +57,152 @@ impl VmwareProvider {
...
+    guestinfo_metadata_encoding: Option<String>,
+    guestinfo_metadata_raw: Option<String>,
+) -> Result<Option<String>> {
+    let guestinfo_metadata =
+        if let Some(guestinfo_metadata_encoding_val) = guestinfo_metadata_encoding {
+            match (
+                guestinfo_metadata_encoding_val.as_str(),
+                guestinfo_metadata_raw,
+            ) {
+                ("base64", Some(guestinfo_metadata_raw_val)) => {
```
akutz: Please note there are multiple, documented options for expressing the encoding:

* b64
* base64
* gz+b64
* gzip+base64

akutz (changes requested):
One minor change requested, and then it's good to go!

bgilbert:
CI failures are a failed lint, missing release notes in `docs/release-notes.md`, and the fact that `netplan-types` isn't packaged in Fedora.  We have a rule that new dependencies must be packaged in Fedora first, so we don't have an unknown amount of packaging to do (including transitive dependencies) at release time.  You're not necessarily responsible for handling that yourself (unless you want to!) but it'll need to be addressed.

I won't be able to do the review here, unfortunately, but hopefully one of the other folks on the team can take a look.  Thanks for your work on this PR!

Cargo.toml
```diff
@@ -38,12 +38,15 @@ debug = true
 
 [dependencies]
 anyhow = "1.0"
+base64 = "0.21.0"
```
bgilbert: Please truncate these to `0.21` etc., so Dependabot doesn't feel the need to bump the Z version.

cgwalters:
>  The distro would be responsible for configuring Netplan to translate to networkd or NetworkManager as desired.

Would flatcar ship code to translate netplan to systemd-networkd?

I am not totally sure we want to own code doing this for NetworkManager.

(Not to architecture astronaut this too much but maybe what could be tractable is translating some networkd configs to NetworkManager...)

jepio:
> > The distro would be responsible for configuring Netplan to translate to networkd or NetworkManager as desired.
> 
> Would flatcar ship code to translate netplan to systemd-networkd?
> 
> I am not totally sure we want to own code doing this for NetworkManager.
> 
> (Not to architecture astronaut this too much but maybe what could be tractable is translating some networkd configs to NetworkManager...)

The plan is like @bgilbert mentioned above "The distro would be responsible for configuring Netplan to translate to networkd or NetworkManager as desired.". There are also plans to write a rust implementation of netplan from what I heard.

src/providers/vmware/amd64.rs
pothos: Ok, added that

Cargo.toml
pothos: Ok, done for the others, too

pothos:
I've addressed the partly review, it's ready for a full review.
Can someone do that?
Is there someone I could ask about Rust Fedora packaging?

> Would flatcar ship code to translate netplan to systemd-networkd?
> 
> I am not totally sure we want to own code doing this for NetworkManager.
> 
> (Not to architecture astronaut this too much but maybe what could be tractable is translating some networkd configs to NetworkManager...)

Netplan supports translation to either networkd or NetworkManager. For Flatcar we would add Netplan to the VMware OEM image and invoke it for translation to ephemeral networkd units under `/run` (probably from the initrd to have networking for Ignition, and this would also be carried over to the final system).

docs/release-notes.md
```diff
@@ -9,6 +9,7 @@ nav_order: 8
 Major changes:
 
 - Add support for Scaleway
+- VMware: `afterburn multi --netplan-configs FOLDER --provider vmware` can now process `guestinfo.metadata` netplan configuration
```
jlebon: These notes are usually brief. E.g.

```suggestion
- Add Netplan support on VMware
```

Can you provide usage details in a new `docs/usage/` page instead? See for example `initrd-network-cmdline.md`, which gets rendered here: https://coreos.github.io/afterburn/usage/initrd-network-cmdline/.

Cargo.toml
```diff
@@ -35,12 +35,15 @@ debug = true
...
 anyhow = "1.0"
+base64 = "0.21"
 cfg-if = "1.0"
 clap = { version = "4", "default_features" = false, "features" = ["std", "cargo", "derive", "error-context", "help", "suggestions", "usage", "wrap_help"] }
 ipnetwork = ">= 0.17, < 0.21"
+libflate = "1.3"
 libsystemd = ">= 0.2.1, < 0.7.0"
 mailparse = ">= 0.13, < 0.15"
 maplit = "1.0"
+netplan-types = "0.3"
```
jlebon: One way to avoid having to package this in Fedora is to gate it behind a feature for now. If we do that though, we should also make sure that CI here is still exercising that path.

That said, you're definitely welcome to package it on Fedora if you're interested! See [this page](https://docs.fedoraproject.org/en-US/package-maintainers/Joining_the_Package_Maintainers/) for details on how to get started.

src/providers/vmware/amd64.rs
```diff
@@ -4,9 +4,14 @@
...
 use super::VmwareProvider;
 use anyhow::{bail, Context, Result};
+use base64::{engine::general_purpose, Engine as _};
+use libflate::gzip::Decoder;
+use std::io::Read;
 
 /// Guestinfo key for network kargs.
 static INITRD_NET_KARGS: &str = "guestinfo.afterburn.initrd.network-kargs";
+static METADATA: &str = "guestinfo.metadata";
+static METADATA_ENCODING: &str = "guestinfo.metadata.encoding";
```
jlebon, line 15: Hmm, I'm unsure about the use of `guestinfo.metadata`. The PR description says:

> Since the format should not depend on whether the VM runs
uses Ignition and Afterburn or Cloud-Init, the idea is to also support
the guestinfo.metadata variable as used by Cloud-Init which contains
Netplan YAML network configuration.

But in the cloud-init case, users would normally put their netplan configs under e.g. `write_files` or something, right? I'm worried re-using the same key as cloud-init on systems with Ignition and Afterburn where one would've expected an Ignition config to live there instead. (Another way to say this: either that key should be read by Ignition or not at all.) I'm not sure on the historical reason why Ignition created its own key rather than conform to the established `guestinfo.metadata` key, but at this point I think it'd add more to the confusion to introduce a reader.

If the goal is to try to land on something that could potentially be implemented cross-provisioning tool, then let's create a new key. E.g. `guestinfo.netplan.metadata`? Or maybe `guestinfo.networking.metadata` to make it agnostic to the networking tool too since one could imagine feeding nmstate configs through there instead. Otherwise, something Afterburn-specific is also fine, like `guestinfo.afterburn.netplan-configs`.

src/providers/vmware/amd64.rs
jlebon: Hmm, possibly there's a bug there in `vmw_backdoor` that we need to fix?

> I think the explicit call is better than doing this implicitly through blocks because this will be confusing for the next person to touch.

But blocks have the advantage that the compiler guarantees that the dropping is implemented correctly rather than having to verify the manual implementation. A comment on the first instance should suffice to explain to the reader what's going on.

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +57,160 @@ impl VmwareProvider {
...
+                } else {
+                    // The JSON parsing error is discarded,
+                    // and if YAML parsing fails as well, we only return this error.
+                    serde_yaml::from_str(metadata)?
+                };
+            // Also check the version to make it easier for the OS to say that
+            // this mechanism is supported or not.
+            if netplan_config.network.version != 2 {
+                bail!("only netplan version 2 supported");
+            }
```
jlebon: I don't know much about Netplan though that seems out of scope for Afterburn. Why not be version agnostic and let the OS deal with version compatibility?

pothos: The OS would run afterburn with this flag and then netplan with a particular version. If any netplan data can come in I thought it's better to narrow the supported values and to error out early - also, the netplan_types crate may not support newer versions.

src/providers/vmware/amd64.rs
pothos: `guestinfo.userdata` is for cloud-init userdata (and Ignition could have used that, too). The metadata is meant for platform-provided configuration, not user configuration. The name of the guestinfo field and the content is what VMware "specifies" by reusing what cloud-init already supports, and deviating from to something afterburn-specific would defeat the point - one can already do OS-specific things and this was proposed as an OS-independent way of network configuration. It's not so nice that cloud-init defines this in the end but at least netplan is not directly tied to cloud-init. @akutz can maybe give more background.

Cargo.toml
pothos: Thanks, good idea, did this now

docs/release-notes.md
pothos: Thanks for the hint

pothos:
> But blocks have the advantage that the compiler guarantees that the dropping is implemented correctly rather than having to verify the manual implementation. A comment on the first instance should suffice to explain to the reader what's going on.

Have changed it now to use blocks.

> One way to avoid having to package this in Fedora is to gate it behind a feature for now. If we do that though, we should also make sure that CI here is still exercising that path.

Have changed it to be behind a feature gate, and the CI will enable all features.

src/providers/vmware/amd64.rs
pothos: Ok, have changed it now to use blocks.

src/providers/vmware/amd64.rs
travier: From previous comments:
> The distro would be responsible for configuring Netplan to translate to networkd or NetworkManager as desired.

I don't think we should parse the netplan config in Afterburn. We should leave that to another unit in the initrd or the root to do the translation for the desired backend.

travier:
You probably should link to the cloud-init documentation in the docs if we don't have anything more platform generic.

src/providers/vmware/amd64.rs
jlebon: Ahhh sorry, I missed the `user` vs `meta` distinction. OK, sounds fine to me.

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +61,163 @@ impl VmwareProvider {
     fn fetch_guestinfo(erpc: &mut vmw_backdoor::EnhancedChan, key: &str) -> Result<Option<String>> {
         let guestinfo = erpc
             .get_guestinfo(key.as_bytes())
-            .context("failed to retrieve network kargs")?
+            .context(format!("failed to retrieve guestinfo for {}", key))?
```
jlebon:
```suggestion
            .with_context(|| format!("failed to retrieve guestinfo for {}", key))?
```

src/providers/vmware/amd64.rs
jlebon: If the OS wants to error out before getting all the way to Netplan generation, integrators can ship a systemd service that runs right `After` Afterburn to check the config version.

jlebon: And actually, it seems like the only requirement on `netplan-types`? So if we drop this function, we should be able to get rid of the feature gating.

src/providers/vmware/amd64.rs
jlebon: OK right, I totally missed that `guestinfo.metadata` is used today on VMware + cloud-init to pass Netplan configs. That makes a lot more sense now.

docs/usage/vmware-netplan-guestinfo-metadata.md
```diff
@@ -0,0 +1,36 @@
+---
+nav_order: 2
+parent: Usage
+---
+
+# VMware Netplan guestinfo metadata
+
+The network environment can vary between VMware servers and instead of leaking these requirements into userdata snippets, a well-known guestinfo metadata field can be used.
+The guestinfo metadata field is OS-independent and supported by cloud-init and afterburn. It requires Netplan to render the Netplan YAML format to either NetworkManager or systemd-networkd configuration files.
```
jlebon: Here and below: we capitalize Afterburn in prose.

docs/usage/vmware-netplan-guestinfo-metadata.md
```diff
@@ -0,0 +1,36 @@
...
+By default, systemd-networkd units are used. Since the renderer backend is defined in the Netplan config itself, requiring NetworkManager would rule out support for systems that don't use it (unless they would postprocess the `renderer` field to force it to `networkd`). Since systemd-networkd can work in parallel with NetworkManager, it's expected that the renderer field is left to its default.
```
jlebon: This seems overly specific to how Netplan is actually integrated in the OS. I'd just leave it as e.g.:

> The network environment can vary between VMware servers and instead of leaking these requirements into userdata snippets, a well-known guestinfo metadata field can be used. The guestinfo metadata field is OS-independent and supported by cloud-init and Afterburn. It accepts Netplan configs which the OS can then render down using the backend of choice.

docs/usage/vmware-netplan-guestinfo-metadata.md
```diff
@@ -0,0 +1,36 @@
...
+
+The afterburn invocation is as follows, where `FOLDER` could be `/run/netplan/`:
+
+```
+afterburn multi --netplan-configs FOLDER --provider vmware
+```
+
+Afterwards, `netplan generate` can be used to render the config files. If that is done before `systemd-networkd` runs, this is enough, but if the network already is up, `netplan apply` should be used instead.
```
jlebon: This type of stuff belongs better in `distro.md` since it's not user-facing but for integrators.

src/providers/vmware/amd64.rs
pothos: Sure, I'll drop it (my main motivation for the validation was to follow what usually was done)

jepio: there is one field that it would make sense to parse out of `guestinfo.metadata` and that's `local-hostname`. But we could do that without the `netplan-types` dependency, by carrying our own simple struct definition.

i would suggest doing this as a follow-up, what do you think?

jlebon: Makes sense to me! :+1:

pothos: And we also had the idea that `netplan-types` could be used in the providers for a NetworkManager-capable alternative to the `--network-units=` flag that Flatcar uses. If that needs arises we can still add it and it's not really needed at the moment.

pothos: Found one problem with dropping the parsing: Netplan seems to want YAML but cloud-init is also fine with JSON which the current PR here would convert to YAML. I guess we would have to drop JSON support then. I'll stop for now and leave the feature gate.

pothos:
I looked into removing the `netplan-types` crate but that means dropping JSON support. Cloud-init handles both YAML and JSON, and since @akutz asked for JSON in https://github.com/coreos/afterburn/pull/888#discussion_r1154803578 I don't want to remove it if the feature gate is ok.

docs/usage/vmware-netplan-guestinfo-metadata.md
pothos: Moved, thanks

docs/usage/vmware-netplan-guestinfo-metadata.md
pothos: Thanks, applied

docs/usage/vmware-netplan-guestinfo-metadata.md
pothos: Thanks, done

src/providers/vmware/amd64.rs
pothos: `local-hostname` is another top-level entry and not part of `netplan-types` because it's a cloud-init thing. Supporting and parsing that would be done with a separate struct.

jlebon:
Sorry, can you explain the issue in more details? Is the issue that we're not retaining the original format? Can't we note down which format it was in at deserialization time and make sure to reserialize it in the same format?

pothos:
The metadata can be either [JSON or YAML](https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html#guestinfo-keys) but the output for Netplan must be YAML, thus it gets converted from JSON to YAML with the `netplan-types` crate.
I thought that when we drop the crate we drop JSON support but maybe I can also do a generic JSON to YAML conversion with serde but we need to filter unknown keys from the YAML output.

jepio:
> The metadata can be either [JSON or YAML](https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html#guestinfo-keys) but the output for Netplan must be YAML

I thought YAML is a superset of JSON so valid JSON is automatically also valid YAML.

pothos:
Updated it to remove the netplan-types dependency.

> I thought YAML is a superset of JSON so valid JSON is automatically also valid YAML.

Yeah, maybe I delete the match statement.

pothos:
Removed the match statement because it works to only rely on YAML parsing.

src/cli/multi.rs
```diff
@@ -26,6 +26,9 @@ pub struct CliMulti {
     /// The directory into which network units are written
     #[arg(long = "network-units", value_name = "path")]
     network_units_dir: Option<String>,
+    /// The directory into which netplan configs are written
+    #[arg(long = "netplan-configs", value_name = "path")]
```
jlebon: Since we only write down one config, should this be

```suggestion
    #[arg(long = "netplan-config", value_name = "path")]
```

?

src/cli/multi.rs
```diff
@@ -72,6 +76,11 @@ impl CliMulti {
             .map_or(Ok(()), |x| metadata.write_network_units(x))
             .context("writing network units")?;
 
+        // write netplan configs if configured to do so
+        self.netplan_config_dir
+            .map_or(Ok(()), |x| metadata.write_netplan_configs(x))
+            .context("writing network units")?;
```
jlebon: Needs updating

src/cli/multi.rs
```diff
@@ -72,6 +76,11 @@ impl CliMulti {
             .map_or(Ok(()), |x| metadata.write_network_units(x))
             .context("writing network units")?;
 
+        // write netplan configs if configured to do so
+        self.netplan_config_dir
+            .map_or(Ok(()), |x| metadata.write_netplan_configs(x))
```
jlebon:
```suggestion
            .map_or(Ok(()), |x| metadata.write_netplan_config(x))
```

?

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +63,155 @@ impl VmwareProvider {
...
             .map(|bytes| String::from_utf8_lossy(&bytes).into_owned());
         Ok(guestinfo)
     }
+
+    pub fn parse_netplan_config(&self) -> Result<Option<String>> {
+        if let Some(metadata) = &self.guestinfo_metadata {
+            // We need to parse the netplan config to remove the non-Netplan keys.
+            // The data can either be JSON or YAML, but since JSON is a subset of
+            // YAML we don't need to try serde_json::from_str here.
+            let netplan_config_unfiltered: serde_json::Value = serde_yaml::from_str(metadata)?;
```
jlebon: Let's add error context here.

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +63,155 @@ impl VmwareProvider {
...
+            // Only the "network" key is allowed to be present.
+            // We use the json! macro but this is only about creating a serde value::Value,
+            // even though its name sounds like it would create JSON.
+            let netplan_config_filtered = json!({ "network": netplan_config_unfiltered.get("network").ok_or(anyhow!("no 'network' key found"))?});
```
jlebon: For clarity, can we reformat this as

```rust
let ... = json!({
	...
})
```

?

I don't _think_ `cargo fmt` will revert it.

jlebon:
```suggestion
            let netplan_config_filtered = json!({ "network": netplan_config_unfiltered.get("network").context("no 'network' key found")?});
```

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +63,155 @@ impl VmwareProvider {
...
+            let netplan_config = serde_yaml::to_string(&netplan_config_filtered)?;
+            Ok(Some(netplan_config))
```
jlebon: Minor: I'd collapse those two.

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +63,155 @@ impl VmwareProvider {
...
+            guestinfo_net_kargs: None,
+            guestinfo_metadata: Some(metadata),
+        })
+    }
+}
+
+fn parse_metadata(
+    guestinfo_metadata_encoding: Option<String>,
+    guestinfo_metadata_raw: Option<String>,
+) -> Result<Option<String>> {
```
jlebon, line 101: I think we could avoid some of the indentation here by leveraging pattern matching more. E.g. we could reflow this function as:

```rust
match (
	guestinfo_metadata_encoding_val.as_deref(),
	guestinfo_metadata_raw
) {
	(Some("base64" | "b64")) => {
		...
		Ok(Some(...))
	...
	(Some("") | None, guestinfo_metadata_raw) => Ok(...),
	(Some(enc), _) => Err(anyhow!("unknown guestinfo.metadata.encoding {enc}")),
}
```

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +63,155 @@ impl VmwareProvider {
...
+            },
+            "dhcp4": true
+          }
+        }
+      },
+      "ExcludeNonNetplanField": 0
+    }"#;
+    let provider = VmwareProvider::new_from_metadata(metadata.to_owned()).unwrap();
+    let netplan_config = provider.parse_netplan_config().unwrap().unwrap();
+    // Note: The test relies on the serde 'preserve_order' feature for equality
```
jlebon: It seems like that's the only reason we're adding `preserve_order`?

Can we avoid this by slightly tweaking the tests to just not have multiple subkeys in a key? We're not actually semantically parsing this anyway, and we're still testing the main thing, which is that we can parse JSON and YAML and exclude unknown top-level fields.

src/cli/multi.rs
pothos: My idea was that the Equinix Metal provider could also offer Netplan instead of networkd directly but yes, since it works to write one large config file instead of many being merged, it would still be ok in the singular even then.

src/providers/vmware/amd64.rs
pothos: Yes, done

src/providers/vmware/amd64.rs
pothos: Thanks

src/providers/vmware/amd64.rs
pothos: Yes, works

src/providers/vmware/amd64.rs
```diff
@@ -39,8 +62,144 @@ impl VmwareProvider {
...
+        (Some("gzip+base64" | "gz+b64"), Some(guestinfo_metadata_raw_val)) => {
+            let decoded =
+                general_purpose::STANDARD.decode(guestinfo_metadata_raw_val.as_bytes())?;
+            let mut decompressor = Decoder::new(decoded.as_slice())?;
+            let mut uncompressed = Vec::new();
+            decompressor.read_to_end(&mut uncompressed)?;
+            Ok(Some(String::from_utf8(uncompressed)?))
+        }
+        (Some(""), guestinfo_metadata_raw) => Ok(guestinfo_metadata_raw),
+        (Some(_), _) => bail!("unknown guestinfo.metadata.encoding"),
```
jlebon: Would be nice to the user to print the encoding string found in the error.

jlebon (approved):
One minor comment, but LGTM overall. Nice work!

src/providers/vmware/amd64.rs
pothos: Sure, done

jlebon (approved):

PR #889, opened 2023-03-10
docs: Use upstream theme & update to 0.4.1

travier:
Use a fixed tag for the theme so that we can directly pull it from upstream and skip vendoring the them in our org.


travier:
See preview at https://travier.github.io/afterburn/. It looks identical to me.

bgilbert (approved):

PR #894, opened 2023-03-20
Update mockito to 1.0.1

bgilbert:
The global server object has been removed, so we need to explicitly create servers and configure mocks on them.  Since individual providers can no longer query the global server object for the mock base URL to use, switch to having `retry::Client` rewrite URLs to use a scheme/host/port configured by an individual unit test.

In addition, individual mocks are no longer removed from servers when they go out of scope.  Stop artificially keeping mocks in scope.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -378,25 +378,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23d8666cb01533c39dde32bcbab8e227b4ed6679b2c925eba05feabea39508fb"
 
 [[package]]
-name = "deadpool"
-version = "0.9.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "421fe0f90f2ab22016f32a9881be5134fdd71c65298917084b0c7477cbc3856e"
-dependencies = [
- "async-trait",
- "deadpool-runtime",
- "num_cpus",
- "retain_mut",
- "tokio",
-]
-
-[[package]]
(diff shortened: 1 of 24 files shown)


bgilbert:
CI is blocked on [Fedora packaging](https://bugzilla.redhat.com/show_bug.cgi?id=2058896).

bgilbert:
Apparently [Afterburn is the only Fedora package](https://bugzilla.redhat.com/show_bug.cgi?id=2058896#c12) that uses mockito, so we can coordinate the packaging ratchet ourselves.  This PR is pretty large to carry as a downstream patch but I think that probably makes more sense than doing a dedicated upstream release just for updating a dev-dependency.

cgwalters (approved):
Skimmed this and it looks good!

prestist (approved):
LGTM.

Thank you for getting us updated, it seems like not a small effort to move us to latest version of mockito.

bgilbert:
Fedora [rust-mockito PR](https://src.fedoraproject.org/rpms/rust-mockito/pull-request/5) and [Afterburn package PR](https://src.fedoraproject.org/rpms/rust-afterburn/pull-request/25).

bgilbert:
Fedora 37 [update](https://bodhi.fedoraproject.org/updates/FEDORA-2023-f80d58cb7e) submitted.  CI should pass here once the update goes stable.

bgilbert:
Updated `Cargo.lock` to Mockito 1.0.1.

PR #922, opened 2023-05-11
Write short form hostname

gabriel-samfira:
While setting the hostname as a FQDN is perfectly valid, in most cases the expectation is that ```/etc/hostname``` holds the short form hostname and the FQDN is added as a canonical hostname, i.e. as the first column in ```/etc/hosts``` after the local IP address.

This will allow clients to use:

```
:~$ hostname -f
arrakis.local
```

to get the fqdn and still be able to get the short form hostname by calling:

```
:~$ hostname
arrakis
```

### DIFF
--- a/src/cli/multi.rs
+++ b/src/cli/multi.rs
@@ -21,8 +21,11 @@ pub struct CliMulti {
     #[arg(long)]
     check_in: bool,
     /// The file into which the hostname should be written
-    #[arg(long = "hostname", value_name = "path")]
+    #[arg(group = "hostname", long = "hostname", value_name = "path")]
     hostname_file: Option<String>,
+    /// Parse the hostname retrieved from metadata as a short hostname
+    #[arg(long = "short", requires = "hostname")]
+    short_hostname: bool,
     /// The directory into which network units are written
     #[arg(long = "network-units", value_name = "path")]
     network_units_dir: Option<String>,
@@ -64,7 +67,7 @@ impl CliMulti {
 
         // write hostname if configured to do so
         self.hostname_file
-            .map_or(Ok(()), |x| metadata.write_hostname(x))
+            .map_or(Ok(()), |x| metadata.write_hostname(x, self.short_hostname))
             .context("writing hostname")?;
 
         // write network units if configured to do so
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -241,6 +241,16 @@ pub trait MetadataProvider {
     fn write_hostname(&self, hostname_file_path: String) -> Result<()> {
         if let Some(mut hostname) = self.hostname()? {
             if let Some(maxlen) = max_hostname_len()? {
+                // Get the short form hostname (i.e. without the domain name).
+                // The domain name is usually set as a search domain via DHCP, and
+                // it can optionally be added as a canonical hostname to /etc/hosts
+                // as well, as the first value after the local IP address.
+                //
+                // Example:
+                // 127.0.0.1  example.local example
+                if let Some(idx) = hostname.find('.') {
+                    hostname.truncate(idx);
+                }
                 if hostname.len() > maxlen {
                     // Value exceeds the system's maximum hostname length.
                     // Truncate hostname to the first dot, or to the maximum
@@ -252,9 +262,6 @@ pub trait MetadataProvider {
                         maxlen
                     );
                     hostname.truncate(maxlen);
-                    if let Some(idx) = hostname.find('.') {
-                        hostname.truncate(idx);
-                    }
                 }
             }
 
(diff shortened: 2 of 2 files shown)


bgilbert:
We arrived at Afterburn's current hostname policy after much debate and debugging, and changing it would likely break existing users.  If you'd like to add a command-line option to write only the hostname part, we could consider it.

gabriel-samfira:
Fair enough! Would a similar command line option like ```--short-hostname=/etc/hostname``` be ok? (I am terrible at naming things).

bgilbert:
Works for me.  Note that there isn't a great way to configure the `afterburn-hostname.service` service at runtime, since it runs in the initrd.  (For example, Ignition can't reconfigure the initrd.)  So, if you wanted to use this, you'd have to configure a custom unit that runs in the booted system, which probably means you'd be setting the hostname after things have already started using the old one.

gabriel-samfira:
I was a bit sparse on context 😄 .

The need for this came up recently when we noticed that we can't bring up a k8s cluster on OpenStack using Flatcar as a base OS with ```kops```. Here is the relevant issue with more context: https://github.com/kubernetes/kops/issues/15385

For OpenStack we have a fix merged in [coreos-cloudinit](https://github.com/flatcar/coreos-cloudinit/), but the goal is to eventually use afterburn to set the hostname. The short hostname option will most likely be (eventually) included in future releases of Flatcar.

bgilbert:
Ah, okay, cool.  Happy to ship a command-line option to be invoked by a distro that integrates Afterburn.

gabriel-samfira:
I added a new boolean flag ```--short``` and placed the ```--hostname``` in a group. The ```--short``` flag requires the ```--hostname``` flag and toggles whether or not to write the short form hostname. I also added a new test.

This is my first interaction with rust, so let me know if the changes are not idiomatic or just plain silly :smile:.

gabriel-samfira:
Here is an example of the new flag in action.

Without the ```--short``` flag:

```bash
control-plane-nova-dkfvy2 ~ # ./afterburn --provider openstack-metadata --hostname=/root/test
May 16 14:51:15.352 INFO Fetching http://169.254.169.254/latest/meta-data/hostname: Attempt #1
May 16 14:51:15.542 INFO Fetch successful
May 16 14:51:15.542 INFO wrote hostname control-plane-nova-dkfvy2.novalocal to /root/test
control-plane-nova-dkfvy2 ~ # cat /root/test 
control-plane-nova-dkfvy2.novalocal
```

With the ```--short``` flag:

```bash
control-plane-nova-dkfvy2 ~ # ./afterburn --provider openstack-metadata --short --hostname=/root/test
May 16 14:51:24.876 INFO Fetching http://169.254.169.254/latest/meta-data/hostname: Attempt #1
May 16 14:51:24.908 INFO Fetch successful
May 16 14:51:24.908 INFO wrote hostname control-plane-nova-dkfvy2 to /root/test
control-plane-nova-dkfvy2 ~ # cat /root/test 
control-plane-nova-dkfvy2
```

For systems that don't explicitly set the new flag, things work as they did before. Opting to use the new flag will set the short form hostname.

src/cli/multi.rs
```diff
@@ -21,8 +21,11 @@ pub struct CliMulti {
     #[arg(long)]
     check_in: bool,
     /// The file into which the hostname should be written
-    #[arg(long = "hostname", value_name = "path")]
+    #[arg(group = "hostname", long = "hostname", value_name = "path")]
     hostname_file: Option<String>,
+    /// Parse the hostname retrieved from metadata as a short hostname
+    #[arg(long = "short", requires = "hostname")]
+    short_hostname: bool,
```
bgilbert, line 28: I think I prefer your original idea of having `--short-hostname=/etc/hostname`.  Rather than adding a mode switch that controls the behavior of another option, let's just support `--hostname` and `--short-hostname` as two output switches that operate independently.

src/providers/mod.rs
```diff
@@ -330,30 +342,35 @@ mod tests {
...
         assert_eq!(
-            try_write_hostname("hostname7.example.com"),
+            try_write_hostname("hostname7.example.com", false),
             "hostname7.example.com"
         );
+        // short form hostname
+        assert_eq!(
+            try_write_hostname("hostname7.example.com", true),
+            "hostname7"
+        );
```
bgilbert, line 355: Let's at least also test short hostnames where the input value doesn't have an FQDN.

bgilbert (commented):
Your first commit is a development checkpoint rather than an incremental logical change, so please squash it.  CI also expects a release note to be added in `docs/release-notes.md`.

bgilbert:
Thanks for the PR!  I won't be able to do future review rounds on this, unfortunately, but hopefully someone else on the team can take a look.

Closing/reopening to kick CI.

gabriel-samfira:
Thanks for the review @bgilbert !

I will make the changes today and push them.

PR #926, opened 2023-05-19
Update metadata for OpenSSL `parse2()` migration

bgilbert:
On versions of the `openssl` crate new enough to have `parse2()`, the legacy `parse()` function is a wrapper that panics when parsing the Azure SSH key certificate:

    thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', [...]/openssl-0.10.52/src/pkcs12.rs:42:31

We've switched to `parse2()` in the current codebase, but older versions built with newer `openssl` will fail.  Note the fix.

Also update the minimum version of `openssl` to one that has `parse2()`.

See https://github.com/coreos/fedora-coreos-tracker/issues/1492.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ mailparse = ">= 0.13, < 0.15"
 maplit = "1.0"
 nix = ">= 0.19, < 0.27"
 openssh-keys = ">= 0.5, < 0.7"
-openssl = "0.10"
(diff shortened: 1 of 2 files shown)


prestist (approved):
Thank you for doing this, and good catch.

bgilbert:
CoreOS CI doesn't test this.

PR #934, opened 2023-05-30
build-sys: Use new tier = 2 for cargo-vendor-filterer

cgwalters:
This fixes the problem described in the comment.


bgilbert (approved):

bgilbert:
CoreOS CI doesn't test the vendored tarball, and is having some CI flakes.  Merging over red.

PR #940, opened 2023-06-05
Pre-release-5.4.3

prestist:
https://github.com/coreos/afterburn/issues/939

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -16,6 +16,13 @@ Packaging changes:
 
 
 
+## Afterburn 5.4.3 (2023-06-05)
+
+Packaging changes:
+
+- Fix incomplete package in vendor tarball
+
+
 ## Afterburn 5.4.2 (2023-05-18)
 
 Minor changes:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,12 +36,18 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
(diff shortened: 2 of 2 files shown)


docs/release-notes.md
```diff
@@ -16,6 +16,14 @@ Packaging changes:
 
 
 
+## Afterburn 5.4.3 (2023-06-05)
+
+Minor changes:
+
+- Cargo update all dependencies to avoid missing data in vendor tarball
+
```
bgilbert, line 22: One too many trailing newlines.  There should be two between releases.

docs/release-notes.md
```diff
@@ -16,6 +16,14 @@ Packaging changes:
 
 
 
+## Afterburn 5.4.3 (2023-06-05)
+
+Minor changes:
+
+- Cargo update all dependencies to avoid missing data in vendor tarball
```
bgilbert: The last time this happened, in 5.4.1, we said: `Fix missing static archives in vendor tarball`.  They're not static archives in this case; maybe `Fix incomplete package in vendor tarball`?

Re the `cargo update` part, normally dependency updates don't get a release note.  If we were correcting years of missed updates, I'd definitely say it was worth noting, but the last update was b3e25907f7935e6e01a057e8497dd370e1ab419b in January.  So I'd be inclined to skip.

docs/release-notes.md
prestist: Ugh, sorry I followed the pattern from the (unreleased) section. I should have known better, and double checked.

docs/release-notes.md
prestist: Ok, I opted for your suggestion of `Fix missing static archives in vendor tarball`. Thank you for the feedback.

bgilbert: As I said above:

> They're not static archives in this case; maybe `Fix incomplete package in vendor tarball`?

prestist: Sorry I half read the suggestion; no excuse I fixed it.

bgilbert (approved):

PR #931, opened 2023-05-25
openstack: Add attribute OPENSTACK_INSTANCE_UUID

pierreprinetti:
In the OpenStack attributes, the field `OPENSTACK_INSTANCE_ID` reports
the instance ID fetched from the EC2-compatibility-layer metadata.
However, that identifier (in the format `i-something`) does not match
the one used to refer to machines in the OpenStack API.

With this patch, I propose to add a new OpenStack attribute
`OPENSTACK_INSTANCE_UUID` that reports the actual identifier for
OpenStack instances in the context of the OpenStack API.

Implementation-wise, this patch adds enables fetching the UUID from both
the config-drive and from the Nova metadata service endpoint. In order
to retain compatibility with old versions of the metadata service, this
patch uses the old endpoint `2012-08-10`.

Fixes #930

PR #942, opened 2023-06-06
openstack: Document the two platforms

pierreprinetti:
As detailed in https://github.com/coreos/afterburn/pull/462, Afterburn has two separate OpenStack endpoints, each with a different metadata gathering policy. This patch adds the relevant information to the docs.

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -45,11 +45,15 @@ The following platforms are supported, with a different set of features availabl
   - Attributes
   - SSH Keys
 * openstack
-  - Attributes
-  - SSH Keys
+  * Metadata source: config-drive if present, otherwise metadata service
+  * Features:
+      - Attributes
+      - SSH Keys
 * openstack-metadata
-  - Attributes
-  - SSH Keys
+  * Metadata source: metadata service
+  * Features:
+      - Attributes
+      - SSH Keys
 * packet
   - Attributes
   - First-boot check-in


bgilbert:
I agree that this document is probably the best place for this info, but the current PR draft adds it as an odd parenthetical and only for two of the providers.  I'd suggest something like this:

> - openstack
>   - Metadata source: configuration drive if present, otherwise metadata service
>   - Features:
>     - Attributes
>     - SSH keys

and ideally the PR would add that info for all providers.

docs/platforms.md
```diff
@@ -44,10 +44,10 @@ The following platforms are supported, with a different set of features availabl
 * kubevirt
   - Attributes
   - SSH Keys
-* openstack
+* openstack (from config-drive, then possibly fallback to the Nova metadata service)
```
prestist, line 47: Why don't  we document in a similar way as we did in this PR https://github.com/coreos/afterburn/pull/931/commits/6a06a02666d88450650c15136970f66be5322d7c#diff-7d38dbc6beeb2c1c6db5d85d6d58d87b81a09aee1f02ac76be60caf734a075aa

prestist, line 47: Why dont we document in a similar way so to follow the format fount here => https://github.com/coreos/afterburn/pull/931/commits/6a06a02666d88450650c15136970f66be5322d7c#diff-7d38dbc6beeb2c1c6db5d85d6d58d87b81a09aee1f02ac76be60caf734a075aa

pierreprinetti:
Reformatted as requested. PTAL!

travier (approved):

PR #996, opened 2023-09-04
providers/hetzner: add support for Hetzner Cloud

mhutter:
This commit implements initial support for Hetzner Cloud.

Supported features:
- Attributes
- Hostname
- SSH Keys

"hetzner" was choosen as the platform identifier, as proposed in https://github.com/coreos/fedora-coreos-tracker/issues/1324#issuecomment-1306688612.

Note that I decided to open a new PR over #125, since
- there was no activity in said PR for 4 years
- I did not want to bother with rebasing (I did not even try)
- I cannot push commits to existing PRs

but I don't have any hard feelings regarding merging this into #125 either.

Depending on how this goes I can work on the Ignition part as well later on.

Resolves: https://github.com/coreos/afterburn/pull/125
Relates: https://github.com/coreos/fedora-coreos-tracker/issues/1324


apricote (approved):
Hetzner Cloud employee here :wave: This looks good from a metadata service usage perspective.

Does the provider also need to implement the networks related methods from `MetadataProvider` (e.g. `networks`, `virtual_network_devices`, `rd_network_kargs`) or are these optional?

mhutter:
As I understand them they are optional, and configuration via DHCP "just works" in my tests, however I could imagine that this could be implemented in the future.

For example, it would be interesting to have somewhat stable interface identifiers (if that is even doable) for servers that have both public and internal IP's.

prestist:
@mhutter  Firstly thank you so much for working on this. I wanted to update you and let you know I will be taking a look at this PR in the coming days. I will try and get back to you with any questions soon. Thank you again!

prestist (approved):
After looking through the changes, I do not have any suggestions. Great job, and thank you for working so diligently on this.

ps, thank you for the clippy cleanup !

LGTM!

PR #1014, opened 2023-11-21
build(deps): allow building with libsystemd 0.7.0

decathorpe:
This allows building afterburn with the latest version of the libsystemd crate, v0.7.0. Bumping from v0.6.0 to v0.7.0 does not appear to require any code changes.

Not sure why dependabot didn't pick up this one, it made PRs for all previous versions.

PR #1016, opened 2023-11-22
Pre-release 5.5.0

mike-nguyen:
https://github.com/coreos/afterburn/issues/1013

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -4,10 +4,23 @@ nav_order: 8
 
 # Release notes
 
-## Upcoming Afterburn 5.5.0 (unreleased)
+## Upcoming Afterburn 5.6.0 (unreleased)
 
 Major changes:
 
+
+Minor changes:
+
+
+Packaging changes:
+
+
+
+## Afterburn 5.5.0 (2023-11-22)
+
+Major changes:
+
+- Add support for Hetzner Cloud
 - Add support for Scaleway
 - Add Netplan guestinfo support on VMware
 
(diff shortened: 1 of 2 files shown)


docs/release-notes.md
```diff
@@ -4,7 +4,19 @@ nav_order: 8
 
 # Release notes
 
-## Upcoming Afterburn 5.5.0 (unreleased)
+## Upcoming Afterburn 5.5.1 (unreleased)
```
jlebon: Releases are usually minor version bumps, so that's what we optimize for. So this would be 5.6.0.

docs/release-notes.md
```diff
@@ -13,9 +25,9 @@ Major changes:
 
 Minor changes:
 
+- providers: Add Hetzner Cloud
```
jlebon: Hmm, this should be in the "Major changes" section I think. It's on equal footing with the Scaleway change. And maybe let's reword it so it's similar to that one. (I.e. "Add support for Hetzner Cloud".)

jlebon (approved):
LGTM, thanks!

docs/release-notes.md
mike-nguyen: Thanks! Updated

jlebon:
Wait sorry, I think we want https://github.com/coreos/afterburn/pull/1014 so that we don't ship with a requirement on libsystemd v0.6.0 (https://github.com/lucab/libsystemd-rs/issues/147).

mike-nguyen:
includes libsystemd v0.7.0 now

jlebon (approved):

PR #1023, opened 2023-12-19
providers: support for proxmoxve

bitdriftr:
Hello,

This PR adds support for reading the cloud init drive from a ProxmoxVE instance.
The code mounts `/dev/disk/by-label/cidata` to a temprary location and then parses the YAML files present inside.
It supports : setting hostname, writing ssh keys, and writing systemd network configuration.

The changes introduces a new OEM ID `proxmoxve` that will be matched from the kernel command line. As time of writing, this OEM ID is not yet recognized by Ignition or other tools.

We currently maintain a build script that modifies a Flatcar image to inject this version of Afterburn : https://github.com/enix/flatcar-proxmoxve

Feedback is welcome :)


jlebon:
Thanks for the patch! For reference, Proxmox support has also been discussed in Fedora CoreOS in https://github.com/coreos/fedora-coreos-tracker/issues/736. And while it's not currently being actively pursued, I don't think that should prevent support for it to land in Afterburn. That said, I've tagged the ticket to be discussed again in [the next community meeting](https://github.com/coreos/fedora-coreos-tracker#meetings), at the very least so that we can agree on the platform ID. Feel free to join if you're able!

Are you planning to also implement the Ignition side of this?

travier:
https://github.com/coreos/fedora-coreos-tracker/issues/736#issuecomment-1896298378

abuisine:
> Are you planning to also implement the Ignition side of this?

Hi @jlebon, can you please precise what would be required on the ignition side ?
We would be happy to help (with @arcln)

abuisine:
Hi @jlebon,
Do you think this PR can be merged "as is" as is does not prevent other implementations like ignition / Fedora support of proxmox ?
It is my impression that the provider id is settled and that it was the principal show stopper.

Cheers.

abuisine:
for the record we are implementing the ignition support of the `proxmoxve` provider id cc @jlebon

pothos:
From the Flatcar side we would also be interested to have this.
I don't know if there is any specification about config drive contents from proxmox, but it looks like the `user-data` at least here is expected to be cloud-config to extract the ssh key and hostname. This conflicts with Ignition. Could this get moved to `meta-data`? I don't have a clue on what API proxmox provides to set this up. If none, then I guess we have to live with no ssh and hostname support in afterburn and let the user specify this in Ignition.

One more note about network units: This seems to be netplan - do you want to focus on a subset or support the full netplan spec? Recently I added a `--netplan-config` flag which only has the scope of copying this to, e.g., `/run/netplan/` and then we could run netplan to convert it to NetworkManager/networkd units.

src/providers/proxmoxve/tests.rs
```diff
@@ -0,0 +1,143 @@
...
+        vec![test_ssh_key]
+    );
+}
+
+#[test]
+fn test_network_dhcp() {
+    let config = ProxmoxVECloudConfig::try_new(Path::new("tests/fixtures/proxmoxve/dhcp"))
+        .expect("cannot parse config");
+
+    assert_eq!(
```
pothos, line 60: The resulting config does not work because afterburn doesn't generate a `DHCP=yes` entry to actually enable DHCP. This would require a new setting in `network::Interface` to render it in the `pub fn config` function of `src/network.rs`.

abuisine:
> From the Flatcar side we would also be interested to have this. I don't know if there is any specification about config drive contents from proxmox, but it looks like the `user-data` at least here is expected to be cloud-config to extract the ssh key and hostname. This conflicts with Ignition. Could this get moved to `meta-data`? I don't have a clue on what API proxmox provides to set this up. If none, then I guess we have to live with no ssh and hostname support in afterburn and let the user specify this in Ignition.
> 
> One more note about network units: This seems to be netplan - do you want to focus on a subset or support the full netplan spec? Recently I added a `--netplan-config` flag which only has the scope of copying this to, e.g., `/run/netplan/` and then we could run netplan to convert it to NetworkManager/networkd units.

We should find a way to support ignition in `user_data` but we should keep at the same time compatibility with "cloud-init settings" provided through the proxmox API.
I guess there is a way to have afterburn consume `hostname` and `ssh keys` if there are present, with a fallback to ignition if `user_data` is actually an ignition json.

jlebon:
According to the discussions in https://github.com/coreos/fedora-coreos-tracker/issues/1652 and https://github.com/coreos/fedora-coreos-tracker/issues/736, it's possible to provide an Ignition config but currently quite awkward to do.

I think this path is probably the best compromise, though maybe let's discuss in https://github.com/coreos/fedora-coreos-tracker/issues/1652 since that's where all the context currently is?

jlebon:
Still need to review this, but can we add a new docs page (e.g. under https://coreos.github.io/afterburn/usage/) to describe the situation on this platform a bit more? Basically:
- we're parsing the cloud-init config provided by Proxmox
- we don't generically support custom cloud-init data
- we no-op if the data isn't cloud-init
- we only support specific keys from the cloud-init config
- any limitations re. `network-config` that you may think of as it relates to systemd-networkd

pothos:
FYI: I don't think we can use this in Flatcar when it relies on cloud-init userdata because Flatcar will run coreos-cloudinit anyway if it doesn't see Ignition user-data and then we would have it processed twice. For the network config that can be quite confusing because the afterburn network unit would only support a subset.

abuisine:
> FYI: I don't think we can use this in Flatcar when it relies on cloud-init userdata because Flatcar will run coreos-cloudinit anyway if it doesn't see Ignition user-data and then we would have it processed twice. For the network config that can be quite confusing because the afterburn network unit would only support a subset.

I will let @arcln answer for coreos-cloudinit and focus my answer on the network side of things.

Proxmoxve has a pretty limited Software Defined Networking (since the v7.4 as far as I can remember). To my knowledge, this is not widely used. This will probably improve in the future.
Most users today :
1) either build a standalone network with DHCP server (not provided by PromoxVE)
2) or provide static ip configuration through the cloud-init mechanism, or any other medium.


For this second option, on the ignition side of things ... my understanding is that network should be (must be ?) configured prior to ignition being launched. I remember having a conversation on systemd-sysext extensions for flatcar where an http url was used. Ignition was processing it so early that it was done before static network configuration. Systemd-sysext where therefore not installable from a network repo through Ignition.
We finally had to add kernel command lines in order to configure the network before ... which is not very practical to say the least.

A better option may be:
The network part of the cloud-config is applied through Afterburn (or another mechanism) prior to Ignition being run.
This PR includes network configuration for this reason.

It is not clear to me if Flatcar fully endorse Afterburn, could you confirm ?

pothos:
The static IP addr setup through the kernel args inside Ignition is only needed if Ignition should download something. Using afterburn for that would only be possible if there was support for proper metadata at Proxmox VE. With the proposed approach it wouldn't work to use afterburn because the userdata is Ignition JSON then ;)

If you have a chance of getting into discussions with Proxmox to define metadata separate of userdata, that would be nice. Even if it would be [cloud-config-based metadata](https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html#walkthrough-of-guestinfo-keys-transport) as done for VMware, because there one can use the netplan CLI option in afterburn to extract this (A supported afterburn subset might also be ok as alternative) and one could then get initrd networking that way while using Ignition userdata.

abuisine:
> The static IP addr setup through the kernel args inside Ignition is only needed if Ignition should download something. Using afterburn for that would only be possible if there was support for proper metadata at Proxmox VE. With the proposed approach it wouldn't work to use afterburn because the userdata is Ignition JSON then ;)
> 
> If you have a chance of getting into discussions with Proxmox to define metadata separate of userdata, that would be nice. Even if it would be [cloud-config-based metadata](https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html#walkthrough-of-guestinfo-keys-transport) as done for VMware, because there one can use the netplan CLI option in afterburn to extract this (A supported afterburn subset might also be ok as alternative) and one could then get initrd networking that way while using Ignition userdata.

I do not specifically agree, let me precise my throughts :
The iso generated by ProxmoxVE contains multiple files including userdata AND network-config, so you can have ignition content in userdata and network settings from ui or api in network-config. This would work ... and is actually exactly what we are trying to achieve. This PR consumes the network-config file.

Also some users would override userdata with ignition configuration, but some others won't. Again this PR should work ok in both cases.

bitdriftr:
From my understanding, `coreos-cloudinit` is only executed [if the OEM ID matches a white list](https://github.com/flatcar/init/blob/96e929df6c04df58a87187cf2a769923e5ecdf41/systemd/system/oem-cloudinit.service#L7). In our case the OEM ID is `proxmoxve` so it shouldn't run.

Also `coreos-cloudinit` may not read cloud init config from Proxmox VE because [it looks for a file called `user_data`](https://github.com/coreos/coreos-cloudinit/blob/f1f0405491dfd073bbf074f7e374c9ef85600691/datasource/configdrive/configdrive.go#L80) instead of `user-data`.

pothos:
> The iso generated by ProxmoxVE contains multiple files including userdata AND network-config

Ah, right, it has two separate files and already roughly follows https://cloudinit.readthedocs.io/en/latest/reference/datasources/vmware.html#walkthrough-of-guestinfo-keys-transport similar to the case with the special vmware guestinfo key. One could ask proxmox to set `local-hostname:` (or `instance-id:`?)  in the `network-config` so that this would work from Afterburn if Ignition user-data is used. Only ssh keys would then have to be put into Ignition user-data.

> Also `coreos-cloudinit` may not read cloud init config from Proxmox VE because [it looks for a file called `user_data`](https://github.com/coreos/coreos-cloudinit/blob/f1f0405491dfd073bbf074f7e374c9ef85600691/datasource/configdrive/configdrive.go#L80) instead of `user-data`.

Thanks for checking that, yes, the config drive mechanism will be triggered on Flatcar if not explicitly turned off. We could decide to stick to your afterburn support instead of handling this in coreos-cloudinit and thus say that coreos-cloudinit won't be supported on this platform. What I think would be a compromise for Flatcar is to use afterburn for the `network-config` file and leave the `user-data` file to either Ignition or coreos-cloudinit.

Anyway, since afterburn is always opt-in for distros it won't hurt to merge the PR (after the `DHCP=yes` case is fixed). Note that FCOS doesn't use networkd and thus it would be nice to also wire up the netplan flag here in case someone wants to use afterburn for network setup later, see https://github.com/coreos/afterburn/blob/f918f4c45066265283f71577ee7eabc8ed05cc43/src/providers/vmware/amd64.rs#L70

src/providers/proxmoxve/tests.rs
bitdriftr: I am not sure to understand what the `DHCP=yes` actually does. I successfully spawned VMs with Internet access using this afterburn and the exact config you'll find in `tests/fixtures/proxmoxve/dhcp`. Can you please elaborate on why this is required? Thanks.

bitdriftr:
> Note that FCOS doesn't use networkd and thus it would be nice to also wire up the netplan flag here in case someone wants to use afterburn for network setup later

We agree that netplan support is important for Afterburn to be working on various OS. I will work on that in the next few days.

src/providers/proxmoxve/tests.rs
pothos: Thanks for having a look. I'll also test it again and try to come up with a small diff.

pothos: The `DHCP=` setting default to `no` and thus we have to set it to `yes` or actually `ipv4` to use DHCP. I tested this on a Flatcar VM. In your testing you might have used a wrong MAC address and thus the default Flatcar config got used which is to do DHCP v4+v6.
Here is a patch `0001-proxmoxve-Generate-proper-network-unit-for-the-DHCP-.patch`. It should apply to your branch with `git am -3 FILE` but tell if you need help.

```
From 57150985ad6fc47d03ccec1eec377377859267fb Mon Sep 17 00:00:00 2001
From: Kai Lueke <kailuke@microsoft.com>
Date: Mon, 6 May 2024 17:20:17 +0900
Subject: [PATCH] proxmoxve: Generate proper network unit for the DHCP case

When DHCP is not specified in a systemd network unit, the default is
"no". For a "type: dhcp4" entry in the Proxmox VE network-config file
the systemd network unit should set DHCP=ipv4, and similar for "dhcp6".
---
 src/network.rs                         | 62 ++++++++++++++++++++++++++
 src/providers/digitalocean/mod.rs      |  1 +
 src/providers/ibmcloud_classic/mod.rs  |  1 +
 src/providers/packet/mod.rs            |  3 ++
 src/providers/proxmoxve/cloudconfig.rs | 10 ++++-
 src/providers/proxmoxve/tests.rs       |  5 ++-
 6 files changed, 80 insertions(+), 2 deletions(-)

diff --git a/src/network.rs b/src/network.rs
index 7795735..26c0a6f 100644
--- a/src/network.rs
+++ b/src/network.rs
@@ -79,6 +79,8 @@ pub struct Interface {
     pub priority: u8,
     pub nameservers: Vec<IpAddr>,
     pub ip_addresses: Vec<IpNetwork>,
+    // Optionally enable DHCP
+    pub dhcp: Option<DhcpSetting>,
     pub routes: Vec<NetworkRoute>,
     pub bond: Option<String>,
     pub unmanaged: bool,
@@ -128,6 +130,31 @@ impl NetDevKind {
     }
 }
 
+/// Optional use of DHCP.
+#[allow(dead_code)]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum DhcpSetting {
+    Both,
+    V4,
+    V6,
+}
+
+impl DhcpSetting {
+    /// Return DHCP setting according to `systemd.network`
+    ///
+    /// See [systemd documentation](dhcp) for the full list.
+    ///
+    /// dhcp: https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html#DHCP=
+    fn sd_dhcp_setting(&self) -> String {
+        let setting = match *self {
+            DhcpSetting::Both => "yes",
+            DhcpSetting::V4 => "ipv4",
+            DhcpSetting::V6 => "ipv6",
+        };
+        setting.to_string()
+    }
+}
+
 impl Interface {
     /// Return a deterministic `systemd.network` unit name for this device.
     pub fn sd_network_unit_name(&self) -> Result<String> {
@@ -158,6 +185,9 @@ impl Interface {
 
         // [Network] section
         writeln!(config, "\n[Network]").unwrap();
+        if let Some(dhcp) = &self.dhcp {
+            writeln!(config, "DHCP={}", dhcp.sd_dhcp_setting()).unwrap();
+        }
         for ns in &self.nameservers {
             writeln!(config, "DNS={ns}").unwrap()
         }
@@ -246,6 +276,7 @@ mod tests {
                     priority: 20,
                     nameservers: vec![],
                     ip_addresses: vec![],
+                    dhcp: None,
                     routes: vec![],
                     bond: None,
                     unmanaged: false,
@@ -261,6 +292,7 @@ mod tests {
                     priority: 10,
                     nameservers: vec![],
                     ip_addresses: vec![],
+                    dhcp: None,
                     routes: vec![],
                     bond: None,
                     unmanaged: false,
@@ -276,6 +308,7 @@ mod tests {
                     priority: 20,
                     nameservers: vec![],
                     ip_addresses: vec![],
+                    dhcp: None,
                     routes: vec![],
                     bond: None,
                     unmanaged: false,
@@ -291,6 +324,7 @@ mod tests {
                     priority: 20,
                     nameservers: vec![],
                     ip_addresses: vec![],
+                    dhcp: None,
                     routes: vec![],
                     bond: None,
                     unmanaged: false,
@@ -306,6 +340,7 @@ mod tests {
                     priority: 20,
                     nameservers: vec![],
                     ip_addresses: vec![],
+                    dhcp: None,
                     routes: vec![],
                     bond: None,
                     unmanaged: false,
@@ -330,6 +365,7 @@ mod tests {
             priority: 20,
             nameservers: vec![],
             ip_addresses: vec![],
+            dhcp: None,
             routes: vec![],
             bond: None,
             unmanaged: false,
@@ -387,6 +423,7 @@ mod tests {
                             Ipv6Network::new(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1), 128).unwrap(),
                         ),
                     ],
+                    dhcp: None,
                     routes: vec![NetworkRoute {
                         destination: IpNetwork::V4(
                             Ipv4Network::new(Ipv4Addr::new(127, 0, 0, 1), 8).unwrap(),
@@ -428,6 +465,7 @@ Gateway=127.0.0.1
                     priority: 10,
                     nameservers: vec![],
                     ip_addresses: vec![],
+                    dhcp: None,
                     routes: vec![],
                     bond: None,
                     unmanaged: false,
@@ -447,6 +485,7 @@ Gateway=127.0.0.1
                     priority: 10,
                     nameservers: vec![],
                     ip_addresses: vec![],
+                    dhcp: None,
                     routes: vec![],
                     bond: None,
                     unmanaged: false,
@@ -470,6 +509,7 @@ RequiredForOnline=no
                     priority: 10,
                     nameservers: vec![],
                     ip_addresses: vec![],
+                    dhcp: None,
                     routes: vec![],
                     bond: None,
                     unmanaged: true,
@@ -482,6 +522,28 @@ Name=*
 
 [Link]
 Unmanaged=yes
+",
+            ),
+            // test the DHCP setting
+            (
+                Interface {
+                    name: Some("*".to_owned()),
+                    mac_address: None,
+                    path: None,
+                    priority: 10,
+                    nameservers: vec![],
+                    ip_addresses: vec![],
+                    dhcp: Some(DhcpSetting::V4),
+                    routes: vec![],
+                    bond: None,
+                    unmanaged: false,
+                    required_for_online: None,
+                },
+                "[Match]
+Name=*
+
+[Network]
+DHCP=ipv4
 ",
             ),
         ];
diff --git a/src/providers/digitalocean/mod.rs b/src/providers/digitalocean/mod.rs
index 2583696..dd817dc 100644
--- a/src/providers/digitalocean/mod.rs
+++ b/src/providers/digitalocean/mod.rs
@@ -156,6 +156,7 @@ impl DigitalOceanProvider {
                     mac_address: Some(mac),
                     nameservers: self.dns.nameservers.clone(),
                     ip_addresses: addrs,
+                    dhcp: None,
                     routes,
                     bond: None,
                     name: None,
diff --git a/src/providers/ibmcloud_classic/mod.rs b/src/providers/ibmcloud_classic/mod.rs
index 6546f08..d210aca 100644
--- a/src/providers/ibmcloud_classic/mod.rs
+++ b/src/providers/ibmcloud_classic/mod.rs
@@ -247,6 +247,7 @@ impl IBMClassicProvider {
                 priority: 10,
                 nameservers: nameservers.clone(),
                 ip_addresses: vec![ip_net],
+                dhcp: None,
                 routes,
                 bond: None,
                 unmanaged: false,
diff --git a/src/providers/packet/mod.rs b/src/providers/packet/mod.rs
index 314c642..4a8ec35 100644
--- a/src/providers/packet/mod.rs
+++ b/src/providers/packet/mod.rs
@@ -216,6 +216,7 @@ impl PacketProvider {
                 priority: 10,
                 nameservers: Vec::new(),
                 ip_addresses: Vec::new(),
+                dhcp: None,
                 routes: Vec::new(),
                 // the interface should be unmanaged if it doesn't have a bond
                 // section
@@ -241,6 +242,7 @@ impl PacketProvider {
                     path: None,
                     bond: None,
                     ip_addresses: Vec::new(),
+                    dhcp: None,
                     routes: Vec::new(),
                     unmanaged: false,
                     required_for_online: Some("degraded-carrier".to_owned()),
@@ -334,6 +336,7 @@ impl PacketProvider {
             bond: None,
             nameservers: Vec::new(),
             ip_addresses: Vec::new(),
+            dhcp: None,
             routes: Vec::new(),
             required_for_online: None,
         };
diff --git a/src/providers/proxmoxve/cloudconfig.rs b/src/providers/proxmoxve/cloudconfig.rs
index a213e50..50bc42c 100644
--- a/src/providers/proxmoxve/cloudconfig.rs
+++ b/src/providers/proxmoxve/cloudconfig.rs
@@ -1,5 +1,5 @@
 use crate::{
-    network::{self, NetworkRoute},
+    network::{self, DhcpSetting, NetworkRoute},
     providers::MetadataProvider,
 };
 use anyhow::Result;
@@ -207,6 +207,8 @@ impl ProxmoxVECloudNetworkConfigEntry {
             ip_addresses: vec![],
             // filled below
             routes: vec![],
+            // filled below
+            dhcp: None,
             // filled below because Option::try_map doesn't exist yet
             mac_address: None,
 
@@ -257,6 +259,12 @@ impl ProxmoxVECloudNetworkConfigEntry {
                 }
             }
 
+            if subnet.subnet_type == "dhcp" || subnet.subnet_type == "dhcp4" {
+                iface.dhcp = Some(DhcpSetting::V4)
+            }
+            if subnet.subnet_type == "dhcp6" {
+                iface.dhcp = Some(DhcpSetting::V6)
+            }
             if subnet.subnet_type == "ipv6_slaac" {
                 warn!("subnet type \"ipv6_slaac\" not supported, ignoring");
             }
diff --git a/src/providers/proxmoxve/tests.rs b/src/providers/proxmoxve/tests.rs
index ddef64b..e4171fe 100644
--- a/src/providers/proxmoxve/tests.rs
+++ b/src/providers/proxmoxve/tests.rs
@@ -1,6 +1,6 @@
 use super::ProxmoxVECloudConfig;
 use crate::{
-    network::{self, NetworkRoute},
+    network::{self, DhcpSetting, NetworkRoute},
     providers::MetadataProvider,
 };
 use ipnetwork::IpNetwork;
@@ -64,6 +64,7 @@ fn test_network_dhcp() {
             mac_address: Some(MacAddr::from_str("01:23:45:67:89:00").unwrap()),
             path: None,
             priority: 20,
+            dhcp: Some(DhcpSetting::V4),
             nameservers: vec![
                 IpAddr::from_str("1.1.1.1").unwrap(),
                 IpAddr::from_str("8.8.8.8").unwrap()
@@ -98,6 +99,7 @@ fn test_network_static() {
                     IpNetwork::from_str("192.168.1.1/24").unwrap(),
                     IpNetwork::from_str("2001:0db8:85a3:0000:0000:8a2e:0370:0/24").unwrap(),
                 ],
+                dhcp: None,
                 routes: vec![
                     NetworkRoute {
                         destination: IpNetwork::from_str("0.0.0.0/0").unwrap(),
@@ -123,6 +125,7 @@ fn test_network_static() {
                     IpNetwork::from_str("192.168.42.1/24").unwrap(),
                     IpNetwork::from_str("2001:0db8:85a3:0000:0000:8a2e:4242:0/24").unwrap(),
                 ],
+                dhcp: None,
                 routes: vec![
                     NetworkRoute {
                         destination: IpNetwork::from_str("0.0.0.0/0").unwrap(),
-- 
2.45.0
```

src/providers/proxmoxve/configdrive.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Debug)]
+pub struct ProxmoxVEConfigDrive {
+    mount_path: PathBuf,
+    config: ProxmoxVECloudConfig,
+}
+
+impl ProxmoxVEConfigDrive {
+    pub fn try_new() -> Result<Self> {
+        const CONFIG_DRIVE_LABEL: &str = "cidata";
+        const TARGET_FS: &str = "iso9660";
```
jlebon: Let's just hoist this to the toplevel of the module.

src/providers/proxmoxve/configdrive.rs
```diff
@@ -0,0 +1,66 @@
...
+            .context("failed to create temporary directory")?;
+
+        crate::util::mount_ro(
+            &Path::new("/dev/disk/by-label/").join(CONFIG_DRIVE_LABEL),
+            target.path(),
+            TARGET_FS,
+            3,
+        )?;
+
+        let mount_path = target.path().to_owned();
```
jlebon: Hmm, I think we need to transfer the `TempDir` here into `ProxmoxVEConfigDrive`. Otherwise... well I presume what happens is that once `target` goes out of scope, it tries to clean itself up but can't because of the mount. So then we'll leak the tempdir once `ProxmoxVEConfigDrive` is dropped.

src/providers/proxmoxve/cloudconfig.rs
```diff
@@ -0,0 +1,279 @@
...
+        let user_data_file = BufReader::new(File::open(path.join("user-data"))?);
+        let mut user_data = None;
+
+        if let Some(first_line) = user_data_file.lines().next() {
+            if let Ok(first_line) = first_line {
+                if first_line.starts_with("#cloud-config") {
+                    user_data = serde_yaml::from_reader(File::open(path.join("user-data"))?)?;
+                }
+            }
+        }
```
jlebon: Hmm, opening the file twice looks a bit clunky. Why not just read in the whole file, and then parse it as YAML only if it starts with `#cloud-config\n`?

src/providers/proxmoxve/cloudconfig.rs
```diff
@@ -0,0 +1,279 @@
...
+}
+
+impl MetadataProvider for ProxmoxVECloudConfig {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        let mut out = HashMap::new();
+
+        out.insert(
+            "PROXMOXVE_HOSTNAME".to_owned(),
+            self.hostname()?.unwrap_or_default(),
+        );
```
jlebon: If no hostname is set, I would expect no `PROXMOXVE_HOSTNAME`, not an empty one.

src/providers/proxmoxve/tests.rs
jlebon: It seems like this is resolved?

jlebon (commented):
Some comments, but this looks sane overall.

bitdriftr:
Everything looks good to me, I think this can be merged.

@pothos asked for netplan support. I implemented it, should I push to this PR ?

jlebon:
Can you squash it down to one commit and also address the lints from CI?

> @pothos asked for netplan support. I implemented it, should I push to this PR ?

Let's do that in a follow-up PR? We'll want to add docs as well for this (see the similar VMware-related netplan support).

bitdriftr:
Fixed clippy errors and squashed commits.

I'll open a netplan support PR after this one is merged.

jlebon (approved):

jlebon:
Hmm, Jenkins is having trouble merging the release notes for testing:

```
[2024-06-19T17:06:21.573Z] Also:   org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: b18e670c-7d20-4916-b2af-bda4ac6213a3
[2024-06-19T17:06:21.573Z] hudson.plugins.git.GitException: Command "git merge a9a70f2ee7d96f90aeaa7c2ca96fc8a5e2df7834" returned status code 1:
[2024-06-19T17:06:21.573Z] stdout: Auto-merging src/providers/ibmcloud_classic/mod.rs
[2024-06-19T17:06:21.573Z] Auto-merging docs/release-notes.md
[2024-06-19T17:06:21.573Z] CONFLICT (content): Merge conflict in docs/release-notes.md
[2024-06-19T17:06:21.573Z] Automatic merge failed; fix conflicts and then commit the result.
```

Can you rebase onto the latest main?

jlebon:
(I tried rebasing it myself, but it looks like that option to allow maintainers to push was disabled.)

bitdriftr:
done

jlebon (approved):

abuisine:
Cool ! We will work on the netplan PR + focus on the ignition and flatcar side of things :)

PR #1062, opened 2024-04-01
providers: Add "akamai" provider

nesv:
The "akamai" provider adds support for retrieving configuration from Akamai Connected Cloud's (a.k.a. Linode) [Metadata Service][metadata-service].

References: https://github.com/flatcar/Flatcar/issues/1404
References: https://github.com/coreos/fedora-coreos-tracker/issues/1701
References: https://github.com/coreos/ignition/pull/1841

## Note

I'll push any requested changes as fixup commits, just so that the Github comment threads don't get messed up. If this PR is approved, I will squash these fixup commits out prior to merge.

[metadata-service]: https://www.linode.com/docs/products/compute/compute-instances/guides/metadata/


prestist:
@nesv Awesome, thank you for doing this!

Took a quick look at it, and overall it looks sane, I need to do some double checking on some of the meat of the mod.rs but so far so good!

nesv:
> @nesv Awesome, thank you for doing this!
> 
> Took a quick look at it, and overall it looks sane, I need to do some double checking on some of the meat of the mod.rs but so far so good!

Thanks a bunch, @prestist!

I misunderstood how serde handles deserializing untagged enums. I had hoped I could use an enum to enumerate all possible _values_ for a field, but it would seem that isn't possible; I have removed the `InterfacePurpose` enum, in favour of an `Option<String>`, but I am not too enthused with that tradeoff. It feels like a loss in precision, though it does keep things flexible should the API change in the future.

nesv:
I was hoping to provide an as-complete-as-possible `MetadataProvider` implementation, but I don't think there is enough networking information that comes from Linode's Metadata Service to be able to implement the `MetadataProvider::networks` method.

nesv:
The latest patches have been checked, and they work! All of the fixup commits have been squashed and rebased onto `main`.

tormath1 (approved):
FWIW, same as the Ignition PR (https://github.com/coreos/ignition/pull/1841) it has been tested on Flatcar and it works as expected on Akamai/Linode from a functional aspect.

src/providers/akamai/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+const TOKEN_URL: &'static str = "http://169.254.169.254/v1/token";
+const INSTANCE_METADATA_URL: &'static str = "http://169.254.169.254/v1/instance";
+const NETWORK_METADATA_URL: &'static str = "http://169.254.169.254/v1/network";
+const SSHKEYS_URL: &'static str = "http://169.254.169.254/v1/ssh-keys";
+
+static HDR_TOKEN_EXPIRY: &str = "metadata-token-expiry-seconds";
+static HDR_TOKEN: &str = "metadata-token";
+
+static TOKEN_TTL: &str = "3600";
```
jlebon: Let's match Ignition and cloud-init here too and use 300?

src/providers/akamai/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+use serde::Deserialize;
+use std::collections::HashMap;
+
+use crate::providers::MetadataProvider;
+use crate::retry;
+
+const TOKEN_URL: &'static str = "http://169.254.169.254/v1/token";
+const INSTANCE_METADATA_URL: &'static str = "http://169.254.169.254/v1/instance";
+const NETWORK_METADATA_URL: &'static str = "http://169.254.169.254/v1/network";
+const SSHKEYS_URL: &'static str = "http://169.254.169.254/v1/ssh-keys";
```
jlebon: Minor: this could be DRYer. See e.g. `endpoint_for()` in `aws/mod.rs`.

src/providers/akamai/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    fn token(&self) -> Result<HeaderValue> {
+        let token: String = self
+            .client
+            .put(retry::Raw, TOKEN_URL.to_string(), None)
+            .header(
+                HeaderName::from_static(HDR_TOKEN_EXPIRY),
+                HeaderValue::from_static(TOKEN_TTL),
+            )
+            .dispatch_put()?
+            .ok_or_else(|| anyhow!("get metadata token"))?;
```
jlebon:
```suggestion
            .context("get metadata token")?;
```

src/providers/akamai/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    fn fetch_instance_metadata(&self) -> Result<Instance> {
+        let token = self.token()?;
+        let instance: Instance = self
+            .client
+            .get(retry::Json, INSTANCE_METADATA_URL.to_string())
+            .header(HeaderName::from_static(HDR_TOKEN), token)
+            .header(
+                HeaderName::from_static("accept"),
+                HeaderValue::from_static("application/json"),
+            )
```
jlebon: Since we basically always specify this, we could also just get a token and add the headers on the client at construction time.

Fine as is too.

src/providers/akamai/mod.rs
jlebon: And similarly in the other functions that currently return static string errors.

src/providers/akamai/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        let ssh_keys: SshKeys = self
+            .client
+            .get(retry::Json, SSHKEYS_URL.to_string())
+            .header(HeaderName::from_static(HDR_TOKEN), self.token()?)
+            .header(
+                HeaderName::from_static("accept"),
+                HeaderValue::from_static("application/json"),
+            )
+            .send()?
+            .ok_or_else(|| anyhow!("ssh keys not found"))?;
```
jlebon: Is this always guaranteed to return something that we can deserialize into a `HashMap` even if no SSH keys were configured? (I think basically that boils down to, does it return `{}` if there are no SSH keys?)

src/providers/akamai/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+        // Compute the capacity of the Vec for holding all of the network attributes.
+        // The +2 is for the IPv6 SLAAC and link-local addresses.
+        let capacity: usize = data.ipv4.public.len()
+            + data.ipv4.private.len()
+            + data.ipv4.shared.len()
+            + 2
+            + data.ipv6.ranges.len()
+            + data.ipv6.shared_ranges.len();
+        let mut net_attrs: Vec<(String, String)> = Vec::with_capacity(capacity);
```
jlebon: I don't follow the reasoning behind having a different `Vec` instead of directly pushing to `attrs`.

src/providers/akamai/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+
+impl MetadataProvider for AkamaiProvider {
+    fn attributes(&self) -> Result<HashMap<String, String>> {
+        let attrs = self.parse_attrs()?;
+        Ok(attrs.into_iter().collect())
+    }
+
+    fn hostname(&self) -> Result<Option<String>> {
+        let data = self.fetch_instance_metadata()?;
+        Ok(Some(data.label.clone()))
```
jlebon: Does this match what cloud-init does? `label` sounds like a loose attribute name for a hostname.

src/providers/akamai/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        let ssh_keys = self.fetch_ssh_keys()?;
+        let all_keys: Vec<String> = ssh_keys.users.into_values().flatten().collect();
+
+        let mut public_keys: Vec<PublicKey> = Vec::with_capacity(all_keys.len());
+        for k in all_keys {
+            let key = PublicKey::parse(&k)?;
+            public_keys.push(key);
+        }
+
+        Ok(public_keys)
```
jlebon:
```suggestion
        Ok(self
            .fetch_ssh_keys()?
            .users
            .values()
            .flatten()
            .map(|k| PublicKey::parse(k))
            .collect::<Result<_, _>>()?)
```

src/providers/akamai/mod.rs
nesv: I'm sure it made sense to me at the time I wrote it, but I'll be damned if I can remember why I did that. :laughing: 

I'll just directly push to `attrs`.

src/providers/akamai/mod.rs
nesv: If no SSH keys are provided through the Metadata service, the result is: 

```json
{"users": {}}
```

so I think we would need the `SshKeys` struct around, at the very least, for deserialization.

Instead of returning the `SshKeys` struct, would you prefer to have `fetch_ssh_keys` return a `HashMap<String, Vec<String>>`?

jlebon: No, that's fine. Just wanted to make sure we were handling the no-SSH-keys case correctly.

src/providers/akamai/mod.rs
nesv: Ah, thank you for pointing this out. I will probably just remove this method implementation, and rely on the default one that comes with the trait.

Currently, cloud-init does not automatically set a user name on Linode images that support the metadata service; logging in to a fresh Linode that supports metadata, has a hostname of `localhost`, unless you override it with a config passed in through userdata.

> `label` sounds like a loose attribute name for a hostname.

I agree. What Linode calls a "label", other providers commonly refer to as a "name". It must be unique across all instances in an account. However, the label _can_ be changed after the instance has been created.

I will rip this method out.

src/providers/akamai/mod.rs
nesv: Okay. I opted to use a `HashMap<String, Vec<String>>` anyways, since it was fewer custom types to carry around after deserialization.

nesv:
@jlebon Requested changes have been made! Thank you for taking the time to review this PR.

jlebon (commented):
LGTM overall, though looks like there are a few lints to address. While we're here, can you squash your fixup commits together? We don't "Squash and merge" in this repo.

nesv:
> LGTM overall, though looks like there are a few lints to address.

Gah, oversights on my part. I'll get those fixed up.

> While we're here, can you squash your fixup commits together? We don't "Squash and merge" in this repo.

Absolutely! I was planning on squashing my own fixup commits out anyways. :smile:

nesv:
@jlebon Fixup commits squashed, and I've applied the linting suggestions.

jlebon (approved):

PR #1074, opened 2024-05-08
microsoft/azure: allow empty certificate chain in PKCS12 file

### LINKED ISSUE #1553: `coreos.ignition.ssh.key` failing in `kola-azure`
`coreos.ignition.ssh.key` has been failing in the `kola-azure` jenkins job since August 15, 2023 with the following error:
```
[2023-08-17T15:08:35.184Z] --- FAIL: coreos.ignition.ssh.key (331.60s)
[2023-08-17T15:08:35.184Z]         harness.go:1789: mach.Start() failed: machine "kola-975f064f-cccec36d80" failed basic checks: detected failed or stuck systemd units
```
Looking into the journal log file for the test shows:
[journal.txt](https://github.com/coreos/fedora-coreos-tracker/files/12411622/journal.txt)
[console.txt](https://github.com/coreos/fedora-coreos-tracker/files/12411627/console.txt)

```
Aug 17 19:46:17.528507 afterburn[1523]: Aug 17 19:46:17.528 INFO Fetching http://168.63.129.16/?comp=versions: Attempt #1
Aug 17 19:46:17.552714 afterburn[1523]: Aug 17 19:46:17.552 INFO Fetch successful
Aug 17 19:46:17.569624 afterburn[1523]: Aug 17 19:46:17.569 INFO Fetching http://168.63.129.16/machine/?comp=goalstate: Attempt #1
Aug 17 19:46:17.574856 afterburn[1523]: Aug 17 19:46:17.574 INFO Fetch successful
Aug 17 19:46:17.883244 afterburn[1523]: Aug 17 19:46:17.883 INFO Fetching http://168.63.129.16/machine/5c91c37d-f891-4428-93c9-81e602a91a74/a1fc44fd%2Dc611%2D4ea8%2Dbc6b%2D678f2ba3b71e.%5Fkola%2Dc954503e%2D36b448701c?comp=certificates&incarnation=1: Attempt #1
Aug 17 19:46:17.898618 afterburn[1523]: Aug 17 19:46:17.898 INFO Fetch successful
Aug 17 19:46:17.927929 afterburn[1523]: Error: failed to run
Aug 17 19:46:17.927929 afterburn[1523]: Caused by:
Aug 17 19:46:17.927929 afterburn[1523]:     0: writing ssh keys
Aug 17 19:46:17.927929 afterburn[1523]:     1: failed to convert pkcs12 blob to ssh pubkey
Aug 17 19:46:17.927929 afterburn[1523]:     2: failed to get chain from pkcs12
Aug 17 19:46:17.929015 systemd[1]: afterburn-sshkeys@core.service: Main process exited, code=exited, status=1/FAILURE
Aug 17 19:46:17.929193 systemd[1]: afterburn-sshkeys@core.service: Failed with result 'exit-code'.
Aug 17 19:46:17.934176 systemd[1]: Failed to start afterburn-sshkeys@core.service - Afterburn (SSH Keys).
Aug 17 19:46:17.942790 systemd[1]: Dependency failed for afterburn-sshkeys.target - Synchronize afterburn-sshkeys@.service template instances.
Aug 17 19:46:17.947667 systemd[1]: Dependency failed for multi-user.target - Multi-User System.
Aug 17 19:46:17.951000 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=afterburn-sshkeys@core comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'
Aug 17 19:46:17.952533 systemd[1]: multi-user.target: Job multi-user.target/start failed with result 'dependency'.
Aug 17 19:46:17.952960 kernel: audit: type=1130 audit(1692301577.951:80): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=afterburn-sshkeys@core comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'
Aug 17 19:46:17.952564 systemd[1]: afterburn-sshkeys.target: Job afterburn-sshkeys.target/start failed with result 'dependency'.
```
Afterburn fails to [convert pkcs12 blob to ssh pubkey](https://github.com/coreos/afterburn/blob/06a80b6edbe2bde448073b9b7f1b171a2507cd65/src/providers/microsoft/azure/mod.rs#L275)
Before this failure, the fetching section looked like this:
```
Aug 14 00:49:48.769452 afterburn[1511]: Aug 14 00:49:48.769 INFO Fetching http://168.63.129.16/?comp=versions: Attempt #1    
Aug 14 00:49:48.790498 afterburn[1511]: Aug 14 00:49:48.790 INFO Fetch successful                                            
Aug 14 00:49:48.809532 afterburn[1511]: Aug 14 00:49:48.809 INFO Fetching http://168.63.129.16/machine/?comp=goalstate: Attem
Aug 14 00:49:48.819160 afterburn[1511]: Aug 14 00:49:48.818 INFO Fetch successful                                            
Aug 14 00:49:48.819462 afterburn[1511]: Aug 14 00:49:48.819 WARN SSH pubkeys requested, but not provisioned for this instance
```
Which is what other test's logs look like when an SSH key IS set. So it's looking like an SSH key is being set somehow, even though one isnt specified. This is causing the `.ssh/authorized_keys.d/afterburn` file to be created and the kola test fails.

The certificates endpoint is now included in `http://168.63.129.16/machine/?comp=goalstate` so now it's being fetched and afterburn is failing directly after [fetching the certificate](https://github.com/coreos/afterburn/blob/06a80b6edbe2bde448073b9b7f1b171a2507cd65/src/providers/microsoft/azure/mod.rs#L266).

marmijo:
Azure is populating the certs endpoint even when an SSH key is not
provided. The certificate chain in the PKCS12 file is empty in this
case causing afterburn to fail with the current logic. Check if the
certificate chain is empty before retrieving the SSH public key
and handle this case as a valid option.

The more general `crypto/mod.rs` was updated here which also affected
the azurestack code. Update azurestack to accommodate the changes when
retrieving the SSH public key.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1553

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -13,6 +13,7 @@ Major changes:
 Minor changes:
 
 - Hetzner: fix duplicate attribute prefix
+- Fix Azure SSH key fetching when no key provisioned
 
 Packaging changes:
 
--- a/src/providers/microsoft/azure/mod.rs
+++ b/src/providers/microsoft/azure/mod.rs
@@ -251,7 +251,7 @@ impl Azure {
     }
 
     // put it all together
-    fn get_ssh_pubkey(&self, certs_endpoint: String) -> Result<PublicKey> {
+    fn get_ssh_pubkey(&self, certs_endpoint: String) -> Result<Option<PublicKey>> {
         // we have to generate the rsa public/private keypair and the x509 cert
         // that we use to make the request. this is equivalent to
         // `openssl req -x509 -nodes -subj /CN=LinuxTransport -days 365 -newkey rsa:2048 -keyout private.pem -out cert.pem`
@@ -271,10 +271,7 @@ impl Azure {
             .context("failed to decrypt cms blob")?;
 
         // convert that to the OpenSSH public key format
-        let ssh_pubkey = crypto::p12_to_ssh_pubkey(&p12)
-            .context("failed to convert pkcs12 blob to ssh pubkey")?;
-
-        Ok(ssh_pubkey)
+        crypto::p12_to_ssh_pubkey(&p12).context("failed to convert pkcs12 blob to ssh pubkey")
     }
 
     #[cfg(test)]
@@ -402,18 +399,17 @@ impl MetadataProvider for Azure {
         let goalstate = self.fetch_goalstate()?;
         let certs_endpoint = match goalstate.certs_endpoint() {
             Some(ep) => ep,
-            None => {
-                warn!("SSH pubkeys requested, but not provisioned for this instance");
-                return Ok(vec![]);
-            }
+            None => return Ok(vec![]),
         };
 
         if certs_endpoint.is_empty() {
             bail!("unexpected empty certificates endpoint");
         }
 
-        let key = self.get_ssh_pubkey(certs_endpoint)?;
-        Ok(vec![key])
+        let maybe_key = self.get_ssh_pubkey(certs_endpoint)?;
+        let key: Vec<PublicKey> = maybe_key.into_iter().collect();
+
+        Ok(key)
     }
 
     fn boot_checkin(&self) -> Result<()> {
--- a/src/providers/microsoft/crypto/mod.rs
+++ b/src/providers/microsoft/crypto/mod.rs
@@ -54,7 +54,7 @@ pub fn decrypt_cms(smime: &[u8], pkey: &PKey<Private>, x509: &X509) -> Result<Ve
     Ok(p12_der)
 }
 
-pub fn p12_to_ssh_pubkey(p12_der: &[u8]) -> Result<PublicKey> {
+pub fn p12_to_ssh_pubkey(p12_der: &[u8]) -> Result<Option<PublicKey>> {
     // the contents of that encrypted cms blob we got are actually a different
     // cryptographic structure. we read that in from the contents and parse it.
     // PKCS12 has the ability to have a password, but we don't have one, hence
@@ -65,7 +65,14 @@ pub fn p12_to_ssh_pubkey(p12_der: &[u8]) -> Result<PublicKey> {
     // ParsedPKCS12_2 has three parts: a pkey, a main x509 cert, and a list of other
     // x509 certs. The list of other x509 certs are called the `certificate chain`
     // currently denoted as `ca`; there is only one cert in this `certificate chain`,
-    // which is the ssh public key.
+    // which is the ssh public key. The certs endpoint may still be populated even if
+    // an SSH public key hasn't been provided, which would lead to this code being
+    // executed. The certificate chain will be empty in this case, so just return
+    // None and handle it further up the stack.
+    if p12.ca.is_none() {
+        return Ok(None);
+    }
+
     let ca = p12
         .ca
         .ok_or_else(|| anyhow!("failed to get chain from pkcs12"))?;
@@ -86,5 +93,5 @@ pub fn p12_to_ssh_pubkey(p12_der: &[u8]) -> Result<PublicKey> {
     let n = ssh_pubkey_rsa.n().to_vec();
     let ssh_pubkey = PublicKey::from_rsa(e, n);
 
-    Ok(ssh_pubkey)
+    Ok(Some(ssh_pubkey))
 }
--- a/src/providers/microsoft/azurestack/mod.rs
+++ b/src/providers/microsoft/azurestack/mod.rs
@@ -267,7 +267,7 @@ impl AzureStack {
     }
 
     // put it all together
-    fn get_ssh_pubkey(&self, certs_endpoint: String) -> Result<PublicKey> {
+    fn get_ssh_pubkey(&self, certs_endpoint: String) -> Result<Option<PublicKey>> {
         // we have to generate the rsa public/private keypair and the x509 cert
         // that we use to make the request. this is equivalent to
         // `openssl req -x509 -nodes -subj /CN=LinuxTransport -days 365 -newkey rsa:2048 -keyout private.pem -out cert.pem`
@@ -287,10 +287,7 @@ impl AzureStack {
             .context("failed to decrypt cms blob")?;
 
         // convert that to the OpenSSH public key format
-        let ssh_pubkey = crypto::p12_to_ssh_pubkey(&p12)
-            .context("failed to convert pkcs12 blob to ssh pubkey")?;
-
-        Ok(ssh_pubkey)
+        crypto::p12_to_ssh_pubkey(&p12).context("failed to convert pkcs12 blob to ssh pubkey")
     }
 
     fn fetch_hostname(&self) -> Result<Option<String>> {
@@ -326,18 +323,17 @@ impl MetadataProvider for AzureStack {
         let goalstate = self.fetch_goalstate()?;
         let certs_endpoint = match goalstate.certs_endpoint() {
             Some(ep) => ep,
-            None => {
-                warn!("SSH pubkeys requested, but not provisioned for this instance");
-                return Ok(vec![]);
-            }
+            None => return Ok(vec![]),
         };
 
         if certs_endpoint.is_empty() {
             bail!("unexpected empty certificates endpoint");
         }
 
-        let key = self.get_ssh_pubkey(certs_endpoint)?;
-        Ok(vec![key])
+        let maybe_key = self.get_ssh_pubkey(certs_endpoint)?;
+        let key: Vec<PublicKey> = maybe_key.into_iter().collect();
+
+        Ok(key)
     }
 
     fn boot_checkin(&self) -> Result<()> {


src/providers/microsoft/azure/mod.rs
```diff
@@ -412,7 +412,22 @@ impl MetadataProvider for Azure {
             bail!("unexpected empty certificates endpoint");
         }
 
-        let key = self.get_ssh_pubkey(certs_endpoint)?;
+        // The certs endpoint is being populated even when the user does not supply
+        // an SSH key when created a VM. In this case the certificate chain is empty
```
dustymabe:
```suggestion
        // an SSH key when creating a VM. In this case the certificate chain is empty
```

src/providers/microsoft/azure/mod.rs
```diff
@@ -412,7 +412,22 @@ impl MetadataProvider for Azure {
...
+        // causing a failure when trying to access it. Check if the certificate chain
+        // is empty, meaning that an SSH key hasn't been provided, and only warn
+        // instead of fail.
+        let key = match self.get_ssh_pubkey(certs_endpoint) {
+            Ok(k) => k,
+            Err(e) => match e.root_cause().to_string().as_ref() {
+                "SSH public key not found in chain" => {
+                    warn!("SSH pubkeys requested, but not provisioned for this instance");
```
dustymabe: I'm not sure this is the right warning to give to users, this makes it seem like we desired something (`pubkeys requested`), but couldn't find them.

src/providers/microsoft/crypto/mod.rs
```diff
@@ -65,7 +65,14 @@ pub fn p12_to_ssh_pubkey(p12_der: &[u8]) -> Result<PublicKey> {
     // ParsedPKCS12_2 has three parts: a pkey, a main x509 cert, and a list of other
     // x509 certs. The list of other x509 certs are called the `certificate chain`
     // currently denoted as `ca`; there is only one cert in this `certificate chain`,
-    // which is the ssh public key.
+    // which is the ssh public key. The certs endpoint may still be populated even if
+    // an SSH public key hasn't been provided, which would lead to this code being
+    // executed. The certificate chain will be empty in this case, so error out and
+    // handle it further up the stack.
+    if p12.ca.is_none() {
+        return Err(anyhow!("SSH public key not found in chain"));
```
dustymabe: maybe something like:

```suggestion
        return Err(anyhow!("detected empty certificate chain"));
```

and then up the stack we give a better warning message?

src/providers/microsoft/azure/mod.rs
dustymabe: maybe something like:

```
No SSH keys found in the certificate chain for this instance. 
```

dustymabe (commented):
mostly LGTM

Thanks for working on this!

src/providers/microsoft/azure/mod.rs
marmijo: Thanks for catching this one. I rewrote this comment several times.

src/providers/microsoft/azure/mod.rs
marmijo: That sounds like a much better warning. 
While we have this open, should I change the message a few lines up as well? I was attempting to show the warning we would have expected prior to https://github.com/coreos/fedora-coreos-tracker/issues/1553

marmijo: https://github.com/coreos/afterburn/blob/f1c93ed5a6acde61f11942731766fd385f44a2e4/src/providers/microsoft/azure/mod.rs#L406

marmijo: Since that warning would display when the certs endpoint doesn't exist, maybe that one could just read: 
```
SSH pubkeys not provisioned for this instance
```

src/providers/microsoft/crypto/mod.rs
marmijo: SGTM

src/providers/microsoft/azure/mod.rs
dustymabe: sounds good to me

dustymabe (approved):
LGTM

docs/release-notes.md
```diff
@@ -12,6 +12,8 @@ Major changes:
 
 Minor changes:
 
+- Fix Azure SSH key fetching when no key provisioned.
```
jlebon: Minor: we usually don't have a trailing period in these notes

src/providers/microsoft/azure/mod.rs
```diff
@@ -412,7 +412,22 @@ impl MetadataProvider for Azure {
...
-        let key = self.get_ssh_pubkey(certs_endpoint)?;
+        // The certs endpoint is being populated even when the user does not supply
+        // an SSH key when creating a VM. In this case the certificate chain is empty
+        // causing a failure when trying to access it. Check if the certificate chain
+        // is empty, meaning that an SSH key hasn't been provided, and only warn
+        // instead of fail.
+        let key = match self.get_ssh_pubkey(certs_endpoint) {
+            Ok(k) => k,
+            Err(e) => match e.root_cause().to_string().as_ref() {
+                "detected empty certificate chain" => {
```
jlebon: So instead of string matching errors, a more rustic way to do this is to have `get_ssh_pubkey()` return a `Result<Option<PublicKey>>`. Then, `get_ssh_pubkey()` would return `Ok(None)` for this scenario. And here, we'd do e.g.:

```rust
let key: Vec<PublicKey> = self.get_ssh_pubkeys(certs_endpoint)?.unwrap_or_default();
```

src/providers/microsoft/azure/mod.rs
```diff
@@ -412,7 +412,22 @@ impl MetadataProvider for Azure {
...
+                    warn!("No SSH keys found in the certificate chain for this instance");
```
jlebon: Is this cause for warning?

src/providers/microsoft/azure/mod.rs
```diff
@@ -403,7 +403,7 @@ impl MetadataProvider for Azure {
         let certs_endpoint = match goalstate.certs_endpoint() {
             Some(ep) => ep,
             None => {
-                warn!("SSH pubkeys requested, but not provisioned for this instance");
+                warn!("SSH pubkeys not provisioned for this instance");
```
jlebon: I think the "requested" wording in this warning refers to the fact that *something* called `afterburn --ssh-keys USER` but we couldn't find anything. And so at this layer of this stack, the warning as written previously was accurate. (Even though we know that at the FCOS layer, it's because we unconditionally run this service for the `core` user.)

src/providers/microsoft/azure/mod.rs
dustymabe: maybe not warn, but at least an info log statement?

src/providers/microsoft/azure/mod.rs
dustymabe: what does `afterburn --ssh-keys USER` imply, though?

It's hard not to look at this from the lens of FCOS where we are saying essentially "if there are SSH keys give them to me".

src/providers/microsoft/azure/mod.rs
marmijo: I'm not sure if it is cause for warning, but I chose to warn here because that's the behavior we expected when an SSH key wasn't supplied before this issue arose.

jlebon: So to make sure I understand this: before, Azure would not offer up a certificate endpoint if no SSH keys were provided, but now it always does and instead signals no SSH keys through the certificate itself?

marmijo: Basically, yes. I don't think Azure is directly signaling that no SSH key is provided through the certificate, but that's what we've found so far: that the certificate chain is empty when no SSH key is provided.

src/providers/microsoft/azure/mod.rs
jlebon: After looking at the code for the other providers and the history of this line, I think we should drop this warning entirely. Other providers don't warn in the case that no SSH keys were found, and AFAICT there is no reason to treat Azure differently.

src/providers/microsoft/azure/mod.rs
jlebon: In that case, if the new behaviour is that the certificate chain is empty when no SSH keys are present, then it should not even be considered a warning since it's completely supported to not pass an SSH key. For the same reason as the other comment, I wouldn't even `info!` it (just like other providers don't `info!` if they didn't get a key), but not strongly against. We already log something when we do write SSH keys.

`debug!()` sounds appropriate as a debugging aid if we want and could live directly in `src/providers/microsoft/crypto/mod.rs` where we `return Ok(None)` so that the calling code here can stay simple with `unwrap_or_default()` (otherwise, to log it here, we'd need to inspect the `Option` first to know if it's `None`).

src/providers/microsoft/azure/mod.rs
marmijo: I modified the line based on your suggestion. Since `PublicKey` doesn't implement Default, I used `unwrap_or(vec![])` instead to assign an empty vector if a None value is returned.

marmijo:
I reworked this PR and updated the description based on the review comments.

marmijo:
The "**Rust / Lints, pinned toolchain (pull_request)**" check is failing on `cargo fmt -- --check -l`, but I get no errors when I run this command locally on this branch. What am I missing here?

src/providers/microsoft/azure/mod.rs
```diff
@@ -271,10 +271,13 @@ impl Azure {
...
-        let ssh_pubkey = crypto::p12_to_ssh_pubkey(&p12)
-            .context("failed to convert pkcs12 blob to ssh pubkey")?;
+        let ssh_pubkey = match crypto::p12_to_ssh_pubkey(&p12)
+            .context("failed to convert pkcs12 blob to ssh pubkey")? {
+                Some(key) => key,
+                None => return Ok(None)
+            };
 
-        Ok(ssh_pubkey)
+        Ok(Some(vec![ssh_pubkey]))
```
jlebon: How about making the signature of this function as `Result<Option<PublicKey>>` and then here it'd just be

```suggestion
        Ok(Some(ssh_pubkey))
```

src/providers/microsoft/azure/mod.rs
```diff
@@ -402,18 +405,16 @@ impl MetadataProvider for Azure {
...
 
         if certs_endpoint.is_empty() {
             bail!("unexpected empty certificates endpoint");
         }
 
-        let key = self.get_ssh_pubkey(certs_endpoint)?;
-        Ok(vec![key])
+        let key: Vec<PublicKey> = self.get_ssh_pubkey(certs_endpoint)?.unwrap_or(vec![]);
+
+        Ok(key)
```
jlebon: Then this code here would remain untouched.

IOW, the lower-level `get_ssh_pubkey()` function can just focus on doing the obvious thing and this function here is where we have to deal with what our callers expect from us.

jlebon (commented):
> The "Rust / Lints, pinned toolchain (pull_request)" check is failing on cargo fmt -- --check -l, but I get no errors when I run this command locally on this branch. What am I missing here?

This likely means that you're using a different version than what CI is using. You can see the version it's using in https://github.com/coreos/afterburn/blob/547c950a97b8db67d27f36e1979052f434c57f5f/.github/workflows/rust.yml#L21 and compare to yours. You don't have to match it; it should be enough to make sure you have the latest toolchain from Fedora (e.g. `dnf update rust cargo clippy rustfmt`) or from upstream if you used rustup (e.g. `rustup update`).

src/providers/microsoft/azure/mod.rs
marmijo: I was having a difficult time figuring out how to return a vector on `None` without inspecting the Option like 
https://github.com/coreos/afterburn/blob/c3aa867278ab4c89a6a45a3a5ee239d31ee4509e/src/providers/microsoft/azure/mod.rs#L406-L408

src/providers/microsoft/azure/mod.rs
jlebon: Actually, I think once we do the above, we can shrink this whole thing to just call `crypto::p12_to_ssh_pubkey(&p12).context("failed to convert pkcs12 blob to ssh pubkey")` as the final line of the function since it already matches the signature we need.

src/providers/microsoft/azure/mod.rs
jlebon: Oh right sorry, I missed that detail here. There's a neat Rust trick which is that `Option` actually supports iteration, in which case it'll iterate zero times or once (if the item exists). So something like `Ok(maybe_key.into_iter().collect())` should work.

marmijo: Perfect, thanks! I made that change.

src/providers/microsoft/azure/mod.rs
```diff
@@ -271,10 +271,14 @@ impl Azure {
...
 
         // convert that to the OpenSSH public key format
-        let ssh_pubkey = crypto::p12_to_ssh_pubkey(&p12)
-            .context("failed to convert pkcs12 blob to ssh pubkey")?;
+        let ssh_pubkey = match crypto::p12_to_ssh_pubkey(&p12)
+            .context("failed to convert pkcs12 blob to ssh pubkey")?
+        {
+            Some(key) => key,
+            None => return Ok(None),
+        };
```
jlebon: What I mean is that all this can now be

```suggestion
        crypto::p12_to_ssh_pubkey(&p12)
            .context("failed to convert pkcs12 blob to ssh pubkey")
```

And we can remove the `Ok()` line below.

jlebon (approved):
One nit, LGTM otherwise!

src/providers/microsoft/azure/mod.rs
marmijo: Updated! Thank you.

marmijo: I didnt update it properly. I'll work on it.

jlebon (approved):
Nice work!

PR #1131, opened 2024-11-14
openstack provider: ignore ec2 metadata if not present

rccrdpccl:
On baremetal platform, when reading metadata from a configdrive with openstack layout  we get the following error:
```
# afterburn --provider openstack --attributes /etc/afterburn_metadata
Error: failed to run

Caused by:
    0: writing metadata attributes
    1: failed to open file '"/tmp/afterburn-7Uu4g0/ec2/latest/meta-data.json"'
    2: No such file or directory (os error 2)
```

This PR ignores ec2 metadata if not present:

```
# afterburn --provider openstack --attributes /etc/afterburn_metadata
# cat /etc/afterburn_metadata
AFTERBURN_OPENSTACK_INSTANCE_UUID=30aed4c7-05a4-4d15-be9b-9194dc1b5ad1
```


src/providers/openstack/configdrive.rs
```diff
@@ -144,17 +144,21 @@ impl OpenstackConfigDrive {
...
-        if let Some(instance_id) = metadata_ec2.instance_id {
-            out.insert("OPENSTACK_INSTANCE_ID".to_string(), instance_id);
-        }
         if let Some(uuid) = metadata_openstack.uuid {
             out.insert("OPENSTACK_INSTANCE_UUID".to_string(), uuid);
         }
+
+        let metadata_ec2 = match self.read_metadata_ec2() {
+            Ok(metadata) => metadata,
+            Err(error) => return Ok(out),
```
jlebon: First, let's instead change `read_metadata_ec2()` to return a `Result<Option<MetadataEc2JSON>>`. So then it would return `Ok(None)` in this case. It should only return that if the error was `ENOENT`. Grep the codebase for `NotFound` for examples of this.

Second, this is a lot of metadata we're not injecting now. Are you sure that information is not available in the `openstack/meta_data.json`? I was under the impression when we talked about this that it was.

rccrdpccl: All the metadata we are not injecting seems to be either ec2 related (INSTANCE_ID, INSTANCE_TYPE), or somewhat irrelevant, at least in my use case: IPV4_LOCAL and IPV4_PUBLIC .

The information provided by metal3 is the following:
```
{
  "local-hostname": "bmh-vm-02",
  "local_hostname": "bmh-vm-02",
  "metal3-name": "bmh-vm-02",
  "metal3-namespace": "test-capi",
  "name": "bmh-vm-02",
  "uuid": "30aed4c7-05a4-4d15-be9b-9194dc1b5ad1"
}
```

We can see uuid, hostname - although none of its label it's matching the current provider's key - and other metal3 related keys (useful to build the providerID pattern expected by metal3).

src/providers/openstack/configdrive.rs
```diff
@@ -101,20 +101,27 @@ impl OpenstackConfigDrive {
...
+    fn read_metadata_ec2(&self) -> Result<Option<MetadataEc2JSON>> {
+        use std::io::ErrorKind::NotFound;
+
         let filename = self.metadata_dir("ec2").join("meta-data.json");
-        let file =
-            File::open(&filename).with_context(|| format!("failed to open file '{filename:?}'"))?;
+        let file = match File::open(&filename) {
+            Ok(file) => file,
+            Err(e) if e.kind() == NotFound => return Ok(None),
+            Err(e) => return Err(anyhow::Error::new(e).context(format!("failed to open file '{filename:?}'"))),
```
jlebon: with_context

src/providers/openstack/configdrive.rs
```diff
@@ -101,20 +101,27 @@ impl OpenstackConfigDrive {
...
+        };
         let bufrd = BufReader::new(file);
-        Self::parse_metadata_ec2(bufrd)
-            .with_context(|| format!("failed to parse file '{filename:?}'"))
+        match Self::parse_metadata_ec2(bufrd)
+            .with_context(|| format!("failed to parse file '{filename:?}'")) {
+                Ok(metadata) => Ok(Some(metadata)),
+                Err(e) => Err(e),
+            }
```
jlebon: Instead of the `match`, you can `.map(Some)` to wrap it in an Option.

src/providers/openstack/configdrive.rs
```diff
@@ -144,17 +151,22 @@ impl OpenstackConfigDrive {
...
-        }
         if let Some(uuid) = metadata_openstack.uuid {
             out.insert("OPENSTACK_INSTANCE_UUID".to_string(), uuid);
         }
+
+        let metadata_ec2 = match self.read_metadata_ec2() {
+            Ok(Some(metadata)) => metadata,
+            Ok(None) => return Ok(out),
+            Err(e) => return Err(e),
+        };
```
jlebon: You can do something like:

```rust
let Some(...) = self...()? else {
    return Ok(out);
};
```

docs/release-notes.md
```diff
@@ -10,6 +10,8 @@ Major changes:
 
 Minor changes:
 
+- Openstack: do not fail if ec2 metadata is not found
```
jlebon:
```suggestion
- OpenStack: do not fail if ec2 metadata is not found
```

rccrdpccl:
@jlebon could you PTAL? Not urgent but I'd like to merge it eventually

rccrdpccl:
@jlebon :pray:

jlebon:
Sorry for the delay on this. As it happens, @olljanat contributed a semantically identical patch in https://github.com/coreos/afterburn/pull/1155 for a different purpose. That patch also added a unit test.

I've taken the liberty of combining the best of both patches into one and gave authorship to you both. Let me know if you both agree with the final commit, and we can get this in.

olljanat (approved):
LGTM

rccrdpccl:
lgtm! thanks @jlebon for handling it

jlebon (approved):

PR #1165, opened 2025-02-26
Proxmoxve: Support static network configuration from cloud-init

jdoss:
This PR adds in support for static network configuration from cloud-init data along with tests. It follows the same methods of configuring the network in the [initramfs](https://coreos.github.io/afterburn/usage/initrd-network-cmdline/#initrd-first-boot-network-arguments) via `afterburn-network-kargs.service`. Instead of reading from injected kargs for the network config like the VMware provider, we read mount the cidata disk and read the values out of the network-config file.

### DIFF
--- a/src/providers/proxmoxve/cloudconfig.rs
+++ b/src/providers/proxmoxve/cloudconfig.rs
@@ -34,18 +34,8 @@ pub struct ProxmoxVECloudMetaData {
 #[derive(Debug, Deserialize)]
 pub struct ProxmoxVECloudUserData {
     pub hostname: String,
-    pub manage_etc_hosts: bool,
-    pub fqdn: String,
-    pub chpasswd: ProxmoxVECloudChpasswdConfig,
-    pub users: Vec<String>,
-    pub package_upgrade: bool,
     #[serde(default)]
     pub ssh_authorized_keys: Vec<String>,
-}
-
-#[derive(Debug, Deserialize)]
-pub struct ProxmoxVECloudChpasswdConfig {
-    pub expire: bool,
 }
 
 #[derive(Debug, Deserialize)]
@@ -53,7 +43,6 @@ pub struct ProxmoxVECloudVendorData {}
 
 #[derive(Debug, Deserialize)]
 pub struct ProxmoxVECloudNetworkConfig {
-    pub version: u32,
     pub config: Vec<ProxmoxVECloudNetworkConfigEntry>,
 }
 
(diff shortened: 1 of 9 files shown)


travier (commented):
From a quick look, this seems reasonable, but I haven't followed the discussion and the context closely so I'll let other review and test this :)

src/providers/proxmoxve/configdrive.rs
```diff
@@ -16,23 +16,34 @@ pub struct ProxmoxVEConfigDrive {
...
             .context("failed to create temporary directory")?;
 
-        crate::util::mount_ro(
-            &Path::new("/dev/disk/by-label/").join(CONFIG_DRIVE_LABEL),
-            mount_dir.path(),
-            TARGET_FS,
-            3,
-        )?;
+        let device_path = Self::find_cidata_device()
+            .ok_or_else(|| anyhow::anyhow!("could not find cidata device"))?;
```
dustymabe, line 39: The `/dev/disk/by-label` that was used before wasn't sufficient?" Might be worth putting that in a separate commit with reasoning here.

src/providers/proxmoxve/cloudconfig.rs
```diff
@@ -33,26 +34,15 @@ pub struct ProxmoxVECloudMetaData {
 #[derive(Debug, Deserialize)]
 pub struct ProxmoxVECloudUserData {
     pub hostname: String,
-    pub manage_etc_hosts: bool,
-    pub fqdn: String,
-    pub chpasswd: ProxmoxVECloudChpasswdConfig,
-    pub users: Vec<String>,
-    pub package_upgrade: bool,
```
dustymabe, line 40: What is it OK to delete this? Was nothing using it? 

Maybe in a separate first commit we should delete the things that are unused first?

dustymabe (commented):
Thanks for working on this @jdoss. Sorry it's taken so long to get around to a review. I'm no rust programmer (only done a little learning here and there) so can't comment much on syntactical errors, but I can comment on logic and structure.

Can we run cargo fmt in a first commit and then only add properly formatted code in new commits?

Also it seems like we do a lot in the first commit. Can we delete unneeded code in a separate initial commit and also add extra context/logging in a separate initial commit?

src/providers/proxmoxve/configdrive.rs
jdoss: There is a condition where it won't have a label because labeling doesn't happen until after udev runs. So we need to use blkid to find the drive based on label instead.

src/providers/proxmoxve/cloudconfig.rs
jdoss: Yes they are not used and it can cause serde errors if they are not present. Specifically package_upgrade if missing from the YAMLs it would cause afterburn to fail.

src/providers/proxmoxve/configdrive.rs
dustymabe: Nice.. That's a great explanation of "why". Can you put that in a separate commit with a nice commit message?

jdoss:
@dustymabe I have updated my PR based on your feedback. PTAL.

dustymabe (approved):

PR #1215, opened 2025-06-05
oraclecloud: implement oraclecloud provider

alaviss:
Implement new provider for Oracle Cloud Infrastructure. The metadata is fetched via IMDSv2 as documented at
https://docs.oracle.com/en-us/iaas/Content/Compute/Tasks/gettingmetadata.htm

This provider supports:
- Attributes
- Hostname
- SSH Keys


src/providers/oraclecloud/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    fn virtual_network_devices(&self) -> Result<Vec<crate::network::VirtualNetDev>> {
+        Ok(std::vec![])
+    }
+}
+
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct Instance {
```
jlebon, line 130: Hmm, any reason this is at the end instead of upfront?

src/providers/oraclecloud/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+        Self::try_new_with_client(&client)
+    }
+
+    pub(crate) fn try_new_with_client(client: &retry::Client) -> Result<OracleCloudProvider> {
+        let instance = OracleCloudProvider::fetch_instance_metadata(client)?;
+        Ok(OracleCloudProvider { instance })
+    }
+
+    fn endpoint_for(name: &str) -> String {
+        format!("http://169.254.169.254/opc/v2/{name}")
```
jlebon: Let's make a constant for the base URL.

jlebon (commented):
Thanks! Were you able to test this?

jlebon:
Release notes here as well.

src/providers/oraclecloud/mod.rs
alaviss: I just copied how the Akamai provider does it.

alaviss:
> Thanks! Were you able to test this?

Yep, I verified the output files and they looked correct. Can't show that here since it'd include details about my instance.

jlebon (approved):
Thank you!

PR #1216, opened 2025-06-06
dracut: Ignore module when building kdump initrd

Johan-Liebert1:
Related to: coreos/fedora-coreos-tracker#1832

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -9,6 +9,7 @@ nav_order: 8
 Major changes:
 
 - Add support for Oracle Cloud Infrastructure
+- dracut: don't include module by default
 
 Minor changes:
 
--- a/dracut/30afterburn/module-setup.sh
+++ b/dracut/30afterburn/module-setup.sh
@@ -3,7 +3,11 @@
 # ex: ts=8 sw=4 sts=4 et filetype=sh
 
 check() {
-    return 0
+    # Return 255 so this module is only included if some other module depends on it
+    # See: https://github.com/coreos/fedora-coreos-tracker/issues/1832
+    #
+    # This module requires integration with the rest of the initramfs, so don't include it by default.
+    return 255
 }
 
 depends() {


jlebon:
I think this applies here too: https://github.com/coreos/ignition/pull/2090#issuecomment-2949343711

Though looks like we would need to change the return code here to be 255. Which... is technically a breaking change, but I think does makes sense. The initrd units here require integration on the OS side.

Johan-Liebert1:
Thanks for the review. I've updated to unconditionally return 255 from the check function. Should we keep this behind an if guard?

jlebon:
I think first we need to explicitly pull in the `afterburn` dracut module in f-c-c. I think https://github.com/coreos/fedora-coreos-config/blob/1af57afab31a884c7b173d72e6877b02758d71f6/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-network/module-setup.sh is probably the most appropriate spot.

dracut/30afterburn/module-setup.sh
```diff
@@ -3,7 +3,7 @@
 # ex: ts=8 sw=4 sts=4 et filetype=sh
 
 check() {
-    return 0
+    return 255
```
dustymabe, line 10: Feels like this could definitely use a comment for casual readers of the code.

IIUC we are now not being included by default but only if something else pulls our module in?

dustymabe:
```
module-setup.sh: check()
       check() is called by dracut to evaluate the inclusion of a dracut
       module in the initramfs.

       $hostonly
           If the $hostonly variable is set, then the module check()
           function should be in "hostonly" mode, which means, that the
           check() should only return 0, if the module is really needed
           to boot this specific host.

       check() should return with:

       0
           Include the dracut module in the initramfs.

       1
           Do not include the dracut module. The requirements are not
           fulfilled (missing tools, etc.)

       255
           Only include the dracut module, if another module requires it
           or if explicitly specified in the config file or on the
           argument list.
```

Johan-Liebert1: > Feels like this could definitely use a comment for casual readers of the code.

Definitely. I'll add comments explaining the reasoning behind this change.

dustymabe:
The `Release notes / Require release note` check is failing so let's add a release note.

dracut/30afterburn/module-setup.sh
```diff
@@ -3,7 +3,11 @@
 # ex: ts=8 sw=4 sts=4 et filetype=sh
 
 check() {
-    return 0
+    # Return 255 so this module is only included if some other module depends on it
+    # See: https://github.com/coreos/fedora-coreos-tracker/issues/1832
+    #
+    # TLDR: We don't need afterburn in kdump initrd, so this won't be included in it
```
jlebon: FWIW, I personally wouldn't mention kdump here. Even without the kdump impetus, I think we should still do this. I would just say e.g.

```bash
# This module requires integration with the rest of the initramfs, so don't include it by default.
```

Johan-Liebert1: I see. This makes sense. I'll update accordingly

docs/release-notes.md
```diff
@@ -8,6 +8,8 @@ nav_order: 8
 
 Major changes:
 
+- Don't include afterburn module in initramfs by default
```
jlebon: How about:

```suggestion
- dracut: don't include module by default
```

?

jlebon (approved):
Minor comment on the release notes, but otherwise LGTM

docs/release-notes.md
Johan-Liebert1: Sure, I'll update it. Wasn't sure how we write the release notes

Johan-Liebert1: Updated

PR #1230, opened 2025-08-21
providers/azure: switch SSH key retrieval from certs endpoint to IMDS

### LINKED ISSUE #1217: Afterburn can no longer fetch ssh public keys from Azure API Endpoint
# Bug #

Afterburn can no longer fetch ssh public keys from Azure API Endpoint

Fedora CoreOS 42.20250512.3.0
## Afterburn Version ##
afterburn-5.7.0-4.fc42.x86_64 
## Environment ##
Azure

## Expected Behavior ##
ssh keys are deployed by afterburn

## Actual Behavior ##
no ssh keys are deployed

## Reproduction Steps ##

  1. run afterburn --cmdline --ssh-keys=core

## Other Information ##
it seems that the ssh public key is no longer included in the comp=certificates response.
I did not find a chain at all in the encrypted pkcs12.
This worked one or two weeks ago with the same VM image.

marmijo:
Azure no longer reliably provides SSH public keys in the goal-state certificates endpoints (`comp=certificates`) but Azure VM provisioning populates keys in the Instance Metadata Service (IMDS) under `compute/publicKeys`.

Update the Azure provider ssh key retrieval to remove the deprecated certificates flow. Instead, query the IMDS and parse the returned keyData entries into `PublicKey`s.

Fixes: https://github.com/coreos/afterburn/issues/1217

### DIFF
--- a/src/providers/microsoft/azure/mod.rs
+++ b/src/providers/microsoft/azure/mod.rs
@@ -14,19 +14,17 @@
 
 //! Azure provider, metadata and wireserver fetcher.
 
-use super::crypto;
 use super::goalstate;
 
 use std::collections::HashMap;
 use std::net::IpAddr;
 
-use anyhow::{anyhow, bail, Context, Result};
+use anyhow::{anyhow, Context, Result};
 use openssh_keys::PublicKey;
 use reqwest::header::{HeaderName, HeaderValue};
 use serde::Deserialize;
 use slog_scope::warn;
 
-use self::crypto::x509;
 use crate::providers::MetadataProvider;
 use crate::retry;
 use nix::unistd::Uid;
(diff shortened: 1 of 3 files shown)


travier:
Thanks! Do you have a docs/blog post reference for that change that we could link here?

travier (approved):
Completely untested but looks sane

travier:
Should we add a `Fixes: https://github.com/coreos/afterburn/issues/1217` in this PR?

src/providers/microsoft/azure/mod.rs
```diff
@@ -351,6 +291,50 @@ impl Azure {
         Ok(vmsize)
     }
 
+    /// Fetch SSH public keys from Azure Instance Metadata Service (IMDS)
+    fn fetch_ssh_keys(&self) -> Result<Vec<PublicKey>> {
+        const URL: &str = "metadata/instance/compute/publicKeys?api-version=2021-02-01";
```
prestist, line 297: I think it might be good to add to the above comment linking to the doc's for the url referenced here. Just in case it changes or something.

src/providers/microsoft/azure/mod.rs
```diff
@@ -351,6 +291,50 @@ impl Azure {
...
+            .client
+            .clone()
+            .header(
+                HeaderName::from_static("metadata"),
+                HeaderValue::from_static("true"),
+            )
+            .get(retry::Raw, url)
+            .send::<String>()
+            .context("failed to query IMDS for publicKeys")?
+            .ok_or_else(|| anyhow::anyhow!("IMDS did not return a publicKeys payload"))?;
```
prestist, line 310: slick

prestist (commented):
Changes lgtm, thank you for working on this, looks like you just need a comment in release notes as well.

marmijo:
> Thanks! Do you have a docs/blog post reference for that change that we could link here?

No, I wasn't able to find much documentation related to the VM goalstate, or changes in the certificates endpoint.

> Should we add a Fixes: https://github.com/coreos/afterburn/issues/1217 in this PR?

Done!

> Completely untested but looks sane

I tested this in an Azure VM and Afterburn populated my SSH key from the IMDS as expected.

src/providers/microsoft/azure/mod.rs
marmijo: Done!

prestist (approved):
Thank you for the changes and +1000 for adding the tests. lgtm

PR #1243, opened 2025-10-09
kubevirt: modprobe for virtio_blk; remove dracut preload

prestist:
As a followup to https://github.com/coreos/afterburn/pull/1238 Lets make sure we dont affect other platforms.


qinqon:
I will test it with rhcos on kubevirt since it was the problematic one

qinqon (approved):
Tested it, and works for RHCOS both with config drive and no cloud

But looks like I broke ignition for kubevirt for configdrive (meaning all hypershift kubevirt provider nodes will not work) , this is the fix:
-  https://github.com/coreos/ignition/pull/2139

tlbueno (approved):
lgtm

PR #1251, opened 2025-11-26
kubevirt: Support static gateway and DNS with DHCP

qinqon:
Support configuring static gateway routes and DNS servers on DHCP-enabled
interfaces in KubeVirt. This allows mixed configurations where IP addresses
are obtained via DHCP but routing and DNS are statically configured.

Changes:
- **ConfigDrive**: Always process routes on DHCP network types (`ipv4_dhcp`/`ipv6_dhcp`)
  per the OpenStack schema — no custom fields needed
- **NoCloud v1**: Support both `destination` (CIDR) and `network`/`netmask` route formats;
  put routes/DNS directly on dhcp subnets; skip legacy `gateway` when explicit routes exist
- **NoCloud v2**: Parse `dhcp-overrides` for schema compatibility; routes and nameservers
  always emitted when defined
- **Schema validation**: Add test-time validation of all fixtures against upstream JSON
  schemas from [OpenStack Ironic](https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/network-data-schema.json)
  and [cloud-init](https://github.com/canonical/cloud-init/tree/main/cloudinit/config/schemas)
- **Fixture fixes**: Add required fields (`ethernet_mac_address`, `network_id`, `name`),
  move v2 nameservers to per-interface, remove non-standard fields


src/providers/kubevirt/configdrive.rs
```diff
@@ -256,16 +264,40 @@ impl NetworkData {
...
+                        {
+                            should_configure_static_dns = true;
+                        }
+                    } else {
+                        // If accept_dhcp_option is not specified, use legacy behavior:
+                        // include global DNS servers for backwards compatibility
+                        should_configure_static_dns = true;
+                    }
                 }
                 "ipv6_dhcp" => {
```
prestist, line 260: Hmm so really accept_dhcp_option is IPv4 limited right? Is this a technical difference or intentional absence of functionality for ipv6

prestist (commented):
One small question, otherwise it LGTM from my perspective. 

@travier would you mind also taking a look as well :)

src/providers/kubevirt/configdrive.rs
qinqon: Good catch — `accept_dhcp_option` was actually an invented field that doesn't exist in the [OpenStack network_data.json schema](https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/network-data-schema.json). I've removed it in the latest force-push.

The OpenStack schema already allows `routes` on `ipv4_dhcp`/`ipv6_dhcp` network types, so the presence of routes on a DHCP network entry is sufficient signal to configure them statically — no custom field needed.

I also added test-time schema validation (using the upstream JSON schemas from Ironic and cloud-init) to catch this kind of deviation going forward. The fixtures have been fixed to comply with the upstream schemas as well.

travier:
Can you split the IBM parts into another PR or at least another commit? Thanks

Cargo.lock
```diff
@@ -37,7 +37,7 @@ dependencies = [
  "mailparse",
  "maplit",
  "mockito",
- "nix 0.29.0",
+ "nix 0.30.1",
```
travier: Can you split this into another commit? Is this needed for this PR?

qinqon:
> Can you split the IBM parts into another PR or at least another commit? Thanks

@travier I will put the IBM part at different commit, the important part though if we don't change production code related to that is just the test itself.

Cargo.lock
qinqon: Not needed, this is from and old version where I used the rust crate to validate fixtures, but that got simplified by just using a python strict that validate fixtures before using them on tests at the Makefile.

qinqon: Not needed, this is from and old version where I used the rust crate to validate fixtures, but that got simplified by just using a python strict that validate fixtures before using them on tests at the Makefile.

src/providers/kubevirt/nocloud.rs
```diff
@@ -258,32 +308,83 @@ impl NetworkConfigV1Entry {
...
+
+                iface.routes.push(NetworkRoute {
+                    destination,
+                    gateway,
+                });
+            }
 
+            // Handle legacy gateway field only if no explicit routes were defined,
+            // to avoid duplicate default routes
+            if subnet.routes.is_empty() {
```
prestist, line 387: The v1 parser guarded against duplicates, when both gateway and explicit routs are set. But in the v2 I don't think we are? I don't think that this could cause issues in runtime but might make sense to fix?

src/providers/kubevirt/cloudconfig.rs
```diff
@@ -227,6 +227,15 @@ impl MetadataProvider for KubeVirtCloudConfig {
                 }
             }
 
+            // Add static routes for the interface (including DHCP interfaces)
+            // This allows DHCP interfaces to have static gateway configuration
+            for route in &iface.routes {
```
prestist, line 220: Hmmm, the rd.route loop runs for all interfaces, but Is that needed for static IP? I would think the gateway is already set.

Makefile
```diff
@@ -44,4 +44,5 @@ lint:
 
 .PHONY: test
 test:
+	python3 tests/fixtures/validate.py
```
prestist: hmmm, maybe we can make this a separate target, it would be good to not require local dev to need python3 if they want to "make test"

prestist: Maybe "make validate-fixtures"?

qinqon: Make sense, I will add the make target and convert this into a github CI gating thing, at the end if github CI check this is enough, we know the merged fixture will be good.

src/providers/kubevirt/nocloud.rs
qinqon: > The v1 parser guarded against duplicates, when both gateway and explicit routs are set. But in the v2 I don't think we are? I don't think that this could cause issues in runtime but might make sense to fix?

Right, in fact we are missing a pair of fixtures to test that

src/providers/kubevirt/cloudconfig.rs
qinqon: Looks like we can configure static IP without gateway and pass it as route 
```
ip=192.168.1.10:::255.255.255.0::eth0:none rd.route=0.0.0.0/0:192.168.1.1:eth0
```
Let's do this for both static and dynamic IP so we have the same logic.

qinqon:
I have also found that for ipv6 dracut mandates brackets around IPs/routes since ":" is already the dracut separator.

qinqon:
[(pushed)](https://github.com/coreos/afterburn/compare/783c67a78e2fff77f7393482968bcecbab5d3668..795b327babc755e4ef1fba48599d023384dd8bae):
- prevent duplicate default geteway for v2 when both route and gateway is configured
- fix ipv6, at dracut it need to use brackets, since brakcet is the dracut's separator.

prestist (approved):
From a code perspective this lgtm. Thank you for working on this, have you had a chance to test the latest changes in kubevirt?

qinqon:
> From a code perspective this lgtm. Thank you for working on this, have you had a chance to test the latest changes in kubevirt?

I have start three VMs with the three formats and I get on the three of them for dynamic address and static gateway and dns 
```
 Using kernel command line parameters:  rd.driver.pre=btrfs ip=eth0:dhcp,dhcp6 rd.route=0.0.0.0/0:192.168.1.1 rd.route=[::/0]:[2001:db8::1] nameserver=8.8.8.8 nameserver=8.8.4.4
```
I have found that this is invalid the dualstack  `ip=eth0:dhcp,dhcp6` should be `ip=eth0:dhcp ip=eth0:dhcp6`

qinqon:
@prestist pushed the fix, I will ping you after retesting with real VMs on kubevirt.

qinqon:
> @prestist pushed the fix, I will ping you after retesting with real VMs on kubevirt.

So looks like there is a bug at nm-initrd-generator and doing two eth0 configs one override the other so ipv4 get disabled
```
# Created by nm-initrd-generator
[connection]
id=enp1s0
uuid=b98261cd-d6d8-45fa-85f1-bd93edc4b682
type=ethernet
autoconnect-priority=-100
autoconnect-retries=1
interface-name=enp1s0
multi-connect=1

[ethernet]

[ipv4]
method=disabled <<< --- this
route1=0.0.0.0/0,192.168.1.1

[ipv6]
dhcp-timeout=90
may-fail=false
method=auto
route1=::/0,2001:db8::1

[proxy]

[user]
org.freedesktop.NetworkManager.origin=nm-initrd-generator
```

for network manager generator we should fallback to ip=eth0:dhcp,dhcp6 puff... clearly unit test is not enoug for all this :-/

qinqon:
@prestist looks like dracut do not support IPv6 DHCPv6 Stateful for NetworkManager so we cannot implement full IPv6 support is that ok ?

This is the wrong thing at NetworkManager

 For the NM patch, you'd need nm-initrd-generator to:
  - auto6 → ipv6.method=auto (SLAAC, current behavior)
  - dhcp6 → ipv6.method=dhcp (stateful DHCPv6 only, new behavior)

 
Maybe we can merge this since ipv4 is working fine and continue with ipv6 support later on, for the use case we need ipv4 is enough.

prestist:
In am going get back to this soon sorry, been jumping around too much.

prestist:
Yes, this should be fine to merge as-is. The IPv6 DHCPv6 stateful limitation is an upstream dracut/NetworkManager issue, not something we can fix in afterburn.
                                                                               
We can address full IPv6 DHCPv6 stateful support in a future PR once the upstream nm-initrd-generator adds support for ipv6.method=dhcp. The current implementation is correct and doesn't block that work.                                                                                                                                
                                                                                                                                                                         
LGTM to merge. Thank you again for the thorough testing and documentation of the limitation!

PR #1264, opened 2026-03-19
feat(azure): Add Azure Ignition fragment generation

peytonr18:
## Summary

This PR adds a `render-ignition` CLI subcommand to Afterburn that generates Ignition config fragment files from Azure IMDS and OVF metadata. These fragments are written to `/etc/ignition/base.platform.d/azure/` during initrd, allowing Ignition to natively merge platform-provided user data (hostname, SSH keys, password) without requiring a custom Ignition config.

### New `render-ignition` subcommand
- Added `src/cli/render_ignition.rs` with clap-derived argument parsing.
- Accepts `--provider`/`--cmdline`, `--render-ignition-dir`, and per-feature flags (`--hostname`, `--platform-user`, `--platform-extensions`).
- Currently gated to Azure only; other providers bail with an informative error.

### Azure fragment generation (`src/providers/microsoft/azure/config.rs`)
- `generate_hostname_fragment()`: fetches hostname via `MetadataProvider::hostname()`, writes `hostname.ign` with a `storage.files` entry for `/etc/hostname` using a `data:,<hostname>` URI.
- `generate_user_fragment()`: fetches admin username and SSH keys from IMDS via `MetadataProvider`, optionally reads `adminPassword` from OVF (`ovf-env.xml` on CD-ROM), hashes it with SHA-512, and writes `user.ign` with a `passwd.users` entry.

### OVF / IMDS source rules
- Username and SSH keys are sourced from IMDS.
- `adminPassword` is the only field read from OVF. If present and non-empty, it is SHA-512 hashed and included as `passwordHash` in the Ignition user fragment. If absent or empty, it is omitted.

### Systemd / Dracut integration
- New `afterburn-ignition-fragment.service` unit (dracut module) runs during initrd.
- Conditioned on `ignition.platform.id=azure`.
- Runs after `network-online.target`, before `ignition-kargs.service`.
- Creates the output directory and invokes `afterburn render-ignition --cmdline --render-ignition-dir /etc/ignition/base.platform.d/azure/ --platform-extensions`.

## CLI usage

```
afterburn render-ignition --cmdline --render-ignition-dir <path> --hostname --platform-user
afterburn render-ignition --cmdline --render-ignition-dir <path> --platform-extensions
```

| Flag | Effect |
|------|--------|
| `--hostname` | Generate `hostname.ign` (`storage.files` entry for `/etc/hostname`) |
| `--platform-user` | Generate `user.ign` (`passwd.users` entry with username + SSH keys + optional password hash) |
| `--platform-extensions` | Implies `--hostname --platform-user` |

Each feature writes its own standalone fragment file. Missing data (e.g., no hostname available) logs a warning and skips, matching existing Afterburn behavior.

### Existing functionality is untouched
The `multi` subcommand's `--hostname=<path>` direct-file-write path is unchanged. The `afterburn-hostname.service` (including its Azure/AzureStack conditions) is not modified.

**The CLI design is up for discussion — would love feedback on the flag names, subcommand structure, and overall approach.**

## Summary by CodeRabbit

* **New Features**
  * Added Azure support for generating Ignition configuration fragments via the new `render-ignition` command (with options to disable hostname and/or user fragments).
  * Added first-boot integration to automatically generate Azure fragments during Ignition startup.
  * Enhanced Azure metadata handling, including admin username and admin password hashing from provisioning configuration.

* **Bug Fixes**
  * Improved Azure SSH key parsing and hostname truncation/omission behavior.
  * Fragment writing is now safer and atomic, with correct file permissions based on whether a password hash is present.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,6 +32,7 @@ dependencies = [
  "cfg-if",
  "clap",
  "ipnetwork",
+ "libc",
  "libflate",
  "libsystemd",
  "mailparse",
@@ -40,6 +41,7 @@ dependencies = [
  "nix 0.30.1",
  "openssh-keys",
  "openssl",
+ "percent-encoding",
  "pnet_base",
  "pnet_datalink",
  "reqwest",
@@ -54,6 +56,7 @@ dependencies = [
  "tempfile",
  "uzers",
  "vmw_backdoor",
+ "xml-rs",
  "zbus",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,6 +17,7 @@ dependencies = [
  "cfg-if",
  "clap",
  "ipnetwork 0.21.1",
+ "libc",
  "libflate",
  "libsystemd",
  "mailparse",
@@ -25,6 +26,7 @@ dependencies = [
  "nix 0.31.3",
  "openssh-keys",
  "openssl",
+ "percent-encoding",
  "pnet_base",
  "pnet_datalink",
  "reqwest",
@@ -39,6 +41,7 @@ dependencies = [
  "tempfile",
  "uzers",
  "vmw_backdoor",
+ "xml-rs",
  "zbus",
 ]
 
@@ -3108,6 +3111,12 @@ name = "xml"
 version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "636f85e5ca6488e96401b61eb7de54f4e44755c988af0f52cf90230c312a1a89"
+
+[[package]]
+name = "xml-rs"
+version = "0.8.28"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3ae8337f8a065cfc972643663ea4279e04e7256de865aa66fe25cec5fb912d3f"
 
 [[package]]
 name = "yoke"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ base64 = "0.22"
 cfg-if = "1.0"
 clap = { version = "4", "default-features" = false, "features" = ["std", "cargo", "derive", "error-context", "help", "suggestions", "usage", "wrap_help"] }
 ipnetwork = ">= 0.17, < 0.22"
+libc = "0.2"
 libflate = "2.1"
 libsystemd = ">= 0.2.1, < 0.8.0"
 mailparse = ">= 0.13, < 0.17"
@@ -46,6 +47,7 @@ maplit = "1.0"
 nix = { version = ">= 0.19, < 0.31", "default-features" = false, "features" = [ "mount", "user"] }
 openssh-keys = ">= 0.5, < 0.7"
 openssl = ">= 0.10.46, < 0.11"
+percent-encoding = "2.3.2"
 pnet_base = ">= 0.26, < 0.36"
 pnet_datalink = ">= 0.26, < 0.36"
 reqwest = { version = ">= 0.10, < 0.13", features = [ "blocking" ] }
@@ -54,6 +56,7 @@ serde-xml-rs = ">= 0.4, < 0.9"
 serde_json = "1.0"
 serde_yaml = ">= 0.8, < 0.10"
 slog = { version = "2.7", features = ["max_level_trace", "release_max_level_info"] }
+xml-rs = "0.8"
 slog-async = ">= 2.5, < 3"
 slog-scope = "4.3"
 slog-term = ">= 2.6, < 3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ base64 = "0.22"
 cfg-if = "1.0"
 clap = { version = "4", "default-features" = false, "features" = ["std", "cargo", "derive", "error-context", "help", "suggestions", "usage", "wrap_help"] }
 ipnetwork = ">= 0.17, < 0.22"
+libc = "0.2"
 libflate = "<2.4"
 libsystemd = ">= 0.2.1, < 0.8.0"
 mailparse = ">= 0.13, < 0.17"
@@ -47,6 +48,7 @@ maplit = "1.0"
 nix = { version = ">= 0.19, < 0.32", "default-features" = false, "features" = [ "mount", "user"] }
 openssh-keys = ">= 0.5, < 0.7"
 openssl = ">= 0.10.46, < 0.11"
+percent-encoding = "2.3.2"
 pnet_base = ">= 0.26, < 0.36"
 pnet_datalink = ">= 0.26, < 0.36"
 reqwest = { version = ">= 0.10, < 0.14", features = [ "blocking" ] }
@@ -61,6 +63,7 @@ slog-term = ">= 2.6, < 3"
 tempfile = ">= 3.2, < 4"
 uzers = "0.12"
 vmw_backdoor = "0.2"
+xml-rs = "0.8"
 zbus = ">= 2.3, < 6"
 
 [dev-dependencies]
--- a/dracut/30afterburn/afterburn-ignition-fragment.service
+++ b/dracut/30afterburn/afterburn-ignition-fragment.service
@@ -1,0 +1,26 @@
+[Unit]
+Description=Afterburn Ignition Config Fragment Generator
+Documentation=https://coreos.github.io/afterburn/
+
+# Only run on platforms that support ignition fragment generation.
+ConditionKernelCommandLine=|ignition.platform.id=azure
+
+# Run after networking is available (needed for the Azure IMDS call). Use
+# network.target (not network-online.target) to match ignition-fetch.service
+# and avoid stalling on multi-NIC DHCP; IMDS is link-local and afterburn
+# has its own retry logic.
+#
+# Run between fetch and disks so writes don't race the fetch stage's parse
+# of base.platform.d (Ignition re-reads it every stage; applied at files).
+After=network.target
+After=ignition-fetch.service
+Before=ignition-disks.service
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+ExecStartPre=/usr/bin/mkdir -p /etc/ignition/base.platform.d/azure
+ExecStart=/usr/bin/afterburn render-ignition --cmdline --render-ignition-dir /etc/ignition/base.platform.d/azure/ --platform-extensions
+Type=oneshot
+RemainAfterExit=yes
--- a/dracut/30afterburn/afterburn-ignition-fragment.service
+++ b/dracut/30afterburn/afterburn-ignition-fragment.service
@@ -1,0 +1,26 @@
+[Unit]
+Description=Afterburn Ignition Config Fragment Generator
+Documentation=https://coreos.github.io/afterburn/
+
+# Only run on platforms that support ignition fragment generation.
+ConditionKernelCommandLine=|ignition.platform.id=azure
+
+# Run after networking is available (needed for the Azure IMDS call). Use
+# network.target (not network-online.target) to match ignition-fetch.service
+# and avoid stalling on multi-NIC DHCP; IMDS is link-local and afterburn
+# has its own retry logic.
+#
+# Run between fetch and disks so writes don't race the fetch stage's parse
+# of base.platform.d (Ignition re-reads it every stage; applied at files).
+After=network.target
+After=ignition-fetch.service
+Before=ignition-disks.service
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+ExecStartPre=/usr/bin/mkdir -p /etc/ignition/base.platform.d/azure
+ExecStart=/usr/bin/afterburn render-ignition --cmdline --render-ignition-dir /etc/ignition/base.platform.d/azure/
+Type=oneshot
+RemainAfterExit=yes
--- a/dracut/30afterburn/afterburn-ignition-fragment.service
+++ b/dracut/30afterburn/afterburn-ignition-fragment.service
@@ -21,6 +21,6 @@ OnFailureJobMode=isolate
 
 [Service]
 ExecStartPre=/usr/bin/mkdir -p /etc/ignition/base.platform.d/azure
-ExecStart=/usr/bin/afterburn render-ignition --cmdline --render-ignition-dir /etc/ignition/base.platform.d/azure/ --platform-extensions
+ExecStart=/usr/bin/afterburn render-ignition --cmdline --render-ignition-dir /etc/ignition/base.platform.d/azure/
 Type=oneshot
 RemainAfterExit=yes
--- a/dracut/30afterburn/module-setup.sh
+++ b/dracut/30afterburn/module-setup.sh
@@ -23,9 +23,13 @@ install() {
     inst_simple "$moddir/afterburn-network-kargs.service" \
         "$systemdutildir/system/afterburn-network-kargs.service"
 
+    inst_simple "$moddir/afterburn-ignition-fragment.service" \
+        "$systemdutildir/system/afterburn-ignition-fragment.service"
+
     # These services are only run once on first-boot, so they piggyback
     # on Ignition completion target.
     mkdir -p "$initdir/$systemdsystemunitdir/ignition-complete.target.requires"
     ln -s "../afterburn-hostname.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-hostname.service"
     ln -s "../afterburn-network-kargs.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-network-kargs.service"
+    ln -s "../afterburn-ignition-fragment.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-ignition-fragment.service"
 }
--- a/src/cli/render_ignition.rs
+++ b/src/cli/render_ignition.rs
@@ -1,0 +1,78 @@
+//! `render-ignition` CLI sub-command.
+//!
+//! Generates per-feature Ignition config fragment files in an output directory.
+//! Each enabled feature writes its own `.ign` file; Ignition merges them
+//! natively from `base.platform.d/<platform>/` under a system config directory
+//! such as `/etc/ignition`.
+
+use anyhow::{bail, Context, Result};
+use clap::{ArgGroup, Parser};
+
+/// Render Ignition config fragments from cloud provider metadata
+#[derive(Debug, Parser)]
+#[command(group(ArgGroup::new("provider-group").args(["cmdline", "provider"]).required(true)))]
+pub struct CliRenderIgnition {
+    /// The name of the cloud provider
+    #[arg(long, value_name = "name")]
+    provider: Option<String>,
+    /// Read the cloud provider from the kernel cmdline
+    #[arg(long)]
+    cmdline: bool,
+    /// Directory to write Ignition config fragment files into
+    #[arg(long = "render-ignition-dir", value_name = "path")]
+    render_ignition_dir: String,
+    /// Include hostname in a fragment file
+    #[arg(long)]
+    hostname: bool,
+    /// Include platform user and SSH keys in a fragment file
+    #[arg(long)]
+    platform_user: bool,
+    /// Enable all platform extensions (implies --hostname --platform-user)
+    #[arg(long)]
+    platform_extensions: bool,
+}
+
+impl CliRenderIgnition {
+    const SUPPORTED_PROVIDERS: &[&str] = &["azure"];
+
+    pub(crate) fn run(self) -> Result<()> {
+        let provider_id = super::get_provider(self.provider.as_deref())?;
+
+        if !Self::SUPPORTED_PROVIDERS.contains(&provider_id.as_str()) {
+            bail!(
+                "render-ignition is only supported for providers {:?}, got '{}'",
+                Self::SUPPORTED_PROVIDERS,
+                provider_id,
+            );
+        }
+
+        let hostname = self.hostname || self.platform_extensions;
+        let platform_user = self.platform_user || self.platform_extensions;
+
+        if !hostname && !platform_user {
+            slog_scope::warn!("render-ignition: no features specified");
+            return Ok(());
+        }
+
+        let metadata = crate::metadata::fetch_metadata(&provider_id)
+            .context("fetching metadata from provider")?;
+
+        if hostname {
+            crate::providers::microsoft::azure::config::generate_hostname_fragment(
+                metadata.as_ref(),
+                &self.render_ignition_dir,
+            )
+            .context("generating hostname ignition fragment")?;
+        }
+
+        if platform_user {
+            crate::providers::microsoft::azure::config::generate_user_fragment(
+                metadata.as_ref(),
+                &self.render_ignition_dir,
+            )
+            .context("generating platform-user ignition fragment")?;
+        }
+
+        Ok(())
+    }
+}
--- a/src/cli/render_ignition.rs
+++ b/src/cli/render_ignition.rs
@@ -1,24 +1,17 @@
 //! `render-ignition` CLI sub-command.
 //!
-//! Generates per-feature Ignition config fragment files in an output directory.
-//! Each enabled feature writes its own `.ign` file; Ignition merges them
-//! natively from `base.platform.d/<platform>/` under a system config directory
-//! such as `/etc/ignition`.
-//!
-//! The Ignition config fragment types and `write_fragment`/`hostname_data_uri`
-//! helpers are provider-agnostic and live here so other platforms can reuse
-//! them; any provider-specific data (e.g. the admin password) is sourced
-//! through the `MetadataProvider` trait rather than referenced directly.
+//! Fetches metadata from a cloud provider and writes Ignition config fragment
+//! files for the enabled features. The fragment types and their serialization
+//! live in [`crate::ignition`]; this module turns provider metadata into those
+//! fragments and handles argument parsing and dispatch.
 
 use anyhow::{bail, Context, Result};
 use clap::{ArgGroup, Parser};
-use serde::Serialize;
 use slog_scope::{info, warn};
-use std::fs;
-use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
 
-const IGNITION_VERSION: &str = "3.0.0";
+use crate::ignition::IgnitionConfig;
+use crate::providers::MetadataProvider;
 
 /// Render Ignition config fragments from cloud provider metadata
 #[derive(Debug, Parser)]
@@ -80,78 +73,7 @@ impl CliRenderIgnition {
     }
 }
 
-#[derive(Debug, Serialize)]
-pub(crate) struct IgnitionConfig {
-    pub ignition: IgnitionMeta,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub storage: Option<Storage>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub passwd: Option<Passwd>,
-}
-
-#[derive(Debug, Serialize)]
-pub(crate) struct IgnitionMeta {
-    pub version: String,
-}
-
-#[derive(Debug, Serialize)]
-pub(crate) struct Storage {
-    pub files: Vec<StorageFile>,
-}
-
-#[derive(Debug, Serialize)]
-pub(crate) struct StorageFile {
-    pub path: String,
-    pub mode: u32,
-    pub overwrite: bool,
-    pub contents: FileContents,
-}
-
-#[derive(Debug, Serialize)]
-pub(crate) struct FileContents {
-    pub source: String,
-}
-
-#[derive(Debug, Serialize)]
-pub(crate) struct Passwd {
-    pub users: Vec<PasswdUser>,
-}
-
-#[derive(Debug, Serialize)]
-#[serde(rename_all = "camelCase")]
-pub(crate) struct PasswdUser {
-    name: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub ssh_authorized_keys: Option<Vec<String>>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub password_hash: Option<String>,
-}
-
-pub(crate) fn hostname_data_uri(hostname: &str) -> String {
-    let encoded =
-        percent_encoding::utf8_percent_encode(hostname, percent_encoding::NON_ALPHANUMERIC)
-            .to_string();
-    format!("data:,{encoded}")
-}
-
-pub(crate) fn write_fragment(config: &IgnitionConfig, path: &Path) -> Result<()> {
-    if let Some(parent) = path.parent() {
-        fs::create_dir_all(parent)
-            .with_context(|| format!("failed to create directory {}", parent.display()))?;
-    }
-    let json =
-        serde_json::to_string_pretty(config).context("failed to serialize ignition config")?;
-    fs::write(path, json.as_bytes())
-        .with_context(|| format!("failed to write {}", path.display()))?;
-    fs::set_permissions(path, fs::Permissions::from_mode(0o644))
-        .with_context(|| format!("failed to set permissions on {}", path.display()))?;
-    Ok(())
-}
-
-fn generate_hostname_fragment(
-    provider: &dyn crate::providers::MetadataProvider,
-    output_dir: &str,
-) -> Result<()> {
+fn generate_hostname_fragment(provider: &dyn MetadataProvider, output_dir: &str) -> Result<()> {
     let hostname = match provider.hostname()? {
         Some(h) => h,
         None => {
@@ -160,33 +82,13 @@ fn generate_hostname_fragment(
         }
     };
 
-    let config = IgnitionConfig {
-        ignition: IgnitionMeta {
-            version: IGNITION_VERSION.to_string(),
-        },
-        storage: Some(Storage {
-            files: vec![StorageFile {
-                path: "/etc/hostname".into(),
-                mode: 420,
-                overwrite: true,
-                contents: FileContents {
-                    source: hostname_data_uri(&hostname),
-                },
-            }],
-        }),
-        passwd: None,
-    };
-
     let path = Path::new(output_dir).join("hostname.ign");
-    write_fragment(&config, &path)?;
+    IgnitionConfig::hostname_fragment(&hostname).write_to(&path)?;
     info!("wrote hostname ignition fragment"; "path" => path.display().to_string());
     Ok(())
 }
 
-fn generate_user_fragment(
-    provider: &dyn crate::providers::MetadataProvider,
-    output_dir: &str,
-) -> Result<()> {
+fn generate_user_fragment(provider: &dyn MetadataProvider, output_dir: &str) -> Result<()> {
     let username = provider
         .admin_username()
         .context("failed to query admin username from provider")?;
@@ -209,26 +111,8 @@ fn generate_user_fragment(
         .admin_password_hash()
         .context("failed to query admin password hash from provider")?;
 
-    let config = IgnitionConfig {
-        ignition: IgnitionMeta {
-            version: IGNITION_VERSION.to_string(),
-        },
-        storage: None,
-        passwd: Some(Passwd {
-            users: vec![PasswdUser {
-                name: username,
-                ssh_authorized_keys: if ssh_keys.is_empty() {
-                    None
-                } else {
-                    Some(ssh_keys)
-                },
-                password_hash,
-            }],
-        }),
-    };
-
     let path = Path::new(output_dir).join("user.ign");
-    write_fragment(&config, &path)?;
+    IgnitionConfig::user_fragment(username, ssh_keys, password_hash).write_to(&path)?;
     info!("wrote platform-user ignition fragment"; "path" => path.display().to_string());
     Ok(())
 }
@@ -237,6 +121,7 @@ fn generate_user_fragment(
 mod tests {
     use super::*;
     use openssh_keys::PublicKey;
+    use std::fs;
 
     /// Minimal generic provider used to exercise the fragment generators
     /// without any platform-specific plumbing.
@@ -247,7 +132,7 @@ mod tests {
         admin_password_hash: Option<String>,
     }
 
-    impl crate::providers::MetadataProvider for FakeProvider {
+    impl MetadataProvider for FakeProvider {
         fn hostname(&self) -> Result<Option<String>> {
             Ok(self.hostname.clone())
         }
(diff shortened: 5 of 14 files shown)


prestist:
@peytonr18, I have started looking into the changes, I will add my feedback in the next day or so.

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,69 @@
...
+use anyhow::{Context, Result};
+use clap::{ArgGroup, Parser};
+
+/// Render Ignition config fragments from cloud provider metadata
+#[derive(Debug, Parser)]
+#[command(group(ArgGroup::new("provider-group").args(["cmdline", "provider"]).required(true)))]
+pub struct CliRenderIgnition {
+    /// The name of the cloud provider
+    #[arg(long, value_name = "name")]
+    provider: Option<String>,
```
prestist, line 18: So we are accepting any provider here it might make sense to limit it to azure only as that is the only intended support platform?

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    let path = Path::new(output_dir).join("hostname.ign");
+    write_fragment(&config, &path)?;
+    info!("wrote hostname ignition fragment"; "path" => path.display().to_string());
+    Ok(())
+}
+
+pub(crate) fn generate_user_fragment(
+    provider_id: &str,
+    _provider: &dyn crate::providers::MetadataProvider,
```
prestist: Its odd that we have the _provider but are not actually using it in this function.

Rather than creating a new IMDS client maybe we should create it in the provider and call through the _provider here.

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,480 @@
...
+    linux_prov_conf_set: LinuxProvisioningConfigurationSet,
+}
+
+#[derive(Debug, Deserialize)]
+struct LinuxProvisioningConfigurationSet {
+    #[serde(rename = "AdminPassword", alias = "adminPassword", default)]
+    admin_password: String,
+}
+
+fn imds_client() -> Result<retry::Client> {
```
prestist: Hmm, maybe we should configure the client to be in the mod.rs https://github.com/coreos/afterburn/blob/main/src/providers/microsoft/azure/mod.rs#L117

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,480 @@
...
+use slog_scope::{info, warn};
+use std::fs;
+use std::os::unix::fs::PermissionsExt;
+use std::path::Path;
+
+use crate::retry;
+use crate::util;
+
+const IGNITION_VERSION: &str = "3.0.0";
+const IMDS_ENDPOINT: &str = "http://169.254.169.254";
```
prestist: Hmm, just noting this is duplicate to the providers/microsoft/azure/mod.rs file.

prestist (commented):
Overall this looks good! just a few concerns around the CLI design and how the IMDS client is built.

dracut/30afterburn/module-setup.sh
```diff
@@ -23,9 +23,13 @@ install() {
...
 
+    inst_simple "$moddir/afterburn-ignition-fragment.service" \
+        "$systemdutildir/system/afterburn-ignition-fragment.service"
+
     # These services are only run once on first-boot, so they piggyback
     # on Ignition completion target.
     mkdir -p "$initdir/$systemdsystemunitdir/ignition-complete.target.requires"
     ln -s "../afterburn-hostname.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-hostname.service"
     ln -s "../afterburn-network-kargs.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-network-kargs.service"
+    ln -s "../afterburn-ignition-fragment.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-ignition-fragment.service"
```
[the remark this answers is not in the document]
prestist: Hmm, the ordering concern is valid but `ignition-kargs.service.requires` might also be too late. Looking at the Ignition boot flow (coreos/ignition#2195), both `ignition-fetch-offline.service` and `ignition-fetch.service` read from `base.platform.d/{platform}/` before `ignition-kargs.service` runs.

Maybe we should use `Before=ignition-fetch-offline.service` or similar to ensure the fragments are written before Ignition discovers them?

peytonr18:
@prestist I should have addressed all of your comments - thank you so much for these! I agreed with all of your points and adjusted the logic accordingly.

prestist:
@peytonr18 I will take another pass today! thank you for the quick changes!

prestist:
@peytonr18 we do have some failing lints in the CI and you would need to add an entry to the release notes file in the "upcoming" release notes section. https://github.com/coreos/afterburn/blob/main/docs/release-notes.md

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,438 @@
...
+    };
+
+    let path = Path::new(output_dir).join("user.ign");
+    write_fragment(&config, &path)?;
+    info!("wrote platform-user ignition fragment"; "path" => path.display().to_string());
+    Ok(())
+}
+
+/// OVF is optional; if present, only `adminPassword` is consulted.
+fn validate_ovf_admin_password_policy() -> Result<()> {
```
anhvoms: If Ignition supports configuring user and password, this should be supported (otherwise this will be a breaking change for existing customers when new version of afterburn + azure-init functionality are baked into the image as replacement for wala)

anhvoms (changes requested):

dracut/30afterburn/afterburn-ignition-fragment.service
```diff
@@ -0,0 +1,23 @@
...
+# still completing before ignition-kargs merge processing.
+Wants=network-online.target
+After=network-online.target
+Before=ignition-kargs.service
+
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+ExecStartPre=/usr/bin/mount -t tmpfs -o mode=0755,nodev,nosuid tmpfs /usr/lib/ignition/base.platform.d
```
dustymabe: obviously this isn't ideal but `/usr/ is definitely read-only here in the initramfs because of recent versions of systemd.

Some prior discussion on this in https://github.com/coreos/ignition/issues/1891
An example of a workaround: https://github.com/coreos/fedora-coreos-config/pull/3031


I wonder if we should invest in actually fixing https://github.com/coreos/ignition/issues/1891 now

peytonr18: This has been addressed with the merging of https://github.com/coreos/ignition/pull/2221.

src/providers/microsoft/azure/config.rs
peytonr18: I should have addressed this in the latest commit! adminPassword is now supported and hashed into an Ignition passwd entry, so existing customers using password-based provisioning won't break.

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,489 @@
...
+        );
+    }
+
+    let result = fs::read_to_string(mount_point.join("ovf-env.xml"));
+    let _ = util::unmount(mount_point, MOUNT_RETRIES);
+    result.context("failed to read ovf-env.xml")
+}
+
+fn parse_ovf_env(xml: &str) -> Result<OvfEnvironment> {
+    // Strip Azure `wa:` namespace prefixes so serde-xml-rs can match element names.
```
cjp256: does serde not have an option to deal with these or is it just simplified to do so?

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,489 @@
...
+    let clean = xml.replace("<wa:", "<").replace("</wa:", "</");
```
cjp256: i'd rather parse the xml and change the name space if required,  the text replacement is possibly going to blow up at some point matching on unrelated text

peytonr18: In the latest commit, I replaced the `string .replace()` with xml-rs `EventReader/EventWriter`, which parses the document structurally and rewrites using only local element names. `serde-xml-rs` doesn't support namespace-aware deserialization on its own unfortunately, so the preprocessing is still needed.

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,719 @@
...
+    let mut salt_bytes = [0u8; 12];
+    openssl::rand::rand_bytes(&mut salt_bytes)
+        .context("failed to generate random salt for password hash")?;
+    let salt = crypt_base64_encode(&salt_bytes);
+    sha512_crypt(password, &salt, PASSWORD_HASH_ROUNDS)
+}
+
+/// SHA-512 crypt implementation per the specification at
+/// <https://www.akkadia.org/dreez/SHA-crypt.txt>.
+fn sha512_crypt(password: &str, salt: &str, rounds: usize) -> Result<String> {
```
prestist: hmm, it feels weird to hand roll sha512; is there a crate we can use?

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,719 @@
...
+            ctx.update(&p_string);
+        }
+        digest_c = ctx.finalize().into();
+    }
+
+    let hash_str = crypt_base64_encode_final(&digest_c);
+
+    Ok(format!(
+        "$6$rounds={rounds}${salt}${hash_str}",
+        salt = std::str::from_utf8(salt).unwrap_or("")
```
prestist: hmm, should we maybe use "?" and create an error at this stage if we somehow hit one?

Cargo.lock
```diff
@@ -37,23 +37,26 @@ dependencies = [
  "mailparse",
  "maplit",
  "mockito",
- "nix 0.30.1",
+ "nix 0.29.0",
```
prestist: why did we downgrade?

prestist (commented):
Quick look over, a few small questions/nits

src/providers/microsoft/azure/config.rs
peytonr18: This was my attempt at using rust-sha2 and avoiding adding the sha-crypt library since that one isn't available in the upstream repo, but let me see if there is a better approach!

Edit: went ahead and just used `sha-crypt` again to avoid any manual logic.

Cargo.lock
peytonr18: I think the downgrade happened when I was merging main in. Looking at this again, I think `libsystemd v0.7.2` pins its dependency on nix 0.29, so Cargo resolves afterburn's direct nix dep to 0.29 as well and I had mismatched versioning before. I should have updated this!

src/providers/microsoft/azure/config.rs
peytonr18: Done!

Cargo.toml
```diff
@@ -46,14 +46,17 @@ maplit = "1.0"
...
 openssl = ">= 0.10.46, < 0.11"
+percent-encoding = "2.3.2"
 pnet_base = ">= 0.26, < 0.36"
 pnet_datalink = ">= 0.26, < 0.36"
 reqwest = { version = ">= 0.10, < 0.13", features = [ "blocking" ] }
 serde =  { version = "1.0", features = [ "derive" ] }
 serde-xml-rs = ">= 0.4, < 0.9"
 serde_json = "1.0"
 serde_yaml = ">= 0.8, < 0.10"
+sha-crypt = "0.5"
```
travier: We likely can not do that as this will not be FIPS compliant. Is there something in the openssl crate that we can use instead?

peytonr18: I swapped the `sha-crypt` crate for a small FFI call to the system's crypt(3) (libxcrypt) — same function passwd and PAM use to write shadow, so there isn't a new dependency to worry about. 

By my searching, openssl doesn't have a crypt equivalent, so going that route would mean re-implementing the password hashing scheme by hand, which I'd like to avoid. 

Let me know your thoughts on this, I was trying to find the middle ground in best practices here but I'm open to whatever implementation feels best for you all.

dracut/30afterburn/module-setup.sh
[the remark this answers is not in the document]
prestist: @peytonr18 how are we going to handle the ordering of this?

Cargo.lock
```diff
@@ -1664,9 +1667,9 @@ checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
 
 [[package]]
 name = "rand"
-version = "0.9.4"
+version = "0.9.2"
```
prestist: Hmm, downgrade?

prestist (dismissed):
From my perspective this looks sane, though I think we still have an ordering issue [here](https://github.com/coreos/afterburn/pull/1264/changes#r2961671116), wdyt?

dracut/30afterburn/module-setup.sh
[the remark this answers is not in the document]
peytonr18: You're right, the ordering we had was likely too late. 

I don't think we can go all the way to `Before=ignition-fetch-offline.service` because the service needs `network-online.target` for the IMDS call, and the network isn't up yet at that point. 

For Azure that's not really a big deal: the provider always requires the network, so fetch-offline exits early with `neednet` and `ignition-fetch.service` is the stage that matters.

With that being said, I think switching the ordering to `Before=ignition-fetch.service` is the best approach? Let me know your thoughts!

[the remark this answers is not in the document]
prestist: That's a great point, yeah I think that should work for this use case.

prestist (commented):
Thank you for all your work on this! from my perspective it lgtm, I would like @travier 's feedback as well.

dracut/30afterburn/afterburn-ignition-fragment.service
```diff
@@ -0,0 +1,22 @@
...
+Documentation=https://coreos.github.io/afterburn/
+
+# Only run on platforms that support ignition fragment generation.
+ConditionKernelCommandLine=|ignition.platform.id=azure
+
+# Run after networking is available (needed for the Azure IMDS call) and
+# before ignition-fetch.service, which produces /run/ignition.json from
+# /etc/ignition/base.platform.d/{platform}/.
+Wants=network-online.target
+After=network-online.target
```
cjp256: to clarify, the current ignition-fetch.service is After=network.target.  So this does somewhat indirectly change behavior if we order before fetch.service (it has been seen that fetch will fail while DHCP is ongoing).  There may be some implications to waiting for online target in multi-nic scenarios where maybe a subset don't dhcp [fast enough].

@prestist what ordering issue specifically are you referring to?

prestist: Hmm, so the concern I had was about module-setup.sh — the fragment service gets symlinked into ignition-complete.target.requires. That would mean the unit only gets pulled into the transaction when ignition-complete.target is being assembled, which is after Ignition has already read from base.platform.d/ and produced /run/ignition.json. The fragments wouldn't be consumed at that point.
  
Its entirely possible I have a misunderstanding here, but I think it might make sense to ensure this service runs before ignition-fetch-offline.service or ignition-fetch.service, since those are what read base.platform.d/{platform}/.

On the network-online.target vs network.target point — good point. If we add Before=ignition-fetch.service here while keeping After=network-online.target, that would transitively force fetch to also wait for network-online.target, which could be a problem in multi-NIC setups where not all interfaces come up quickly. IMDS is a link-local endpoint right? So it should work once the primary interface has an address — would network.target be sufficient here instead?

cjp256: I think they read base.platform.d entries (it's certainly logged), but I'm unsure that it does anything with it.  Can you look into if these parameters are used at all @peytonr18?

prestist: Looking at ignition the fragments are actively merged, not just logged.
  
[FetchBaseConfig()](https://github.com/coreos/ignition/blob/main/internal/providers/system/system.go#L49-L60) reads .ign files from base.platform.d/{platform}/ across all system config dirs (/run/ignition, /etc/ignition, /usr/lib/ignition), parses them as Ignition configs, and merges them via latest.Merge().

  [engine.go:Run()](https://github.com/coreos/ignition/blob/main/internal/exec/engine.go#L75-L128) calls FetchBaseConfig at the start of every stage, then merges the result into the final config at line [127](https://github.com/coreos/ignition/blob/main/internal/exec/engine.go#L127): fullConfig := latest.Merge(baseConfig, latest.Merge(systemBaseConfig, cfg))

Since FetchBaseConfig runs at the start of every stage including fetch-offline, the afterburn fragment service would need to complete before ignition-fetch-offline.service to ensure the fragments are present when Ignition first reads them.

cjp256: In practice though, that fullConfig has little or no impact to behavior.

fetch-offline:
https://github.com/coreos/ignition/blob/e3c445c673d2b5c9af4aa61836bd5911a9bc4c30/internal/exec/stages/fetch_offline/fetch-offline.go#L71

used to determine if it needs networking at this stage (which results in offline bailing).

fetch:
https://github.com/coreos/ignition/blob/e3c445c673d2b5c9af4aa61836bd5911a9bc4c30/internal/exec/stages/fetch/fetch.go#L66

no-op.

prestist: Very good point, you're right that the fetch stages don't act on the merged config in practice. 
  
I think the stage that actually matters is files, where createPasswd() and createFilesystemsEntries() [run](https://github.com/coreos/ignition/blob/main/internal/exec/stages/files/files.go#L83-L95). So the fragments just need to exist before ignition-files.service.

On the network-online.target vs network.target question — since IMDS is link-local (169.254.169.254) and afterburn already has retry logic, would network.target be sufficient here? That'd be consistent with how ignition-fetch.service handles it and would avoid the multi-NIC stalling issue.

I think I confused this thread a bit, I want to make sure so going back to the ordering concern I raised earlier (https://github.com/coreos/afterburn/pull/1264#discussion_r2961671116) — I think the module-setup.sh symlink is still relevant here. The fragment service is currently linked into ignition-complete.target.requires, but ignition-files.service has [Before=ignition-complete.target](https://github.com/coreos/ignition/blob/main/dracut/30ignition/ignition-files.service#L6). Wouldn't that mean the files stage finishes before systemd even pulls in the fragment service? If so, would switching the symlink to ignition-files.service.requires make more sense?

@cjp256

cjp256: Stage order: fetch-offline [-> fetch] [-> kargs] -> disks -> mount -> files.

https://github.com/coreos/ignition/blob/e3c445c673d2b5c9af4aa61836bd5911a9bc4c30/dracut/30ignition/ignition-files.service#L11

Given this, we will want to order before disks.   ignition-complete.target requires would be fine, but it would be redundant once we order before ignition-disks.

I would take the above bot suggestion and just switch from kargs -> disks, because kargs is optional and we don't actually care to order before that either (unless someone knows a good reason to).  We do need to order before disks (and implicitly, mount and files).

prestist: That makes sense; Before=ignition-disks.service covers disks, mount, and files implicitly through the chain. And agreed, no reason to order before kargs since it's optional and the fragments don't contain kargs config anyway.
  
Oook, So to summarize, the service unit would need Before=ignition-disks.service instead of Before=ignition-kargs.service, and it might make sense to switch After=network-online.target to After=network.target to stay consistent with how ignition-fetch.service handles it and avoid the multi-NIC stalling concern.

peytonr18: That's what my takeaways were as well - will get the files updated asap!

dracut/30afterburn/module-setup.sh
```diff
@@ -23,9 +23,15 @@ install() {
...
+        "$systemdutildir/system/afterburn-ignition-fragment.service"
+
     # These services are only run once on first-boot, so they piggyback
     # on Ignition completion target.
     mkdir -p "$initdir/$systemdsystemunitdir/ignition-complete.target.requires"
     ln -s "../afterburn-hostname.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-hostname.service"
     ln -s "../afterburn-network-kargs.service" "$initdir/$systemdsystemunitdir/ignition-complete.target.requires/afterburn-network-kargs.service"
+
+    mkdir -p "$initdir/$systemdsystemunitdir/ignition-files.service.requires"
+    ln -s "../afterburn-ignition-fragment.service" "$initdir/$systemdsystemunitdir/ignition-files.service.requires/afterburn-ignition-fragment.service"
```
cjp256: keep this tied to ignition-complete.target.  the before/after config will order it correctly.   this is just to enable the unit as being enabled if ignition-complete.target is enabled.

cjp256: (just verify that this unit runs parallel to ignition-fetch now)

dracut/30afterburn/afterburn-ignition-fragment.service
```diff
@@ -0,0 +1,26 @@
...
+
+# Run after networking is available (needed for the Azure IMDS call). Use
+# network.target (not network-online.target) to match ignition-fetch.service
+# and avoid stalling on multi-NIC DHCP; IMDS is link-local and afterburn
+# has its own retry logic.
+#
+# Order before ignition-disks.service so fragments exist before the
+# disks/mount/files stages run (files is where they're consumed).
+# ignition-kargs is optional and does not consume these fragments.
+After=network.target
```
cjp256, line 15: upon further consideration, if we are in parallel with other ignition services, is that are in the middle of writing the config fragment when an ignition service is loading the configs and breaks the parser.  unlikely race condition but I suspect possible nonetheless.  we would want to be sure the config write is atomic (file written before moved into target location), or maybe order in between fetch.service and disks.service.

peytonr18: Addressed in latest commit!

dracut/30afterburn/module-setup.sh
peytonr18: Addressed in latest commit!

dracut/30afterburn/afterburn-ignition-fragment.service
cjp256: sorry, let's include after=ignition-kargs.service too.

peytonr18: All good, change has been made!

prestist (approved):
This looks wonderful thank you @peytonr18. I dont see anything that needs to change outside of a small nit about the commit structure. Since we merge directly into the main branch lets squash and rename some of the commits to follow the repo convention of <subsystem>: <Description> (lowercase subsystem, imperative tense). Something
  like:

  - azure: Add render-ignition subcommand for Ignition fragment generation
  - dracut: Add afterburn-ignition-fragment.service ordered between fetch and disks

  Or if you'd prefer to keep more granularity, each commit should at minimum get a subsystem prefix (e.g. azure:, dracut:, cli:).

peytonr18:
@prestist Should have changed the git commit history based on that suggestion, thanks!

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,715 @@
...
+}
+
+pub(crate) fn hostname_data_uri(hostname: &str) -> String {
+    let encoded =
+        percent_encoding::utf8_percent_encode(hostname, percent_encoding::NON_ALPHANUMERIC)
+            .to_string();
+    format!("data:,{encoded}")
+}
+
+pub(crate) fn write_fragment(config: &IgnitionConfig, path: &Path) -> Result<()> {
```
cjp256: should this live in render_ignition?

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Directory to write Ignition config fragment files into
+    #[arg(long = "render-ignition-dir", value_name = "path")]
+    render_ignition_dir: String,
+    /// Include hostname in a fragment file
+    #[arg(long)]
+    hostname: bool,
+    /// Include platform user and SSH keys in a fragment file
+    #[arg(long)]
+    platform_user: bool,
+    /// Enable all platform extensions (implies --hostname --platform-user)
```
cjp256: [draft]

in other words:

enable_hostname_fragment
enable_user_fragment
enable_all_fragments


from a cli perspective if i disable all, i would just disable the service via systemd instead.

if i enable the service (as an opt in) I would likely want it to run all of its intended functionality for providers that support it.  from that perspective, perhaps all should be enabled by default and provide something like:

--disable-hostname-fragment
--disable-user-fragment

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,78 @@
...
+        if !hostname && !platform_user {
+            slog_scope::warn!("render-ignition: no features specified");
+            return Ok(());
+        }
+
+        let metadata = crate::metadata::fetch_metadata(&provider_id)
+            .context("fetching metadata from provider")?;
+
+        if hostname {
+            crate::providers::microsoft::azure::config::generate_hostname_fragment(
```
cjp256: this is azure-specific call in cli code.  this should be generic to provider

peytonr18: Quick design question: after our offline discussion, `ignition.rs` holds the Ignition config structs + serialization (with the `write_to()` method on `IgnitionConfig`), and the `render-ignition` CLI keeps small `generate_hostname_fragment` / `generate_user_fragment` functions that pull data off the provider and write the fragments.

An alternative would be to move that provider→fragment orchestration onto the `MetadataProvider` trait as default methods, mirroring the existing `write_hostname` / `write_network_units` helpers — e.g.:

```rust
// src/providers/mod.rs
fn write_hostname_ignition_fragment(&self, output_dir: &str) -> Result<()> {
    let Some(hostname) = self.hostname()? else { return Ok(()); };
    let path = Path::new(output_dir).join("hostname.ign");
    IgnitionConfig::hostname_fragment(&hostname).write_to(&path)
}

fn write_user_ignition_fragment(&self, output_dir: &str) -> Result<()> {
    // admin_username / ssh_keys / admin_password_hash -> IgnitionConfig::user_fragment(..).write_to(..)
}
```

Do you think the trait-default-method approach would be cleaner/preferred here, or is the current split fine? I'm happy to switch the approach if the trait method seems better - it's a small change either way.

prestist:
@peytonr18, this still has my stamp :) from my perspective we should be able to merge.

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,239 @@
...
+                Self::SUPPORTED_PROVIDERS,
+                provider_id,
+            );
+        }
+
+        let hostname = !self.disable_hostname_fragment;
+        let platform_user = !self.disable_user_fragment;
+
+        if !hostname && !platform_user {
+            slog_scope::warn!("render-ignition: all fragments disabled, nothing to do");
```
Rolv-Apneseth:
```suggestion
            warn!("render-ignition: all fragments disabled, nothing to do");
```

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,239 @@
...
+            generate_user_fragment(metadata.as_ref(), &self.render_ignition_dir)
+                .context("generating platform-user ignition fragment")?;
+        }
+
+        Ok(())
+    }
+}
+
+fn generate_hostname_fragment(provider: &dyn MetadataProvider, output_dir: &str) -> Result<()> {
+    let hostname = match provider.hostname()? {
```
Rolv-Apneseth:
```suggestion
    let Some(hostname) = provider.hostname()? else {
        warn!("hostname requested, but not available from this provider");
        return Ok(());
    };
```

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,239 @@
...
+    info!("wrote hostname ignition fragment"; "path" => path.display().to_string());
+    Ok(())
+}
+
+fn generate_user_fragment(provider: &dyn MetadataProvider, output_dir: &str) -> Result<()> {
+    let username = provider
+        .admin_username()
+        .context("failed to query admin username from provider")?;
+    let username = match username {
+        Some(u) => u,
```
Rolv-Apneseth:
```suggestion
    let Some(username) = username else {
        warn!("platform-user requested, but admin username not available from this provider");
        return Ok(());
    };
```

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,239 @@
...
+        .ssh_keys()
+        .context("failed to query SSH keys from provider")?
+        .into_iter()
+        .map(|k| k.to_key_format())
+        .collect();
+
+    let password_hash = provider
+        .admin_password_hash()
+        .context("failed to query admin password hash from provider")?;
+
```
Rolv-Apneseth: Might be worth checking if any credentials are actually available before writing, WDYT?

```suggestion

    if ssh_keys.is_empty() && password_hash.is_none() {
        warn!("admin username present but no SSH keys or password; skipping user fragment");
        return Ok(());
    }
```

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,239 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use openssh_keys::PublicKey;
+    use std::fs;
+
+    /// Minimal generic provider used to exercise the fragment generators
+    /// without any platform-specific plumbing.
+    struct FakeProvider {
```
Rolv-Apneseth, line 84: Some of the boilerplate below could potentially be cut down with a `#[derive(Default)]`, e.g. 

```rust
let provider = FakeProvider {
    hostname: Some("myhost".into()),
    ..Default::default()
};
// Or, if no differing fields are required
let provider = FakeProvider::default();
```

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,239 @@
...
+        }
+        fn admin_username(&self) -> Result<Option<String>> {
+            Ok(self.admin_username.clone())
+        }
+        fn ssh_keys(&self) -> Result<Vec<PublicKey>> {
+            self.ssh_keys
+                .iter()
+                .map(|s| {
+                    s.parse::<PublicKey>()
+                        .map_err(|e| anyhow::anyhow!("failed to parse test ssh key: {e}"))
```
Rolv-Apneseth:
```suggestion
                        .context("failed to parse test ssh key")
```

src/providers/microsoft/azure/config.rs
```diff
@@ -0,0 +1,420 @@
+// Copyright 2026 CoreOS, Inc.
```
Rolv-Apneseth, line 1: Might be worth adding some `SAFETY` comments to the `unsafe` calls in this file to help us understand the assumptions in case there are any issues. Would you agree or do you think it's fine @prestist

src/providers/microsoft/azure/mod.rs
```diff
@@ -282,8 +330,28 @@ impl Azure {
...
+            .clone()
+            .header(
+                HeaderName::from_static("metadata"),
+                HeaderValue::from_static("true"),
+            )
+            .get(retry::Raw, url)
+            .send::<String>()
+            .context("failed to query IMDS for adminUsername")?;
+
+        match username {
```
Rolv-Apneseth:
```suggestion
        match username.map(|u| u.trim().to_owned()) {
            Some(u) if !u.is_empty() => Ok(Some(u)),
            _ => Ok(None),
        }
```

src/ignition.rs
```diff
@@ -0,0 +1,334 @@
+// Copyright 2017 CoreOS, Inc.
```
Rolv-Apneseth:
```suggestion
// Copyright 2026 CoreOS, Inc.
```

src/ignition.rs
```diff
@@ -0,0 +1,334 @@
...
+    /// storage file.
+    pub(crate) fn hostname_fragment(hostname: &str) -> Self {
+        IgnitionConfig {
+            ignition: IgnitionMeta {
+                version: IGNITION_VERSION.to_string(),
+            },
+            storage: Some(Storage {
+                files: vec![StorageFile {
+                    path: "/etc/hostname".into(),
+                    mode: 420,
```
Rolv-Apneseth: I think this is clearer in octal:

```suggestion
                    mode: 0o644,
```

src/ignition.rs
```diff
@@ -0,0 +1,334 @@
...
+    /// Serialize this config and write it as an Ignition fragment file at
+    /// `path`, creating parent directories and setting mode 0644.
+    pub(crate) fn write_to(&self, path: &Path) -> Result<()> {
+        if let Some(parent) = path.parent() {
+            fs::create_dir_all(parent)
+                .with_context(|| format!("failed to create directory {}", parent.display()))?;
+        }
+        let json =
+            serde_json::to_string_pretty(self).context("failed to serialize ignition config")?;
+        fs::write(path, json.as_bytes())
```
Rolv-Apneseth: Would it be worth writing to a temp file and renaming after the permissions are set to keep this atomic in case of a failure / crash?

src/ignition.rs
```diff
@@ -0,0 +1,334 @@
...
+    #[test]
+    fn test_hostname_storage_fragment_serialization() {
+        let cfg = IgnitionConfig {
+            ignition: IgnitionMeta {
+                version: "3.0.0".into(),
+            },
+            storage: Some(Storage {
+                files: vec![StorageFile {
+                    path: "/etc/hostname".into(),
+                    mode: 420,
```
Rolv-Apneseth:
```suggestion
                    mode: 0o644,
```

Rolv-Apneseth (commented):
Thanks for working on this @peytonr18 . Mostly LGTM but I have a couple questions and suggestions. Realistically though most are nits and not essential.

src/cli/render_ignition.rs
Rolv-Apneseth: I think defaults defined in the trait would be preferable but I would like to hear @prestist's thoughts on the subject

src/ignition.rs
cjp256: I think that's a good idea.  Probably worth a helper function so it can be re-used.

src/cli/render_ignition.rs
Rolv-Apneseth: Feel free to leave this as is if he doesn't see this, we can always do something like that in a follow-up

src/cli/render_ignition.rs
Rolv-Apneseth: hmm, seems some of these didn't get applied to the correct range of lines. Apply manually if you do choose to take these suggestions

yasminvalim:
Hey @peytonr18, I took a look in this PR and looks good. Thank you for working on this!
I just made a question in the code, but its just a question. Overall looks nice!

src/cli/render_ignition.rs
```diff
@@ -0,0 +1,239 @@
...
+        if !hostname && !platform_user {
+            slog_scope::warn!("render-ignition: all fragments disabled, nothing to do");
+            return Ok(());
+        }
+
+        let metadata = crate::metadata::fetch_metadata(&provider_id)
+            .context("fetching metadata from provider")?;
+
+        if hostname {
+            generate_hostname_fragment(metadata.as_ref(), &self.render_ignition_dir)
```
yasminvalim: As far I understood the code, `write_hostname() `truncates to HOST_NAME_MAX, but the new hostname fragment uses the raw value. Was skipping truncation on purpose, or should the new path shorten the hostname the same way as the old one?

src/cli/render_ignition.rs
peytonr18: Addressed in latest commit, thanks!

src/cli/render_ignition.rs
peytonr18: Addressed in latest commit, thanks!

src/ignition.rs
peytonr18: Addressed in latest commit, thanks!

src/providers/microsoft/azure/config.rs
peytonr18: I went ahead and added them, just in case! Figured we could always remove them if they feel unnecessary.

src/providers/microsoft/azure/mod.rs
peytonr18: Addressed in latest commit, thanks!

src/ignition.rs
peytonr18: Addressed in latest commit, thanks!

src/ignition.rs
peytonr18: Addressed in latest commit, thanks!

src/cli/render_ignition.rs
peytonr18: Good catch — that was an oversight on my part and not intentional. The fragment path should behave like write_hostname, so I pulled the truncation into a shared truncate_hostname helper that both now use. It only shortens hostnames exceeding HOST_NAME_MAX (to the first dot, or to the max length if there's no dot in range); anything within the limit is written verbatim. Both paths run against the same kernel, so they'll make the same decision.

src/ignition.rs
peytonr18: Done in the latest commit! I factored into a reusable helper per `util::write_file_atomic(path, contents, mode)` writes to a NamedTempFile in the target directory, sets the mode, then atomically renames it into place. This way, Ignition never sees a partial file and the final file never briefly carries the temp's default perms. `IgnitionConfig::write_to` now just serializes and calls it, so `hostname.ign` and `user.ign` share one path and it's available for future write paths to reuse.

src/cli/render_ignition.rs
peytonr18: I agree with this!

I added a guard so that after collecting the username, SSH keys, and password hash, if there are no keys and no password, it logs a warning and skips `user.ign` instead of emitting a user with no way to authenticate. I also covered this logic with a new `test_write_user_ignition_fragment_skipped_without_credentials`.

src/cli/render_ignition.rs
peytonr18: Done — FakeProvider now derives Default, and the tests use ..Default::default() (or FakeProvider::default() when no fields differ). Thanks!

src/cli/render_ignition.rs
peytonr18: I applied this change! This function moved out of the CLI onto the MetadataProvider trait (the "generic to provider" change), so it now lives in write_hostname_ignition_fragment; but the change is applied in the new location.

src/cli/render_ignition.rs
peytonr18: Applied - likewise moved onto the trait in `write_user_ignition_fragment`.

src/cli/render_ignition.rs
peytonr18: I ended up going with the trait-default-method approach, but if people end up preferring the small free functions in the CLI instead, I can always change it back!

`write_hostname_ignition_fragment` / `write_user_ignition_fragment` are now default methods on `MetadataProvider` (mirroring `write_hostname` / `write_network_units`), and `render-ignition` just does:

```
if hostname {
    metadata.write_hostname_ignition_fragment(&self.render_ignition_dir)?;
}
if platform_user {
    metadata.write_user_ignition_fragment(&self.render_ignition_dir)?;
}
```

So there's no Azure- (or provider-) specific call in the CLI anymore and data is still sourced generically through the trait (`hostname()` / `admin_username()` / `ssh_keys()` / `admin_password_hash()`). The hostname method reuses the shared `truncate_hostname` helper, and the user method skips writing when there are no credentials.

peytonr18:
Thank you for your reviews @yasminvalim, @Rolv-Apneseth, @cjp256, and @prestist!
I should have addressed all the comments, but let me know if there's anything outstanding that I missed. I really appreciate the time spent on this!

Cargo.toml
```diff
@@ -55,6 +57,7 @@ serde-xml-rs = ">= 0.4, < 0.9"
 serde_json = "1.0"
 serde_yaml = ">= 0.8, < 0.10"
 slog = { version = "2.7", features = ["max_level_trace", "release_max_level_info"] }
+xml-rs = "0.8"
```
Rolv-Apneseth: Seems this one is out of alphabetical order - sorry if I just missed that on the last review

Rolv-Apneseth (approved):
LGTM just the nit on alphabetical ordering

Cargo.toml
peytonr18: All good! Just made that change.

Rolv-Apneseth (approved):

Rolv-Apneseth:
>  ___________________________________________________________________
> < This refactor is like rearranging furniture during an earthquake. >
>  -------------------------------------------------------------------
>   \
>    \   (\__/)
>        (•ㅅ•)
>        /   づ

Bad bot

src/ignition.rs
```diff
@@ -0,0 +1,329 @@
...
+    /// `path`, creating parent directories and setting mode 0644. The file is
+    /// written atomically so a reader never observes a partially written config.
+    pub(crate) fn write_to(&self, path: &Path) -> Result<()> {
+        let json =
+            serde_json::to_string_pretty(self).context("failed to serialize ignition config")?;
+        crate::util::write_file_atomic(path, json.as_bytes(), 0o644)
+            .with_context(|| format!("failed to write {}", path.display()))?;
+        Ok(())
+    }
+}
```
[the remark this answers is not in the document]
peytonr18: One correction for the record: this runs in the initramfs, so user.ign is written to the ephemeral in-RAM etc, consumed by Ignition as root, and discarded at switch-root. All to say, it isn't written to the persistent root, so it's not reachable by an unprivileged user. But, I went ahead and made this change for defense-in-depth / future-proofing rather than a fix for an exploitable leak.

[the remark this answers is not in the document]
peytonr18: Intentionally added this as its own commit in case we want to revert it / don't think that much defense in depth is required here.

yasminvalim (approved):
Thanks for working on that @peytonr18. It's LGTM! Nice work :)

src/cli/render_ignition.rs
prestist: Trait defaults is absolutely cleaner, thank you for doing that!

prestist (approved):
LGTM!!

PR #1266, opened 2026-03-23
providers/hetzner: add network configuration and public IPv6 attribute

### LINKED ISSUE #1141: Support IPv6 on Hetzner
# Feature Request #

Afterburn currently supports only IPv4 for Hetzner. IPv6 metadata can be retrieved from Hetzner's API endpoints (`/hetzner/v1/metadata` or `/hetzner/v1/metadata/network-config`) in the `network-config.config[0].subnets[n]` property, where `ipv6` is set to `true`.

## Environment ##

Hetzner

## Desired Feature ##

Enable Afterburn to configure IPv6 automatically on server startup.

## Other Information ##

- If no public network is configured, the config property in network-config will be an empty array.
- An example value of the `network-config` field is shown below.
```
network-config:
  config:
  - mac_address: <redacted>
    name: eth0
    subnets:
    - ipv4: true
      type: dhcp
    - address: 2a01:<redacted>::1/64
      dns_nameservers:
      - 2a01:4ff:ff00::add:1
      - 2a01:4ff:ff00::add:2
      gateway: fe80::1
      ipv6: true
      type: static
    type: physical
  version: 1
```

Rolv-Apneseth:
Closes #1141

This adds support for configuring networking on Hetzner systems, both via `systemd-networkd` unit files and `netplan` configurations, as well as providing a `HETZNER_PUBLIC_IPV6` attribute. This should be enough for, e.g. Flatcar, but Fedora CoreOS (`NetworkManager`) requires some more work.

The main issue is that unlike other providers, it seems (as far as I can tell from everywhere I've looked) that Hetzner only provides the networking metadata from a HTTP endpoint, meaning you need some form of networking up first before further configuration. To my understanding, this means we can't use the `kargs`  approach like with other providers, and need to instead fallback to the automatic network configuration before making a request to the metadata endpoint (which is available locally even on systems with only a public IPv6).

I've confirmed that the `netplan` configuration can work on FCOS using:

```
sudo afterburn multi --cmdline --netplan-config /etc/netplan
# Hetzner reports iface name as eth0 but system uses predictable naming
sudo sed -i 's/eth0:/<actual_interface_name>:/' /etc/netplan/50-afterburn.yaml
echo '  renderer: NetworkManager' | sudo tee -a /etc/netplan/50-afterburn.yaml
sudo rpm-ostree install netplan
sudo reboot
...
ping6 google.com
```

And I'm planning on creating a separate draft PR for generating NetworkManager configurations, which I've also confirmed works with these changes and which I believe would be very nice to have in `afterburn` regardless. I'll create an issue for discussion though.

On a side note, the `netplan` configuration was largely copied from another provider in the codebase. I have some code ready for a default implementation, if we think that's a good change.


src/network.rs
```diff
@@ -153,6 +153,15 @@ impl DhcpSetting {
...
     }
+
+    /// Combine DHCP settings, to easily merge [`Self::V4`] and [`Self::V6`] variants into
+    /// [`Self::Both`] where applicable.
+    pub fn merge(self, other: Self) -> Self {
+        match (self, other) {
+            s @ ((DhcpSetting::V4, DhcpSetting::V4) | (DhcpSetting::V6, DhcpSetting::V6)) => s.0,
+            _ => DhcpSetting::Both,
+        }
+    }
```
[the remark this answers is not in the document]
Rolv-Apneseth: Fair enough

src/providers/hetzner/mod.rs
```diff
@@ -98,6 +114,105 @@ impl MetadataProvider for HetznerProvider {
...
+        }
+
+        let network = serde_yaml::Mapping::from_iter([
+            ("version".into(), 2.into()),
+            ("ethernets".into(), ethernets.into()),
+        ]);
+        let netplan = serde_yaml::Mapping::from_iter([("network".into(), network.into())]);
+
+        Ok(Some(serde_yaml::to_string(&netplan)?))
+    }
```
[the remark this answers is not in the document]
Rolv-Apneseth: Basically copied from an existing provider and this is how it was done, though I kind of agree

src/providers/hetzner/mod.rs
```diff
@@ -113,6 +228,151 @@ struct Metadata {
...
+}
+
+#[derive(Debug, Deserialize)]
+struct SubnetConfig {
+    #[serde(rename = "type")]
+    subnet_type: String,
+    #[allow(unused)]
+    ipv4: Option<bool>,
+    #[allow(unused)]
```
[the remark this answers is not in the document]
Rolv-Apneseth: Alright

src/providers/hetzner/mod.rs
```diff
@@ -146,6 +406,14 @@ impl From<Attributes> for HashMap<String, String> {
             "HETZNER_PUBLIC_IPV4",
             attributes.metadata.public_ipv4,
         );
+        add_value(
+            &mut out,
+            "HETZNER_PUBLIC_IPV6",
+            attributes
+                .metadata
+                .network_config
+                .and_then(|n| n.get_public_ipv6()),
```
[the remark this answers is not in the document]
Rolv-Apneseth: Every other attribute there is also relying on partial moves

src/providers/hetzner/mod.rs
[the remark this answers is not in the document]
Rolv-Apneseth: Might be more useful with a default implementation though, rather than in this PR

src/providers/hetzner/mod.rs
```diff
@@ -113,6 +228,150 @@ struct Metadata {
...
+                        IpNetwork::from_str("0.0.0.0/0")?
+                    };
+
+                    iface.routes.push(NetworkRoute {
+                        destination,
+                        gateway,
+                    });
+                }
+
+                "dhcp" | "dhcp4" | "dhcp6" => {
```
prestist, line 348: hmm I think this is correct, but what a weird state lol... if subnet is set to "dhcp6" but our ipv6=false we end up with dhcp:4?

So by this statement we are saying that ipv6 takes the lead not dhcp.

src/providers/hetzner/mod.rs
```diff
@@ -98,6 +114,105 @@ impl MetadataProvider for HetznerProvider {
...
+                    "nameservers".into(),
+                    serde_yaml::Value::Mapping(serde_yaml::Mapping::from_iter([(
+                        "addresses".into(),
+                        nameservers.into(),
+                    )])),
+                );
+            }
+
+            if let Some(name) = iface.name {
+                ethernets.insert(name.into(), eth_config.into());
```
prestist: I think it could be good to add a warning here since we are dropping whole eth_configs if they are not named? it could be confusing wdyt?

src/providers/hetzner/mod.rs
```diff
@@ -113,6 +228,150 @@ struct Metadata {
...
+            .find(|n| n.network_type == "physical")?
+            .subnets
+            .iter()
+            .find(|sub| sub.ipv6.is_some_and(|b| b))?
+            .address
+            .as_ref()?;
+
+        match IpNetwork::from_str(addr_str).ok()? {
+            IpNetwork::V4(_) => {
+                error!("IPv4 address in IPv6");
```
prestist: Feels rough, maybe "Expected ipv6 but found ipv4 {addr_str}"?

src/providers/hetzner/mod.rs
```diff
@@ -15,25 +15,41 @@
...
+// cloud-init tries http://[fe80::a9fe:a9fe%<interface>]/hetzner/v1/metadata
+// first but Hetzner has no IPv6 address. However, even VMs without
+// public IPv4 addresses still have link-local access to fetch this
+// metadata.
 const HETZNER_METADATA_BASE_URL: &str = "http://169.254.169.254/hetzner/v1/metadata";
 
 /// Metadata provider for Hetzner Cloud
 ///
 /// See: https://docs.hetzner.cloud/#server-metadata
-#[derive(Clone, Debug)]
```
prestist, line 36: I dont think this is wrong, but why did we remove "Clone"?

prestist (commented):
Overall this looks really good @Rolv-Apneseth  great work! I love the test coverage!

src/providers/hetzner/mod.rs
Rolv-Apneseth: Yeah... not ideal. Those `ipv4` and `ipv6` booleans are a deviation from [cloud-init v1 network config](https://docs.cloud-init.io/en/latest/reference/network-config-format-v1.html) and I've only seen them set to `true`. I don't even know if they actually use `dhcp6`, I just thought it should be covered in case.

src/providers/hetzner/mod.rs
Rolv-Apneseth: Sure, I'll add a warning

src/providers/hetzner/mod.rs
Rolv-Apneseth: Yeah makes sense, thanks

src/providers/hetzner/mod.rs
Rolv-Apneseth: It just wasn't required, but I can keep it if you want? A test was using it when it didn't need to.

src/providers/hetzner/mod.rs
Rolv-Apneseth: Went ahead and did the name check + skip earlier, with a warning

src/providers/hetzner/mod.rs
prestist: Nah no need to keep dead code :) happy to drop it

Rolv-Apneseth:
@apricote, in relation to https://github.com/flatcar/Flatcar/issues/1968, if possible could you confirm this would work for Flatcar?

Edit: @tormath1 might be more relevant actually, sorry.

prestist (approved):
Alright, I think we can merge this as it is! thank you for all your work on this @Rolv-Apneseth

tormath1:
@Rolv-Apneseth thank you so much for this, I was AFK in the last weeks and I missed this ping. Thanks to @prestist this landed in my mailbox again.

Better late than never, I kicked-off a Flatcar build to test this patch in Hetzner CI + run manual tests to see if IPv6 is correctly fetched. I'll report back in this comment.

https://github.com/flatcar/scripts/pull/4012

Rolv-Apneseth:
@tormath1 No problem, and thank you for testing. Hope it works as intended.

tormath1:
Flatcar CI on Hetzner is all green and manual tests have shown the correct IPv6 in the Afterburn metadata. With a simple systemd helper one can enable IPv6 connection to its Hetzner instance. :)

Rolv-Apneseth:
> Flatcar CI on Hetzner is all green and manual tests have shown the correct IPv6 in the Afterburn metadata. With a simple systemd helper one can enable IPv6 connection to its Hetzner instance. :)

Great to hear, thanks for the confirmation

PR #1267, opened 2026-03-23
NetworkManager connection profile (keyfile) generation

Rolv-Apneseth:
I will be creating an issue for discussion, but as mentioned in #1266, I think this would be nice to have in `afterburn`. It is particularly important for the Hetzner implementation though as we need to fallback to IPv4 DHCP initially in order to request the network configuration metadata from Hetzner's locally available HTTP endpoint, meaning that as far as I understand it passing network kargs to initrd can't work.

With these changes and the Hetner PR, I've managed to configure an FCOS Hetzner system using the following butane snippet:

```yaml
systemd:
  units:
    - name: afterburn-networkmanager.service
      enabled: true
      contents: |
        [Unit]
        Description=Afterburn Network Configuration First Boot (Hetzner)
        Documentation=https://coreos.github.io/afterburn/
        After=network-online.target
        Wants=network-online.target
        ConditionKernelCommandLine=ignition.platform.id=hetzner
        ConditionPathExists=!/var/lib/afterburn-network-configured

        [Service]
        Type=oneshot
        ExecStart=/usr/bin/afterburn multi \
          --cmdline \
          --networkmanager-profiles /etc/NetworkManager/system-connections
        ExecStartPost=/usr/bin/touch /var/lib/afterburn-network-configured
        # Reboot to apply new network config
        ExecStartPost=/usr/bin/systemctl reboot
        RemainAfterExit=yes

        [Install]
        WantedBy=multi-user.target
```

If we do go ahead with adding this as a feature, I would appreciate any feedback on the code.

Also, I've initially pushed 2 commits, with the difference being how VLANs/Bonds are handled. The first commit tries to directly translate from the `VirtualNetDev::sd_netdev_sections` to the NetworkManager equivalents, though this had some limitations due to the fundamentally different 1 file vs several per connection, and required a lot of conversions of both keys and values. The second commit instead shifts the burden to the provider implementation, requiring a new `VirtualNetDev::nm_sections`, which comes out a lot simpler overall in my opinion but I'm open to suggestions.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -19,6 +19,7 @@
 use anyhow::{anyhow, bail, Context, Result};
 use ipnetwork::IpNetwork;
 use pnet_base::MacAddr;
+use slog_scope::warn;
 use std::fmt::Write;
 use std::net::IpAddr;
 use std::string::String;
@@ -113,14 +114,52 @@ pub struct VirtualNetDev {
     pub kind: NetDevKind,
     pub mac_address: MacAddr,
     pub priority: Option<u32>,
-    pub sd_netdev_sections: Vec<SdSection>,
+    pub custom_sections: CustomNetworkSections,
 }
 
 /// A free-form `systemd.netdev` section.
+///
+/// Visit the [systemd documentation](docs) to learn more.
+///
+/// docs: https://www.freedesktop.org/software/systemd/man/latest/systemd.netdev.html
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct SdSection {
     pub name: String,
     pub attributes: Vec<(String, String)>,
+}
+
+/// A free-form `NetworkManager` section.
+///
+/// Visit the [NetworkManager documentation](docs) to learn more.
+///
+/// docs: https://www.networkmanager.dev/docs/api/latest/ref-settings.html
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct NmSection {
+    pub name: String,
+    pub attributes: Vec<(String, String)>,
+}
+
+/// Defines the free-form network sections used for defining custom attributes for
+/// `systemd.netdev` and `NetworkManager`.
+///
+/// See [`SdSection`] and [`NmSection`] for more information on each.
+//
+// NOTE: These are stored together in this struct to enforce setting
+// both at the same time. This should ensure any provider implementation
+// which uses custom attributes will at least consider both formats.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct CustomNetworkSections {
+    sd_netdev_sections: Vec<SdSection>,
+    nm_sections: Vec<NmSection>,
+}
+
+impl CustomNetworkSections {
+    pub fn new(sd_netdev_sections: Vec<SdSection>, nm_sections: Vec<NmSection>) -> Self {
+        Self {
+            sd_netdev_sections,
+            nm_sections,
+        }
+    }
 }
 
 /// Supported virtual network device kinds.
@@ -174,19 +213,24 @@ impl DhcpSetting {
 }
 
 impl Interface {
-    /// Return a deterministic `systemd.network` unit name for this device.
-    pub fn sd_network_unit_name(&self) -> Result<String> {
-        let iface_name = match (&self.name, &self.mac_address, &self.path) {
-            (Some(ref name), _, _) => name.clone(),
+    /// Return a deterministic name for this device
+    pub fn name(&self) -> Result<String> {
+        Ok(match (&self.name, &self.mac_address, &self.path) {
+            (Some(ref name), _, _) => name.to_owned(),
             (None, Some(ref addr), _) => addr.to_string(),
-            (None, None, Some(ref path)) => path.to_string(),
+            (None, None, Some(ref path)) => path.to_owned(),
             (None, None, None) => bail!("network interface without name, MAC address, or path"),
-        };
-        let unit_name = format!("{:02}-{}.network", self.priority, iface_name);
-        Ok(unit_name)
+        })
     }
 
-    pub fn config(&self) -> String {
+    /// Return a deterministic `systemd.networkd` unit name for this device.
+    pub fn sd_unit_name(&self) -> Result<String> {
+        let iface_name = self.name()?;
+        Ok(format!("{:02}-{iface_name}.network", self.priority))
+    }
+
+    /// Return the `systemd.networkd` configuration for this device.
+    pub fn sd_config(&self) -> String {
         let mut config = String::new();
 
         // [Match] section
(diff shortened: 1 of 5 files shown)


src/network.rs
```diff
@@ -223,6 +245,148 @@ impl Interface {
...
+                writeln!(config, "interface-name={}", name)?;
+            }
+        }
+        writeln!(config, "autoconnect=true")?;
+        writeln!(
+            config,
+            "autoconnect-priority={}",
+            // Lower number means higher priority for systemd, but it's the opposite for NM
+            100 - (self.priority as u16)
+        )?;
```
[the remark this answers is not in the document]
Rolv-Apneseth: Maybe we should instead restrict the priority to a type that only allows 0-99? Seems like a misconfiguration of the priority value otherwise

src/providers/mod.rs
```diff
@@ -305,6 +306,90 @@ pub trait MetadataProvider {
...
+            configs.push(
+            virt_dev.nm_config(None)
+                .with_context(|| {
+                    format!(
+                        "failed to generate NetworkManager connection profile for virtual device '{}'",
+                        name
+                    )
+                })
+                .map(|c| (name, c))?)
+        }
```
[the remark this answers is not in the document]
Rolv-Apneseth: Sure

src/network.rs
```diff
@@ -223,6 +245,148 @@ impl Interface {
...
+        if !ipv6_dns.is_empty() {
+            let dns_list: Vec<_> = ipv6_dns
+                .iter()
+                .map(std::string::ToString::to_string)
+                .collect();
+            writeln!(config, "dns={}", dns_list.join(";"))?;
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
Rolv-Apneseth: Do we think it's worth extracting the common functionality into a function with additional matches etc.? Seemed OK to me at the time to leave these as is but I don't mind changing.

Rolv-Apneseth:
If we're only running this as a pilot for Hetzner at first, are there any specific changes I need to make here to make that clear? Should this be behind a feature flag for now, or does that make actually using this in our FCOS builds too complicated?

src/network.rs
```diff
@@ -241,6 +285,148 @@ impl Interface {
...
+            if let Some(name) = &self.name {
+                writeln!(config, "interface-name={}", name)?;
+            }
+        }
+        writeln!(config, "autoconnect=true")?;
+        writeln!(
+            config,
+            "autoconnect-priority={}",
+            // Lower number means higher priority for systemd, but it's the opposite for NM
+            100 - (self.priority as u16)
```
prestist: if a provider sets priority > 100. With as u16, a priority of 200 would wrap 100 - 200 to 65436, giving the interface the highest NM priority instead of the lowest. The VirtualNetDev::nm_config() path already handles this correctly with 100i32.saturating_sub_unsigned(priority) -- should this  use the same approach?

src/network.rs
```diff
@@ -241,6 +285,148 @@ impl Interface {
...
+        } else if ipv4_addresses.is_empty() {
+            writeln!(config, "method=disabled")?;
+        } else {
+            writeln!(config, "method=manual")?;
+            for (i, addr) in ipv4_addresses.iter().enumerate() {
+                writeln!(config, "address{}={}", i + 1, addr)?;
+            }
+
+            // IPv4 gateway
+            let (default_route, ipv4_routes): (Vec<NetworkRoute>, Vec<NetworkRoute>) = self
```
prestist, line 411: hmm the ipv4 and ipv6 paths mostly have the same structure at a high level, I wonder if we could abstract it? if not this is very readable and functional as is.

prestist (commented):
I think we have 1 bug? maybe? lmk but otherwise this looks very solid imo

prestist:
> If we're only running this as a pilot for Hetzner at first, are there any specific changes I need to make here to make that clear? Should this be behind a feature flag for now, or does that make actually using this in our FCOS builds too complicated?

I don't think a feature flag is needed here. The NM profile generation is already opt-in -- it only runs when --networkmanager-profiles is passed on the CLI, so no existing provider behavior changes. For the Hetzner pilot, we'd just wire up --networkmanager-profiles in the Hetzner service unit and leave other providers untouched.  I belive using a Rust feature flag would add compile-time complexity without buying us much since the runtime gating is already there IMO.

Rolv-Apneseth:
> The NM profile generation is already opt-in -- it only runs when --networkmanager-profiles is passed on the CLI, so no existing provider behavior changes. For the Hetzner pilot, we'd just wire up --networkmanager-profiles in the Hetzner service unit and leave other providers untouched.

I was thinking more along the lines of if we have to remove this feature, would that be an issue? Do we need to mark it as experimental in some way?

src/network.rs
Rolv-Apneseth: Yeah I'll use the same approach, thanks

src/network.rs
Rolv-Apneseth: This I wasn't sure about - I'm sure I can extract a lot out into a function, but I'm not sure the abstraction is worth it. No strong opinion though. Would you lean towards a shared function?

src/network.rs
Rolv-Apneseth: Feels like maybe we should error if a provider is giving invalid numbers though. Is it over-complicating things to create a type that needs to be <=100?

prestist: Ah, I agree ;) that makse sense to me.

src/network.rs
prestist: Yeah, I agree here as well. The only thing we would get is capturing the flow, and future maintenance if any will be more specific. I think its fine to leave as is honestly; but wanted to see what your thoughts were.

prestist:
> > The NM profile generation is already opt-in -- it only runs when --networkmanager-profiles is passed on the CLI, so no existing provider behavior changes. For the Hetzner pilot, we'd just wire up --networkmanager-profiles in the Hetzner service unit and leave other providers untouched.
> 
> I was thinking more along the lines of if we have to remove this feature, would that be an issue? Do we need to mark it as experimental in some way?

If we had to remove the feature, we still would be removing the opt-in flag and the code. That's effectively the same "experimental" guarantee as a feature flag. I think If the feature was not opt-int i.e called by an arg, then a feature flag would make sense, but since it is opt-in and we dont really use feature flags for afterburn (i.e no processes built around them) its safe to keep as is.

Rolv-Apneseth:
> If we had to remove the feature, we still would be removing the opt-in flag and the code. That's effectively the same "experimental" guarantee as a feature flag. I think If the feature was not opt-int i.e called by an arg, then a feature flag would make sense, but since it is opt-in and we dont really use feature flags for afterburn (i.e no processes built around them) its safe to keep as is.

Makes sense, thanks

prestist (approved):
LGTM

Rolv-Apneseth:
Oh, I have a change coming regarding the priorities if that's OK. Just haven't had time to finish it.

Rolv-Apneseth:
@prestist Added 1 more commit (though I can squash if you prefer) - WDYT?

src/providers/mod.rs
```diff
@@ -305,6 +306,88 @@ pub trait MetadataProvider {
...
+                );
+        }
+
+        // Generate configurations for remaining devices, including bond slaves
+        for interface in physical_interfaces {
+            let name = interface.name()?;
+            configs.push((
+                name,
+                interface.nm_config().with_context(|| {
+                    "failed to generate NetworkManager connection profile for interface '{name}'"
```
prestist, line 354: Ah, I think we need a format!(....) here otherwise `{name}` will be a literal string

prestist (commented):
Overall it looks pretty much ready to land!  I think I found a small bug. 

re: squashing the commits, I think they are good separate imo. Tho small nit in the first commit "Support" looks better lowercase wdyt?

src/providers/mod.rs
Rolv-Apneseth: Ah, whoops, thanks

Rolv-Apneseth:
> Tho small nit in the first commit "Support" looks better lowercase wdyt?

Yeah I do prefer all lowercase. I think I had read somewhere in our docs to do otherwise but seems it's up to preference from what I've seen.

prestist (approved):
LGTM, thank you.

PR #1276, opened 2026-05-21
Update MSRV to 1.85

yasminvalim:


prestist:
@yasminvalim what was the trigger of this? just wondering?

yasminvalim:
@prestist  As far I understood, this[ dependabot PR](https://github.com/coreos/afterburn/pull/1274) updates some dependencies that needs an updated MSRV .

prestist:
Ah, that makes sense, we should probably update that pr with this then :)

yasminvalim:
I was thinking about merge this one and then rebase the dependabot PR so we can have those separately to visualize the changes better. WDYT?

prestist:
Its probably better to do in the pr that is requiring it, only so you can be sure that it solves all the issues for those deps. 

Though if you want to keep it this way we can, up to you ?

yasminvalim:
Fixed in: https://github.com/coreos/afterburn/pull/1274

PR #1294, opened 2026-08-02
proxmoxve: Emit one nameserver= karg per address

### LINKED ISSUE #1225: ProxmoxVE: nameservers not configured if multiple nameservers are provided in network-data
# Bug #

Configuring a Proxmox VM with the following network-config causes the IP and gateway to be set during early boot but no nameservers are configured:
```
version: 1
config:
    - type: physical
      name: eth0
      mac_address: 'bc:24:11:e5:98:ff'
      subnets:
      - type: static
        address: '192.168.88.43'
        netmask: '255.255.255.0'
        gateway: '192.168.88.1'
    - type: nameserver
      address:
      - '192.168.88.30'
      - '192.168.88.31'
```

using a single nameserver works correctly.

Fedora CoreOS 42.20250705.1.0

## Afterburn Version ##

Afterburn-multi 5.8.2

## Environment ##

What hardware/cloud provider/hypervisor is being used to run Afterburn?

Proxmox VE 8.4.1

## Expected Behavior ##

[root@test ~]# resolvectl
Global
         Protocols: LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported
  resolv.conf mode: stub

Link 2 (ens18)
    Current Scopes: DNS LLMNR/IPv4 LLMNR/IPv6
         Protocols: +DefaultRoute LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported
Current DNS Server: 192.168.88.30
       DNS Servers: 192.168.88.30
     Default Route: yes
```

## Actual Behavior ##

```
[root@test ~]# resolvectl
Global
         Protocols: LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported
  resolv.conf mode: stub

Link 2 (ens18)
    Current Scopes: LLMNR/IPv4 LLMNR/IPv6
         Protocols: -DefaultRoute LLMNR=resolve -mDNS -DNSOverTLS DNSSEC=no/unsupported
     Default Route: no
```

## Reproduction Steps ##

  1. Create a new Proxmox VM
  2. Import https://builds.coreos.fedoraproject.org/prod/streams/next/builds/42.20250705.1.0/x86_64/fedora-coreos-42.20250705.1.0-proxmoxve.x86_64.qcow2.xz as the root disk 
  3. Add a cloud-init drive
  4. Configure IP, gateway, and two nameservers in Proxmox's cloud-init UI
  5. Boot the OS
  6. Run `resolvectl` and see no nameservers are configured
  7. Check /etc/NetworkManager/system-connections/default_connection.nmconnection and see no nameservers are configured
 
 ## Other Information ##

https://github.com/coreos/afterburn/blob/main/src/providers/proxmoxve/cloudconfig.rs#L246 seems to be wrong.  According to https://www.man7.org/linux/man-pages/man7/dracut.cmdline.7.html, each nameserver needs a separate nameserver= argument, not commas.

### PR DESCRIPTION
**Pull request by Ar-maan05**
Fixes https://github.com/coreos/afterburn/issues/1225

A ProxmoxVE `network-config` that lists more than one nameserver:

```yaml
    - type: nameserver
      address:
      - '192.168.88.30'
      - '192.168.88.31'
```

configures the IP and gateway during early boot but leaves DNS unset. With a single nameserver it works.

`rd_network_kargs()` joined the addresses with a comma and emitted one karg:

```
nameserver=192.168.88.30,192.168.88.31
```

dracut's `nameserver=` takes a single address and is repeated to configure more than one; there is no comma-separated form, so the whole value is discarded and no nameserver is configured at all. One address happens to work because a one-element join is just that address.

The KubeVirt provider already does this correctly, so this change follows it: collect the addresses and emit one `nameserver=` karg each, passing them through `dracut_addr()` so an IPv6 nameserver is bracketed (`nameserver=[2001:db8::1]`), which the comma-joined path also got wrong.

The addresses are deduplicated across interfaces because `networks()` attaches the same nameserver list to every physical interface, so the two-interface fixture previously emitted the karg twice.

The three ProxmoxVE karg tests asserted the comma-joined string, so they encoded the bug; they now assert one karg per address and that the joined form is absent. All three fail against the old implementation. The DHCP fixture gained an IPv6 nameserver to cover the bracketing.

`cargo test --all-targets` (149 passed), `cargo clippy --all-targets -- -D warnings` and `cargo fmt -- --check -l` are clean, and `python3 tests/fixtures/validate.py` passes (14/14) with the updated fixture.


Rolv-Apneseth:
Nice, thank you for this one too @Ar-maan05.

WDYT about using a `HashSet` for `all_nameservers`?

```diff
@@ -186,7 +186,8 @@ impl MetadataProvider for ProxmoxVECloudConfig {

     fn rd_network_kargs(&self) -> Result<Option<String>> {
         let mut kargs = Vec::new();
-        let mut all_nameservers: Vec<IpAddr> = Vec::new();
+        //  The same nameserver list is attached to every interface, so de-duplicate.
+        let mut all_nameservers: HashSet<IpAddr> = HashSet::new();
 
         if let Ok(networks) = self.networks() {
             for iface in networks {
@@ -245,13 +246,7 @@ impl MetadataProvider for ProxmoxVECloudConfig {
                     }
                 }
 
-                // Collect nameservers from all interfaces. The same nameserver
-                // list is attached to every interface, so deduplicate.
-                for nameserver in &iface.nameservers {
-                    if !all_nameservers.contains(nameserver) {
-                        all_nameservers.push(*nameserver);
-                    }
-                }
+                all_nameservers.extend(&iface.nameservers);
             }
         }
 ```

**Comment by Ar-maan05** [comment]
@Rolv-Apneseth Thanks for the review!

I would lean toward keeping an ordered collection here. `HashSet` iteration order is unspecified and randomized per process, so the emitted `nameserver=` kargs would come out in a different order on each boot. dracut hands those to the resolver in the order given, so that turns a stable resolver priority into a per-boot coin flip. The current tests are order-insensitive so they would still pass, but the behavior change seems worth avoiding.

If the `contains` loop is the part that grates, I'm happy to keep the `Vec` for ordering and use a `HashSet` purely as a seen-set:
```
let mut all_nameservers: Vec<IpAddr> = Vec::new();
let mut seen: HashSet<IpAddr> = HashSet::new();
...
    all_nameservers.extend(iface.nameservers.iter().copied().filter(|ns| seen.insert(*ns)));
```

Though for the handful of addresses involved, `Vec::contains` is probably the clearer of the two. It also matches what `kubevirt/cloudconfig.rs` and `kubevirt/configdrive.rs` already do, so the three providers stay consistent.

I'd be happy to push whichever you prefer.

Rolv-Apneseth:
Ah, good point, thanks for the break down. I was just wondering if this could be a little cleaner, but it's not a big deal. I prefer just `Vec::contains` over a separate seen-set for sure, so let's keep it as is. Unless you think a `BTreeSet` would be worth it?

Ar-maan05:
`BTreeSet` would at least be deterministic, so it dodges the per-boot randomization problem. But it orders by numeric address rather than by the order Proxmox handed us, so it's still a silent reordering: an admin who lists their preferred resolver first would get it moved based on its IP value, and all IPv4 addresses would sort ahead of any IPv6 ones. That seems like the same class of issue as before, just stable instead of random, so I'd rather leave the Vec as-is.

Rebased on main to fix the release-notes conflict with #1295, CI should be green again.

Rolv-Apneseth:
Ah, true. Thanks again.

Rolv-Apneseth (approved):

PR #1295, opened 2026-08-02
providers: Don't panic truncating a multi-byte hostname

Ar-maan05:
`truncate_hostname()` truncates the provider-supplied hostname to `sysconf(_SC_HOST_NAME_MAX)` (64 on Linux) with `String::truncate(maxlen)`. `maxlen` counts **bytes**, but `String::truncate` panics unless the index is a UTF-8 character boundary:

```
thread 'main' panicked at src/providers/mod.rs:205:
assertion failed: self.is_char_boundary(new_len)
```

So any hostname longer than 64 bytes whose 65th byte falls inside a multi-byte sequence panics Afterburn. This is reachable from `write_hostname()`, i.e. the ordinary `--hostname-file` path used by every provider that reports a hostname, and from `write_hostname_ignition_fragment()`. On a cloud instance that means `afterburn-hostname.service` fails at boot.

Walk back to the preceding character boundary before truncating, so an over-long hostname is cut to whole characters instead of panicking.

The added test straddles the limit with a two-byte character starting at `maxlen - 1`, so it is deterministic whatever `maxlen` turns out to be at runtime; a test using only three-byte characters would silently pass on any system where `maxlen % 3 == 0`.

Verified: the new test panics at `src/providers/mod.rs` without the fix and passes with it; `cargo test --all-targets` (150 passed), `cargo clippy --all-targets -- -D warnings` and `cargo fmt -- --check -l` are clean.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -22,6 +22,7 @@ Minor changes:
 - ProxmoxVE: Explicitely deactivate Dracut IP autoconf as the IP is statically set
 - Oracle Cloud: Handle nested metadata values for instance pool members
 - Drop the maplit dependency
+- Don't panic truncating an over-long hostname with multi-byte characters
 
 Packaging changes:
 
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -205,7 +205,14 @@ fn truncate_hostname(mut hostname: String) -> Result<String> {
                 hostname,
                 maxlen
             );
-            hostname.truncate(maxlen);
+            // Truncate on a character boundary: String::truncate() panics if the
+            // index lands in the middle of a multi-byte UTF-8 sequence, and
+            // maxlen counts bytes.
+            let mut end = maxlen;
+            while !hostname.is_char_boundary(end) {
+                end -= 1;
+            }
+            hostname.truncate(end);
             if let Some(idx) = hostname.find('.') {
                 hostname.truncate(idx);
             }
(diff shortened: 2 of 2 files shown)


Rolv-Apneseth:
Wow, that's quite the foot gun. The more you know. Thanks for opening this.

If we were to bump the MSRV to 1.92.0, thoughts on doing something like this instead:

```rust
+++ w/src/providers/mod.rs
@@ -208,10 +208,7 @@ fn truncate_hostname(mut hostname: String) -> Result<String> {
             // Truncate on a character boundary: String::truncate() panics if the
             // index lands in the middle of a multi-byte UTF-8 sequence, and
             // maxlen counts bytes.
-            let mut end = maxlen;
-            while !hostname.is_char_boundary(end) {
-                end -= 1;
-            }
+            let end = hostname.as_str().floor_char_boundary(maxlen);
             hostname.truncate(end);
             if let Some(idx) = hostname.find('.') {
                 hostname.truncate(idx);
```

Ar-maan05:
Thanks for taking a look!

`floor_char_boundary` would definitely be the nicer spelling, and it's exactly equivalent to the loop here. My only hesitation is that Cargo.toml currently has `rust-version = "1.88.0"`, and CI runs a "minimum supported toolchain" job, so adopting it means an MSRV bump lands as part of a panic fix. Since this one is a crash at boot on any host with a multi-byte hostname, I'd rather keep it backportable to the current MSRV.

So my suggestion: land the manual loop as-is now, and I'm glad to send a follow-up swapping it for `floor_char_boundary` whenever you bump the MSRV. If you'd prefer to do the bump as part of this PR instead, that's fine too, just let me know and I'll push it.

Rolv-Apneseth:
Hmm, we'll see if either of us actually remember to make the change, but sounds reasonable. Thanks again.

Rolv-Apneseth (approved):

PR #1299, opened 2026-08-21
initrd: Write an env file for systemd-network-generator to consume

chewi:
systemd-network-generator parses /proc/cmdline or `SYSTEMD_PROC_CMDLINE`, so the kargs we inject into Dracut's cmdline.d fragment are invisible to it. Flatcar would like to switch from its own customer parser to the generator's parser (and not Dracut's own). Apart from writing the env file, some compatibility issues had to be addressed.

kubevirt currently uses `static` to disable IP autoconfiguration, but both Dracut and systemd-network-generator reject this as invalid. It was presumably added with NetworkManager in mind. We should therefore use `off` instead.  `none` would also work, but we already use `off` elsewhere.

**Important note!** NetworkManager does not treat `static` and `off` entirely equally. The former sets the method for the other protocol (the one not being configured by `ip=`) to `auto`, enabling DHCP for IPv4 or SLAAC for IPv6, while the latter sets it to `disabled`, giving the other protocol no address at all. Given that `static` is supposed to be exactly that, the current behaviour is arguably unexpected, while the `off` behaviour seems more correct.

systemd-network-generator also does not support multiple autoconf values, and fixing this appears difficult. Even then, Dracut does not treat `dhcp,dhcp6` in quite the same way that systemd and NetworkManager do. It only tries IPv6 if IPv4 fails. I am therefore changing kubevirt and proxmoxve to use `ip=any` rather than `ip=dhcp,dhcp6` for dual stack.

As of Dracut 112, `any` is treated as IPv4-only, despite the name. However, I have submitted dracut-ng/dracut#2628 to change this behaviour so that it is dual stack in the same manner as systemd and NetworkManager.

This would break Afterburn consumers relying on Dracut's network-legacy in the meantime. However, FCOS has already migrated to NetworkManager, and Flatcar is now migrating to systemd-network-manager. I'm not aware of any other consumers.

Dual stack configuration didn't work on proxmoxve anyway because Afterburn would iterate over the subnets, setting the type as either `V4` or `V6`, but never combining them. This includes a fix for that.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -27,6 +27,7 @@ Minor changes:
 - ProxmoxVE: Emit one `nameserver=` karg per address so multiple DNS servers work
 - ProxmoxVE: Prevent rd-network-kargs from failing without a config drive
 - KubeVirt: Use `off` rather than `static` to disable IP autoconf for Dracut and systemd compatibility
+- KubeVirt, ProxmoxVE: Use `ip=any` rather than `ip=dhcp,dhcp6` for systemd compatibility. As of Dracut 112, its network-legacy module treats `ip=any` as IPv4-only, so this change breaks IPv6 in that case. At the time of writing, a Dracut change to make this dual stack is under review.
 
 Packaging changes:
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -26,6 +26,7 @@ Minor changes:
 - Don't panic truncating an over-long hostname with multi-byte characters
 - ProxmoxVE: Emit one `nameserver=` karg per address so multiple DNS servers work
 - ProxmoxVE: Prevent rd-network-kargs from failing without a config drive
+- KubeVirt: Use `off` rather than `static` to disable IP autoconf for Dracut and systemd compatibility
 
 Packaging changes:
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -12,6 +12,7 @@ Major changes:
 - Azure: Add `render-ignition` subcommand to generate Ignition config fragments from IMDS metadata
 - Hetzner: Add support for network configuration
 - Support the generation of NetworkManager connection profiles (keyfile format)
+- Support systemd-network-generator by emitting `/run/afterburn/network-generator.env` from the initrd
 
 Minor changes:
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -25,6 +25,7 @@ Minor changes:
 - Drop the maplit dependency
 - Don't panic truncating an over-long hostname with multi-byte characters
 - ProxmoxVE: Emit one `nameserver=` karg per address so multiple DNS servers work
+- ProxmoxVE: Prevent rd-network-kargs from failing without a config drive
 
 Packaging changes:
 
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -28,6 +28,7 @@ Minor changes:
 - ProxmoxVE: Prevent rd-network-kargs from failing without a config drive
 - KubeVirt: Use `off` rather than `static` to disable IP autoconf for Dracut and systemd compatibility
 - KubeVirt, ProxmoxVE: Use `ip=any` rather than `ip=dhcp,dhcp6` for systemd compatibility. As of Dracut 112, its network-legacy module treats `ip=any` as IPv4-only, so this change breaks IPv6 in that case. At the time of writing, a Dracut change to make this dual stack is under review.
+- ProxmoxVE: Fix dual stack network configuration
 
 Packaging changes:
 
--- a/src/initrd/mod.rs
+++ b/src/initrd/mod.rs
@@ -9,11 +9,24 @@ use crate::providers::proxmoxve::ProxmoxVEConfigDrive;
 use crate::providers::vmware::VmwareProvider;
 use crate::providers::MetadataProvider;
 use anyhow::{Context, Result};
-use std::fs::File;
+use std::fs::{self, File};
 use std::io::Write;
+use std::path::Path;
 
 /// Path to cmdline.d fragment for network kernel arguments.
 static KARGS_PATH: &str = "/etc/cmdline.d/50-afterburn-network-kargs.conf";
+
+/// Path to the environment file consumed by systemd-network-generator.
+///
+/// systemd-network-generator only parses /proc/cmdline (or the command line
+/// provided through the SYSTEMD_PROC_CMDLINE environment variable), so the
+/// kargs we inject into Dracut's cmdline.d fragment above are invisible to it.
+/// We therefore expose the augmented command line through this file, which a
+/// shipped generator drop-in references via EnvironmentFile=.
+static GENERATOR_ENV_PATH: &str = "/run/afterburn/network-generator.env";
+
+/// Path to the kernel command line.
+static PROC_CMDLINE_PATH: &str = "/proc/cmdline";
 
 /// Fetch network kargs for the given provider.
 pub(crate) fn fetch_network_kargs(provider: &str) -> Result<Option<String>> {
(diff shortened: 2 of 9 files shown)


chewi:
Note that I'm currently testing that last proxmoxve dual stack fix. Just waiting on a Flatcar build. I'll report back on Monday.

src/initrd/mod.rs
```diff
@@ -5,30 +5,50 @@
...
 }
 
-/// Write network kargs into a cmdline.d fragment.
+/// Write network kargs into a cmdline.d fragment and hand the augmented kernel
+/// command line to systemd-network-generator.
 pub(crate) fn write_network_kargs(kargs: &str) -> Result<()> {
-    let mut fragment_file = File::create(KARGS_PATH)
-        .with_context(|| format!("failed to create file {KARGS_PATH:?}"))?;
+    write_cmdline_fragment(KARGS_PATH, kargs)?;
+    write_generator_env(GENERATOR_ENV_PATH, PROC_CMDLINE_PATH, kargs)
```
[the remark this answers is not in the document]
chewi: No. It wrote the Dracut fragment unconditionally before, and I do not want to change that behaviour.

In the systemd-network-generator case, a system unit can ignore a missing environment file with `EnvironmentFile=-/path/to/env`, so the difference is intentional.

src/providers/proxmoxve/tests.rs
```diff
@@ -210,6 +210,37 @@ fn test_network_kargs_no_gateway() {
...
+
+    let kargs = config.rd_network_kargs().expect("cannot get network kargs");
+    assert!(kargs.is_some());
+    let kargs = kargs.unwrap();
+
+    // Dual-stack DHCP maps to the "any" autoconf keyword.
+    assert!(kargs.contains("ip=any"));
+
+    // Check nameservers
+    assert!(kargs.contains("nameserver=1.1.1.1,8.8.8.8"));
```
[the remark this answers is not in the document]
chewi: This is overkill. Other similar strings are not going to appear out of nowhere and `dhcp,dhcp6` no longer features anywhere in the codebase.

chewi:
Okay, proxmoxve dual stack is tested now.

chewi:
There were a couple of CI failures, but I'm not sure if they're relevant. ext.config.extensions.package probably isn't. ext.config.networking.network-online-service might be, but it doesn't give me much to go on.

Rolv-Apneseth:
Hey @chewi, thanks for the PR

> There were a couple of CI failures, but I'm not sure if they're relevant. ext.config.extensions.package probably isn't. ext.config.networking.network-online-service might be, but it doesn't give me much to go on.

Yeah does seem like `extensions.pacakge` is unrelated, maybe something to do with the ongoing branching. I haven't had time to fully look into it yet and won't until Monday, but the test file for `network-online-service` is over at https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/networking/network-online-service if that's of any help, but it may well be unrelated too.

src/providers/proxmoxve/tests.rs
```diff
@@ -221,6 +221,39 @@ fn test_network_kargs_no_gateway() {
     assert!(!kargs.contains("nameserver=1.1.1.1,8.8.8.8"));
 }
 
+#[test]
+fn test_network_dhcp_dual() {
```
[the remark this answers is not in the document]
chewi: In this case, I think it makes more sense to keep these tests together with existing similar tests.

chewi:
> Yeah does seem like `extensions.pacakge` is unrelated, maybe something to do with the ongoing branching. I haven't had time to fully look into it yet and won't until Monday, but the test file for `network-online-service` is over at https://github.com/coreos/fedora-coreos-config/blob/testing-devel/tests/kola/networking/network-online-service if that's of any help, but it may well be unrelated too.

I can't see any reason why these changes would trigger network-online. If anything, I'd expect the opposite. Let's give it another run. I had to update the tests following recent changes anyway.

chewi:
Looking better this time. :slightly_smiling_face: That TMT issue is surely unrelated.

Rolv-Apneseth:
> Looking better this time. 🙂 That TMT issue is surely unrelated.

Nice :)

Yes it does seem unrelated, yet again. I'll make a separate fix for that. 

I'm busy for the rest of the day, unfortunately, but I'll be sure to review this properly tomorrow.

docs/release-notes.md
```diff
@@ -12,6 +12,7 @@ Major changes:
 - Azure: Add `render-ignition` subcommand to generate Ignition config fragments from IMDS metadata
 - Hetzner: Add support for network configuration
 - Support the generation of NetworkManager connection profiles (keyfile format)
+- rd-network-kargs: Support systemd-network-generator by emitting `/run/afterburn/network-generator.env`
```
Rolv-Apneseth:
```suggestion
- Support systemd-network-generator by emitting `/run/afterburn/network-generator.env` from the initrd
```

src/initrd/mod.rs
```diff
@@ -39,3 +59,135 @@ pub(crate) fn write_network_kargs(kargs: &str) -> Result<()> {
...
+    };
+
+    if let Some(parent) = Path::new(path).parent() {
+        fs::create_dir_all(parent)
+            .with_context(|| format!("failed to create directory {parent:?}"))?;
+    }
+
+    let mut env_file =
+        File::create(path).with_context(|| format!("failed to create file {path:?}"))?;
+    writeln!(env_file, "SYSTEMD_PROC_CMDLINE={cmdline}")
```
Rolv-Apneseth, line 91: Not sure if there would be any issues in practice, but I think we may as well quote this:

```suggestion
    writeln!(env_file, "SYSTEMD_PROC_CMDLINE='{cmdline}'")
```

docs/release-notes.md
```diff
@@ -27,6 +27,7 @@ Minor changes:
 - ProxmoxVE: Emit one `nameserver=` karg per address so multiple DNS servers work
 - ProxmoxVE: Prevent rd-network-kargs from failing without a config drive
 - KubeVirt: Use `off` rather than `static` to disable IP autoconf for Dracut and systemd compatibility
+- KubeVirt and ProxmoxVE: Use `ip=any` rather than `ip=dhcp,dhcp6` for Dracut and systemd compatibility
```
Rolv-Apneseth: Maybe we should mention the change for `network-legacy` users, though I doubt there are many (if any) relying on this. @prestist may know more

```suggestion
- KubeVirt and ProxmoxVE: Use `ip=any` rather than `ip=dhcp,dhcp6` for Dracut and systemd compatibility (warning: breaking change for network-legacy users)
```

Rolv-Apneseth (commented):
Minor notes but LGTM, thanks a lot for all these. I'll want @prestist's opinion on this one too though.

Also, nit for the double prefix commit, maybe it should be `kubevirt, proxomoxve: ...`. I know for a fact that's done in `containers/toolbox` and seems like a decent convention. Otherwise just `*:` works too.

Rolv-Apneseth:
Oh, and a rebase should address the TMT CI

docs/release-notes.md
chewi: The openSUSE Dracut maintainer told me that they have already migrated to NetworkManager, so it's really just Gentoo now. Neither uses Afterburn.

src/initrd/mod.rs
chewi: No, I think that would cause issues. The command line could potentially include a `'` character. systemd doesn't parse these like a normal shell would. The only reason to explicitly quote would be if the value spanned multiple lines, which would never be the case here. See `EnvironmentFile=` in systemd.exec(5).

Rolv-Apneseth: Right, makes sense, thanks

docs/release-notes.md
Rolv-Apneseth: I'll leave this unresolved for now so @prestist can see, but yes probably not a concern. Was just wondering if it's worth mentioning in the changelog.

krnowak (approved):

Rolv-Apneseth (approved):

docs/release-notes.md
prestist: @chewi / @Rolv-Apneseth thank you for double and triple checking this, we dont want to disrupt anyone. I think given the details we can proceed. I do like being extra cautious when possible, so documenting the breakage is important even if we believe no one to be using it.

docs/release-notes.md
```diff
@@ -24,6 +25,10 @@ Minor changes:
 - Drop the maplit dependency
 - Don't panic truncating an over-long hostname with multi-byte characters
 - ProxmoxVE: Emit one `nameserver=` karg per address so multiple DNS servers work
+- ProxmoxVE: Prevent rd-network-kargs from failing without a config drive
+- KubeVirt: Use `off` rather than `static` to disable IP autoconf for Dracut and systemd compatibility
+- KubeVirt, ProxmoxVE: Use `ip=any` rather than `ip=dhcp,dhcp6` for systemd compatibility. As of Dracut 112, its network-legacy module treats `ip=any` as IPv4-only, so this change breaks IPv6 in that case. At the time of writing, a Dracut change to make this dual stack is under review.
+- ProxmoxVE: Fix dual stack network configuration
```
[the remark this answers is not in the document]
Rolv-Apneseth: `ProxmoxVE` is used consistently elsewhere in the release notes. `dual-stack` vs `dual stack` is pretty minor imo.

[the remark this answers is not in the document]
prestist: @coderabbitai resolve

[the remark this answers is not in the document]
prestist: @coderabbitai help

prestist:
@coderabbitai resolve

prestist (approved):
Overall LGTM, thank you for your work on this @chewi!

Rolv-Apneseth:
Thanks @chewi

PR #1303, opened 2026-09-04
cargo: ensure reqwest uses native tls backend

angelcerveraroldan:
When we upgraded from reqwest 0.12.* to 0.13.*, we changed the default TLS backend from native (openssl) to rustls (rust-tls).

This was not an intended change, and it is better to use openssl so that we maintain FIPS compliance and so that the crypto-config is used.

regression introduced: 6ea7eb8efdc498a5be5d4fef1f8c08429963ef9e

See release notes for the default change [here](https://github.com/seanmonstar/reqwest/releases#release-v0.13.0)

I set the feature flags to be the default ones for the last version: https://docs.rs/crate/reqwest/0.12.28/features#default-tls

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,29 +243,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f2032f911046de80f0a198e0901378627c33f59ea0ac00e363d481118bd70a53"
 
 [[package]]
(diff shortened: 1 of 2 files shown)


Rolv-Apneseth (approved):
LGTM

joelcapitao (approved):

